PR #68, opened 2020-11-28
Add Arch Linux installation instructions

hbarsaiyan:


th-ch (approved):
Thanks for the improvement!

Issue #39, opened 2020-08-02
Touchbar support


k4sud0n:
What a useful app :D

I wish when this app supports touchbar, it would be way more better. (Like displaying song's name..like, dislike etc)

semvis123:
Something like this?
<img width="1085" alt="image" src="https://user-images.githubusercontent.com/28219076/102337768-3234de00-3f93-11eb-828d-5a291cb1fc20.png">
PS: It would be great, If anyone got better icons / characters to but in the buttons.

semvis123:
<img width="1085" alt="image" src="https://user-images.githubusercontent.com/28219076/102391381-ed7e6680-3fd5-11eb-9042-dbb15fe7be25.png">
Even added images

semvis123:
<img width="1085" alt="image" src="https://user-images.githubusercontent.com/28219076/102393371-91691180-3fd8-11eb-86e8-7cadf253a118.png">
Sorry for the spam but likes and dislikes are now also possible

k4sud0n:
Awesome 👏

th-ch:
Closing the issue as TouchBar has been added (credits to @semvis123)!

PR #101, opened 2020-12-16
Added Touch Bar plugin

semvis123:
#39


th-ch (approved):
Looks good and well written! ✅  
Some parts of the plugin (getting song name, title, etc) could even (later) be refactored to be used in multiple plugins (music info is also used in the notifications plugin for example).

Merging, thanks for the contribution! The touchbar plugin will ship in the next version.

PR #102, opened 2020-12-21
Globalized the song info and song controls, and updated Touch Bar for it.

semvis123:
Globalized the song info and song controls.
Now plugins can retrieve the song info.
The plugins don't need to import anything extra to use the song information or controls.
The plugins are able to register a callback for song info changes using
```js 
global.songInfo.onNewData( (songInfo) => {} )
```
`songInfo` is an object, containing the following properties.
```js
title: String
artist: String
views:  String
likes: String
imageSrc: String
image: NativeImage
isPaused: Boolean
onNewData: Void
```
and the song can be controlled using one of the following functions:
```js
global.songControls.previous()
global.songControls.pause()
global.songControls.next()
global.songControls.like()
global.songControls.dislike()
```

 The songInfo plugin should eventually get another location, because it isn't really a plugin.
 
I've already changed the Touch Bar plugin so that it implements this.


plugins/songInfo/back.js
```diff
@@ -0,0 +1,120 @@
+const {nativeImage} = require('electron');
```
th-ch: It's often advised to stick to lowercase filenames because some file systems are case insensitive (e.g. APFS on Macs is by default) and some are case sensitive, so `songInfo` could be renamed to `song-info`

plugins/songInfo/back.js
```diff
@@ -0,0 +1,120 @@
...
+	global.songInfo = {
+		title: '',
+		artist: '',
+		views: '',
+		likes: '',
+		imageSrc: '',
+		image: null,
+		isPaused: true
+	};
+	// The song control funcions
```
th-ch: nit
```suggestion
	// The song control functions
```

th-ch (approved):
Amazing job, thanks for the contribution! 👏 

As you mentioned in the PR description, `song-info` is not really a plugin but more a provider for other plugins, so it should be put in a separate `providers` folders with each plugin using it instantiating the `songInfo` global (here, if a plugin like `notifications` is enabled but not `song-info`, we get `Uncaught Exception: TypeError: Cannot read property 'onNewData' of undefined`)

If you want to make the follow up to have a separate `providers` folder, feel free! Otherwise, I'm fine with merging and doing the follow up, PR is ✅

semvis123:
No problem, I'm glad you liked this 👍
I have changed the little things, this should be enough for a working version, but preferably the providers should not be hard coded.

th-ch (approved):
> this should be enough for a working version

Agreed, I can do the refactoring for provider instantiation in another PR, merging this one! ✅ 🎉

PR #124, opened 2021-01-12
Added Discord rich presence and added extra properties to songInfo provider

semvis123:
* Added Discord rich presence
image

* Added extra properties to songInfo provider:
```javascript
{
    songDuration: Integer,
    ElapsedSeconds: Integer
}
```

### DIFF
--- a/plugins/discord-rpc/back.js
+++ b/plugins/discord-rpc/back.js
@@ -1,0 +1,41 @@
+const DiscordRPC = require('discord-rpc');
+const rpc = new DiscordRPC.Client({
+	transport: 'ipc'
+});
+
+const clientId = '790655993809338398';
+
+module.exports = win => {
+	// If the page is ready, register the callback
+	win.on('ready-to-show', () => {
+		// Startup the rpc client
+		rpc.login({
+			clientId
+		}).catch(console.error);
+
+		// Register the callback
+		global.songInfo.onNewData(songInfo => {
+			// Song information changed, so lets update the rich presence
+
+			const activityInfo = {
+				details: songInfo.title,
+				state: songInfo.artist,
+				largeImageKey: 'logo',
+				largeImageText: songInfo.views + ' - ' + songInfo.likes
+			};
+
+			if (songInfo.isPaused) {
+				// Add an idle icon to show that the song is paused
+				activityInfo.smallImageKey = 'idle';
+				activityInfo.smallImageText = 'idle/paused';
+			} else {
+				// Add the start and end time of the song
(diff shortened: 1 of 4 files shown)


plugins/discord-rpc/back.js
```diff
@@ -0,0 +1,41 @@
+const DiscordRPC = require('discord-rpc');
+const rpc = new DiscordRPC.Client({
+	transport: 'ipc'
+});
+
+const clientId = '790655993809338398';
```
th-ch, line 6: Quick question about the ID, did you register an application in Discord under your own account?

plugins/discord-rpc/back.js
```diff
@@ -0,0 +1,41 @@
+const DiscordRPC = require('discord-rpc');
```
th-ch: nit: would have just called the plugin `discord` (rpc being an implementation detail, it might not be necessary to have it in the plugin name)

th-ch (commented):
Looks promising! Just a question/remark and should be good to go! 
Note: https://github.com/th-ch/youtube-music/pull/125 will refactor the providers, the second PR to be merged will have to be updated to take into account changes.

plugins/discord-rpc/back.js
semvis123: uhm yes, but it can be changed to yours if you want :)

plugins/discord-rpc/back.js
semvis123: Like this?

th-ch: Ah no, sorry for the confusion, I was talking about the filename `plugins/discord-rpc/back.js` -> `plugins/discord/back.js`! The code looks good 👍

th-ch (approved):
Looks good, merging, thanks for the contribution!

alxbrn:
Is there anything specific a individual has to in order to utilize this feature, other than the obvious of enable the plugin in the plugins tab and adding youtube music app as a game in discord settings?

I can't seem to reproduce the same image that was originally posted.

semvis123:
@alxbrn Hmm that shouldn't be necessary, I don't have my YouTube account connected to my Discord and it works fine for me.

Have you restarted both discord and YouTube music before you checked?

Also could you clarify what you mean with "in order to show valid information"?

NNowakowski:
@semvis123 Would it be possible to rename the client to **YouTube Music** instead of **Youtube Music**? :p

semvis123:
@NNowakowski It should be updated right now, thanks for your suggestion!

PR #144, opened 2021-02-01
Reuse the same notification, instead of creating a new one each time the song changes.

semvis123:
This way you don't end up with +99 notifications in your Notification Center


th-ch (approved):

th-ch:
Merged, thanks for the contribution @semvis123! It will ship in the next version!

Issue #151, opened 2021-02-09
[BUG] downloader plugin seems to download the song double


semvis123:
When downloading a song, the song length of the downloaded file is double the length of the song in YouTube music.
for example: 
2:14 in YouTube music
4:28 in the downloaded MP3

PR #152, opened 2021-02-09
Added metadata to downloader plugin, and updated packages

semvis123:
This adds metadata tags to the downloaded file using the downloader plugin, album cover, title and artist.
note: this does *not* fix the double file issue https://github.com/th-ch/youtube-music/issues/151

### DIFF
--- a/plugins/downloader/back.js
+++ b/plugins/downloader/back.js
@@ -1,9 +1,13 @@
+const fs = require('fs');
 const { join } = require("path");
 
-const { dialog } = require("electron");
+const { dialog, ipcMain } = require("electron");
 
 const { injectCSS, listenAction } = require("../utils");
 const { ACTIONS, CHANNEL } = require("./actions.js");
+
+const getSongInfo = require("../../providers/song-info");
+const ID3Writer = require('browser-id3-writer');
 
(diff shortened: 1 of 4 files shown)


plugins/downloader/back.js
```diff
@@ -28,6 +32,37 @@ function handle(win) {
...
+			writer.setFrame('TIT2', songInfo.title)
+				.setFrame('TPE1', [songInfo.artist])
+				.setFrame('APIC', {
+					type: 3,
+					data: coverBuffer,
+					description: ''
+				});
+			writer.addTag();
+
+			// Overwrite the old file, with the new file with metadata
```
th-ch, line 57: Note: instead of writing the file, reading it again into a buffer and finally re-writing it, we can simply pass the buffer along and write the file once at the end!

th-ch (approved):
Looks good, thanks for the contribution! Just a quick note about writing the file only once - merging the PR as it is, I will push the follow up on top of it and it will ship in the next version!

th-ch:
Closing - all changes have been cherry-picked in https://github.com/th-ch/youtube-music/commit/dd1bdae9478ef831ee2a00b29be04c65626933f8

Issue #185, opened 2021-03-10
[Feature Request] Discord Current Song Reporting Timeout (Reposted)


ashirbhallalevine:
As posted in a [previous issue thread](#161) and reposted by recommendation:

Please add a Discord song-reporting timeout such that the app reports which song is being played when paused but clears the song in favor of simply reporting "Playing YouTube Music" after a certain timeout (5 minutes sounds about right to me) and then clears the entire Game Activity status after another defined amount of inactive time until activity is detected again (10 minutes sounds good for this one). I have suggested possible time-frames for these actions, however, a selectable option menu would always be preferred.

Please do let me know if you have any questions about the above feature request and I would be more than happy to clarify!

semvis123:
I made this feature in #192, would love to hear your opinion about it.

ashirbhallalevine:
> I made this feature in #192, would love to hear your opinion about it.

Hey, thank you for picking up this feature request! I took a look at the file changes you made in #192 and, as far as I can tell, it is an accurate reflection of the feature I requested above. Like I said in the original suggestion, an additional dropdown menu with a few different options (say 5 mins, 10 mins, 15 mins) for the "activityTimoutTime" would be even better.

semvis123:
Thanks for your feedback, the time is configurable in the `advanced options`, so I don't think we need a dropdown menu specifically for this setting of this plugin.

ashirbhallalevine:
> Thanks for your feedback, the time is configurable in the `advanced options`, so I don't think we need a dropdown menu specifically for this setting of this plugin.

Sounds good! I figured this may be the case, but I couldn't quite tell from the files you changed as I am still familiarizing myself with the file structure of this project. Thanks!

semvis123:
There should be a (menu) button under `plugins` called `advanced options` which should take you to the specific file you will need to change.

ashirbhallalevine:
> There should be a (menu) button under `plugins` called `advanced options` which should take you to the specific file you will need to change.

Ah, yes. I see that the `activityTimeoutTime` variable is defined in the `config.json` file. Thanks again for adding this feature!

semvis123:
No problem!

PR #192, opened 2021-03-17
Added Discord timeout

semvis123:
This adds a timeout, that clears the Discord rich presence after being paused for a specific time.
This was requested in #185

### DIFF
--- a/plugins/discord/back.js
+++ b/plugins/discord/back.js
@@ -9,7 +9,9 @@ const rpc = new Discord.Client({
 // Application ID registered by @semvis123
 const clientId = "790655993809338398";
 
-module.exports = (win) => {
+let clearActivity;
+
+module.exports = (win, {activityTimoutEnabled, activityTimoutTime}) => {
 	const registerCallback = getSongInfo(win);
 
 	// If the page is ready, register the callback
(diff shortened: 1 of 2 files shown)


plugins/discord/back.js
```diff
@@ -29,7 +31,13 @@ module.exports = (win) => {
 					// Add an idle icon to show that the song is paused
 					activityInfo.smallImageKey = "idle";
 					activityInfo.smallImageText = "idle/paused";
+					// Set start the timer so the activity gets cleared after a while if enabled
+					if (activityTimoutEnabled)
+						clearActivity = setTimeout(()=>rpc.clearActivity(), activityTimoutTime||10,000);
```
th-ch, line 36: nit: fallback (10,000) is not the same as the default (60,000), not sure if it's expected

th-ch (approved):
Looks good, thanks for the contribution! Merging ✅

plugins/discord/back.js
semvis123: Yeah I did that on purpose, if you delete the config / don't have the config, it defaults to 10 seconds, could maybe be changed so that people that are upgrading from older versions generate that config.

Issue #176, opened 2021-03-04
[Improvement] Get the song info from the `/player` request


semvis123:
The song info can be found in the `/player` request made when the song is changed.
It would be a lot better to grab the information from there instead of grabbing the information from the DOM.
But it would be quite a challenge to read the existing request.

PR #194, opened 2021-03-18
Improved songinfo provider, by using the data from the '/player' request

semvis123:
This provides more reliable song information,
Note: songinfo object has been changed a bit,

```js
views: String -> Integer
likes: String -> Removed (not enough information to provide this reliably)
uploadDate: Undefined -> String
```

I also updated the Discord plugin

See #176


providers/song-info.js
```diff
@@ -1,73 +1,20 @@
+const { ipcMain } = require("electron");
```
th-ch, line 1: nit: the 2 electron imports could be bundled in one

th-ch (approved):
Looks good, thanks for the contribution, merging! ✅

providers/song-info.js
semvis123: Oh whoops, didn’t spot this one

PR #196, opened 2021-03-19
Last.fm support

semvis123:
This adds a last.fm plugin with the following features,
* Scrobble support, adds scrobbles to your last.fm account.
* Now playing support, also known as currently scrobbling.

When the plugin is enabled for the first time, it should ask the user to authenticate with last.fm and give permissions to YouTube music.


plugins/last-fm/back.js
```diff
@@ -0,0 +1,192 @@
+const fetch = require('node-fetch');
+const md5 = require('md5');
+const open = require("open");
+const axios = require('axios');
+const { setOptions } = require('../../config/plugins');
+const getSongInfo = require('../../providers/song-info');
+
+const defaultSettings = {
```
th-ch: Can be replaced by `const defaultConfig = require('../../config/defaults');` then getting the lastFM default config

plugins/last-fm/back.js
```diff
@@ -0,0 +1,192 @@
...
+		formData.append(key, params[key]);
+	}
+	return formData;
+}
+const createQueryString = (params, api_sig) => {
+	// creates a querystring
+	const queryData = [];
+	params.api_sig = api_sig;
+	for (key in params) {
+		queryData.push(`${key}=${params[key]}`);
```
th-ch: To avoid issues:
```suggestion
		queryData.push(`${encodeURIComponent(key)}=${encodeURIComponent (params[key])}`);
```

plugins/last-fm/back.js
```diff
@@ -0,0 +1,192 @@
...
+	api_root: "http://ws.audioscrobbler.com/2.0/",
+	api_key: "04d76faaac8726e60988e14c105d421a", // api key registered by @semvis123
+	secret: "a5d2a36fdf64819290f6982481eaffa2",
+	suffixesToRemove: [' - Topic', 'VEVO']
+}
+
+const cleanupArtistName = (config, artist) => {
+	// removes the suffixes of the artist name for more recognition by last.fm
+	let { suffixesToRemove } = config;
+	if (suffixesToRemove === undefined) {
```
th-ch: nit: it does not seem necessary to override the suffixes if not defined, we can just return early:
```suggestion
	if (suffixesToRemove === undefined) {
	    return artist;
	}
```

plugins/last-fm/back.js
```diff
@@ -0,0 +1,192 @@
...
+		artist: songInfo.artist,
+		api_key: config.api_key,
+		sk: config.session_key,
+		format: 'json',
+		method: 'track.scrobble',
+		timestamp: ~~((Date.now() - songInfo.elapsedSeconds) / 1000),
+		duration: songInfo.songDuration,
+	};
+	data.api_sig = createApiSig(data, config.secret);
+	axios.post('https://ws.audioscrobbler.com/2.0/', createFormData(data))
```
th-ch: nit: node-fetch is already there and could be used (but axios is well-known and widely used so no big deal and no need to change it)

plugins/last-fm/back.js
```diff
@@ -0,0 +1,192 @@
...
+		track: songInfo.title,
+		artist: songInfo.artist,
+		api_key: config.api_key,
+		sk: config.session_key,
+		format: 'json',
+		method: 'track.updateNowPlaying',
+		duration: songInfo.songDuration,
+	};
+	data.api_sig = createApiSig(data, config.secret);
+	axios.post('https://ws.audioscrobbler.com/2.0/', createFormData(data))
```
th-ch: Code looks mostly the same as `addScrobble` so it could probably be refactored with a util function.

th-ch (commented):
Looks good, thanks for the contribution! Left a few minor comments, feel free to address them, otherwise I'm ok to merge and do a follow up!

plugins/last-fm/back.js
semvis123: Ah, didn't know you could post data with node-fetch, removed axios, because there is no need for an extra dependency :)

semvis123:
Thanks for your feedback, I have done some additional refactoring and solved the merge conflicts.

yarn.lock
```diff
@@ -1708,6 +1708,13 @@ aws4@^1.8.0:
   resolved "https://registry.yarnpkg.com/aws4/-/aws4-1.11.0.tgz#d61f46d83b2519250e2784daf5b09479a8b41c59"
   integrity sha512-xh1Rl34h6Fi1DC2WWKfxUTVqRsNnr6LsKz2+hfwDxQJWmrx8+c7ylaqBMcHfl1U1r2dsifOvKX3LQuLNZ+XSvA==
 
+axios@^0.21.1:
```
th-ch, line 1711: Not a big deal but it seems the lock file is not up-to-date (axios is not used anymore), `yarn` will need to be run on the project to update it!

th-ch (approved):
Code looks good, thanks for the contribution! Merging so that it can ship in the next version, will also update the `yarn.lock` file!

VanGeko:
any chance to see the lastfm plugin, in the ytm App?
there is a lot of wrong titles and i want to correct them.

Issue #145, opened 2021-02-04
Play/Pause menu item in taskbar icon Context Menu doesn't pause


nshannon123:
The play/pause button in the taskbar icon's menu won't pause the current playing song. It will play the song if the track is currently paused but pausing it doesn't work.

th-ch:
Hey @nshannon123, thanks for the report! I seem able to reproduce, first conclusions point towards the `auto-confirm-when-paused` plugin, which will unpause the video when you try to pause it if the window is not active - so if you have the `auto-confirm-when-paused` plugin enabled, disabling it should do the trick until a better solution comes out!

nshannon123:
Thanks for the reply. At least you found the issue. Ideally it would be better to have the plugin enabled as well as being able to pause from the taskbar. Hopefully this will be a fix included in a future release!

PR #200, opened 2021-03-21
[Plugin] taskbar-mediacontrol  (for Windows)

### LINKED ISSUE #107: [UI\Feature Request] Windows Taskbar Thumbnail Media Buttons
Hi!
I really love your software and appreciate all the effort!!
The only thing missing for me at the moment is the taskbar media controls.
It exists in almost any (binary based) media player and should be simple to implement (as we always say before actually implementing a feature)

This is low priority, but should be simple.
If it's not simple or takes a long time - please close the issue until more people ask, too.

Again, I LOVE your client!

Araxeus:
Add UI Control to windows taskbar preview (official name is Thumbar)
Closes #107
demo

(had to add img's for icons as per win32 api rules)

There seems to be a bug in electron - after `window.hide()` is called, the Thumbar disappears and seems to just break (cant show it again)
I [reported](https://github.com/electron/electron/issues/28319) that bug to them (will be [fixed](https://github.com/electron/electron/pull/28366) in next electron version)
but since we are using Electron 11.xx - my local fix for the problem is using

`win.minimize(); win.setSkipTaskbar(true);`
instead of
`window.hide();`

Tested it and it works great now :)

EDIT:
uses playPause() function to control play state - which means it currently has the same conflict as Play/Pause in tray menu:

- See #145 
autoConfirmWhenPaused plugin doesn't allow pausing when window is minimized/hidden


th-ch (approved):
Looks good, thanks for the contribution ✅ 
nit: PNGs can probably be optimized (using a tool like pngcrush) - merging as it is, will compress the images in a follow up!

PR #201, opened 2021-03-23
[Plugin] styled-bars

Araxeus:
I got bored of the old bars :)
This plugin changes the Titlebar, main menu, and scrollbar


plugins/styled-bars/back.js
```diff
@@ -0,0 +1,300 @@
+const {injectCSS} = require('../utils');
+const {Menu , app} = require('electron');
+const { existsSync } = require("fs");
+const path = require('path');
+const electronLocalshortcut = require("electron-localshortcut");
+const is = require('electron-is');
+const {getAllPlugins} = require('../../plugins/utils');
+const config = require('../../config');
+
+const pluginEnabledMenu = (plugin, label = "") => ({
```
th-ch: It would be better to re-use the `menu.js` file as much as possible (I'm afraid maintenance of 2 menus might be hard 😬 ) - even for `mainMenuTemplate`, we might be able to have a function that returns 2 versions (one for the "standard" menu, one for the custom one)

plugins/styled-bars/front.js
```diff
@@ -0,0 +1,23 @@
...
+	});
+	myBar.updateTitle(' ');
+	document.title = 'Youtube Music';
+
+	ipcRenderer.on('updateMenu', function (event, menu) {
+		if (menu) {
+			myBar.updateMenu(remote.Menu.getApplicationMenu());
+		} else {
+			try {
+				myBar.updateMenu(null);
```
th-ch, line 17: https://github.com/AlexTorresSk/custom-electron-titlebar#dispose does not work?

plugins/styled-bars/back.js
```diff
@@ -0,0 +1,300 @@
...
+		const menu = Menu.buildFromTemplate(template);
+		setMenu(menu);
+
+		//register keyboard shortcut && hide menu if hideMenu is enabled
+		if (config.get('options.hideMenu')) {
+			win.webContents.send('updateMenu', null); 
+			 var enabled=  false;
+			electronLocalshortcut.register(win, 'Esc', () => {
+				if(enabled) {
+					win.webContents.send('updateMenu', null);
```
th-ch: nit: we could put that in a util functions `updateMenu`/`deleteMenu` to avoid repeating the same code + send `false` instead of `null` for consistency

th-ch (commented):
Thanks for the contribution, love the idea! 🤩 

Left a few comments (main one is about re-using the existing menu to avoid duplicating code, otherwise it will make maintenance of the code harder - also, IMHO `styled-bars` feels a bit unclear as plugin name, I would have gone with something explicit like `in-app-menu`) - I am fine with merging it as it is and make the follow ups, or you can update the PR if you feel like it! It's up to you, in both cases this contribution is already awesome 👍 

Note: it does not work for me on MacOS but it might be the same issue as https://github.com/AlexTorresSk/custom-electron-titlebar/issues/145 (from the package)

plugins/styled-bars/front.js
Araxeus: I dont want to dispose the menu because I will have to recreate it from scratch instead of updating it
(also it will probably fuck up the html/css positioning)

plugins/styled-bars/back.js
Araxeus: Done :D
I actually started working on it earlier today 😝

Araxeus:
You are welcome to change the name to whatever you like :)
It was `styled-titlebar` at the beginning , but then I added the scrollbar so I changed the name 😂

It was tested only on windows but from what i've seen - mac users can use their own special titlebar ;)

About the comments on the code
- I changed the menu logic
It nows only fixes the existing template instead of creating a new one. 
Seems to work great for now but i will test drive it more

Also is it ok if I integrate a folder chooser for the `downloader` plugin if you aren't planning on doing that?

(to plugins/downloader/menu.js)

plugins/styled-bars/back.js
Araxeus: Yeah that's better, will commit soon [done]
Also, any other tips/complaints are welcome. Im just starting out with javascript & electron 😅

Araxeus:
I also added to the main menu a change that i had in my copied template :
Quit button + Restart button
I put them in the view submenu for now
Of course this has nothing to do with the plugin, you are welcome to disregard that commit if you want.
https://github.com/th-ch/youtube-music/pull/201/commits/09d2feb15bdcfaadeb94a648d1980088b8245cc9

plugins/styled-bars/back.js
```diff
@@ -0,0 +1,111 @@
...
+		}
+		//customize roles
+		else if (item.role != null) {
+			fixRoles(item)
+		}
+	}
+}
+
+//custom menu doesn't support roles, so they get injected manually
+function fixRoles(MenuItem) {
```
th-ch, line 78: Note: we could add options (like `noRoles`) to the `mainMenuTemplate` to reduce risks of breaking changes in the menu.

tray.js
```diff
@@ -64,6 +64,20 @@ module.exports.setUpTray = (app, win) => {
 				app.quit();
 			},
 		},
-	]);
+	];
+
+	// delete quit button from navigation submenu
+	let navigation = getIndex(template,'Navigation');
+	let quit = getIndex(template[navigation].submenu,'Quit App');
+	delete template[navigation].submenu[quit];
```
th-ch, line 72: Same note: we could have options to return a "tray specific menu"  in `mainMenuTemplate`

PR #215, opened 2021-03-31
In-app-menu plugin - rename plugin & configure menu builder

th-ch:
This PR follows up on https://github.com/th-ch/youtube-music/pull/201 by:
- renaming the plugin to `in-app-menu`
- tweaking the menu builder to accept options

cc @Araxeus who authored the plugin for testing that everything still works as expected


Araxeus:
Some things I notcied in this version of PR https://github.com/th-ch/youtube-music/pull/215/commits/c09de4315402fe0604def6bba41164a43d5884a4
* Roles could pretty much deleted at this point right? they don't seem to serve any purpose
* Tray Submenu's doesn't call `clickCb(item)` properly and because of that doesn't close on click (not sure why tho)
*  Plugin Menu doesn't use `clickCb `and in result doesn't update visual checkbox on click
* I think that in this state, imported menu's (from plugins) that use Checkbox/Radio/Roles won't work correctly, unless I'm missing something. - for example the new menu for notifications plugin use radio buttons - they wont close automatically
and if any plugin use checkbox in the future, it wont work 

In my opinion my original solution was quite nice because it didn't require special consideration and interaction with other parts of the application, with it you couldjust code a menu in a normal electron way and the plugin fix what it needs - on launch.

th-ch:
Updated the PR:
- Removed click callback, it made the code more complex and brought more cons than pros, reverted to the click override instead
- tray seems to work ok now - only drawback, updating plugins from the tray is not reflected in the custom menu, but it seems that was already the case before
- roles could indeed be deleted, I left like that for now but we could do without them

Araxeus:
Maybe the plugin + options submenu's should be removed from the tray menu - since it doesn't work properly, 
and also the tray menu is quite cluttered anyways and those options aren't that useful in tray mode

anyways it should probably be either fixed or removed, because its actually kind of a mess right now 😅

th-ch:
Merged master into the branch to get updates from https://github.com/th-ch/youtube-music/pull/217 - PR looks ok and should be ready to be merged.
About the tray menu, it can definitely be cleaned in a follow up PR, should be straightforward to implement!

PR #217, opened 2021-04-01
Refresh menu on plugin enable/disable (show/hide submenu)

Araxeus:
This allow users to change plugin state + its options, without having to restart the application just to see the options submenu

https://github.com/th-ch/youtube-music/pull/217/commits/adc0d145c320e745f7dc05baf59085741318fae3 fix styled-bars (old name) original function to support all menu changes in the application (not disregarding the new PR, just keeping compatibility with current main branch)

https://github.com/th-ch/youtube-music/pull/217/commits/02d45bed7489dd497d80aa6aad29b1422b2b1b9d  - refresh only if plugin has a menu.js (avoid needless menu refresh)

### DIFF
--- a/plugins/styled-bars/back.js
+++ b/plugins/styled-bars/back.js
@@ -22,7 +22,6 @@ let win;
 let done = false;
 
 module.exports = winImport => {	
-	//override menu template for custom menu
 	win = winImport;
 	// css for custom scrollbar + disable drag area(was causing bugs)
 	injectCSS(win.webContents, path.join(__dirname, 'style.css'));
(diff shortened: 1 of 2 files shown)


plugins/styled-bars/back.js
```diff
@@ -55,12 +55,9 @@ function switchMenuVisibility() {
...
 		let item = template[index];
-		//apply function on submenu
-		if (item.submenu != null) {
-			fixMenu(item.submenu);
-		}
-		//change onClick of checkbox+radio
-		else if (item.type === 'checkbox' || item.type === 'radio') {
+		//change onClick of checkbox+radio if not fixed
+		if ((item.type === 'checkbox' || item.type === 'radio') && !item.fixed) {
+			item.fixed = true;
```
Araxeus: Useless if everything works as expected - but guarantee that original onClick won't be overwritten more than once if somehow a submenu get checked again
[might be redundant]

th-ch (approved):
Thanks for the improvement! ✅

PR #218, opened 2021-04-01
remove 'shortcuts' from default plugins

### LINKED ISSUE #50: Display current song in volume bar
Currently when music plays and you increase the volume you only see the volume like this:
grafik

It would be nice if the currently playing song could be displayed there.
Groove Musik for example adds songs to there.
grafik

I don't know if this is possible with Linux or Mac but I know it's possible with Windows 10

### LINKED ISSUE #204: shortcuts plugin blocks native MediaSession api
see #50
After spending tons of hours researching the [`MediaSession`](https://googlechrome.github.io/samples/media-session/audio.html) API and way of implementation,
I realized Electron supported it all along, it just wasnt working because of the shortcuts plugin

Youtube Music even set it up automatically for each song, no need to do anything,,
⚠️ that includes media key controls, so => 
disabling the shortcuts plugin will still let you use mediakeys control + have native MediaSession support

[This has been tested on Windows && Linux with the help of helpful guys over at Electron]

So anyway I think you probably should atleast remove shortcuts from the **default** plugins, at this state it only *removes* features. - people that want to disable that overlay could just enable the plugin

Example screenshot to show MediaSession api working with youtube-music on windows

Araxeus:
Resolve #204 , Resolve #50 
shortcuts plugin currently blocks more features than it adds, so it probably shouldn't be enabled by default
* `shortcuts` block native MediaSession OS integration
* The media keys [buttons] controls works natively - without the plugin

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -21,21 +21,22 @@ const defaultConfig = {
 		navigation: {
 			enabled: true,
 		},
-		shortcuts: {
-			enabled: true,
-		},
 		adblocker: {
 			enabled: true,
 			cache: true,
 			additionalBlockLists: [], // Additional list of filters, e.g "https://raw.githubusercontent.com/uBlockOrigin/uAssets/master/filters/filters.txt"
 		},
 		// Disabled plugins
+		shortcuts: {
+			enabled: false,
+		},
 		downloader: {
 			enabled: false,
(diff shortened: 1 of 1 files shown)


th-ch (approved):
Thanks for the contribution!

PR #221, opened 2021-04-03
Allows downloading songs that aren't currently playing

### LINKED ISSUE #84: Doesn't download some music
when I want to download music without video I get this error:
Screen Shot 2020-12-06 at 10 53 04 PM

If I understand correctly youtube-dl is used to download, and in that case it is understandable why it doesn't work (because I think youtube-dl can only download music from youtube not youtube music, and most video-music on youtube music is also on youtube.
If that's the case, than this issue is for people who don't know that.

Anyway thanks for this great app.

### LINKED ISSUE #210: Download button not working.
when you right click any music or playlist and then the download button, appears this error box.

image

image

### LINKED ISSUE #219: Improve downloader
The url of the opened page is what's used to download songs with the downloader plugin. This is a problem because it means that the song needs to be "opened" to be able to download it. 
I often find my self trying to download a song from search results by clicking on the 3 dots menu on the side, only to find that the plugin is using the url of the search results and not the url of the song and therefore doesn't work.

Araxeus:
Fix #210 , Fix #219

This makes it so `Downloader` plugin use the URL of the song that the menu was opened from
(More accurately, it will use the URL of the radio mix for the song, taken from the same menu)

Also Fix #84 - plugin will no longer randomly download a playlist when trying to download a single song from the menu

This is also fully compatible with playlist downloading - effectively allowing users to download 'anything' without having to play it

### DIFF
--- a/plugins/downloader/front.js
+++ b/plugins/downloader/front.js
@@ -31,11 +31,16 @@ const reinit = () => {
 	}
 };
 
+const baseUrl = "https://music.youtube.com";
 // TODO: re-enable once contextIsolation is set to true
 // contextBridge.exposeInMainWorld("downloader", {
 // 	download: () => {
 global.download = () => {
-	const videoUrl = global.songInfo.url || window.location.href;
+	let videoUrl = getSongMenu()
+		.querySelector("ytmusic-menu-navigation-item-renderer")
+		.querySelector("#navigation-endpoint")
+		.getAttribute("href");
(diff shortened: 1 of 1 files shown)


plugins/downloader/front.js
```diff
@@ -31,11 +31,16 @@ const reinit = () => {
 	}
 };
 
+const baseUrl = "https://music.youtube.com";
```
th-ch, line 34: nit: could be imported from default config

th-ch (approved):

PR #222, opened 2021-04-03
Updated download plugin icon color to match other icons

Keyboardsheep:
Current download icon: 
image

With pull request:
image

The only thing that changed was adding `fill="#aaaaaa"`


th-ch (approved):
Thanks for the contribution! ✅

PR #224, opened 2021-04-03
Menu tweaks

Araxeus:
### Follows-up PR #215 
> commit order below is non-chronological, ordered for readability

- 2b3a20c **:**
  - Tray menu doesn't use mainTemplate anymore - added instead just a restart button 
🔽
That fixes the visual bug that cause desync between checkbox/radio buttons in tray menu and main app menu
(could be tweaked, but I'm not sure if any other buttons are needed there)

- 80a7d2c **:**
  - Remove roles from main menu template (since their original goal was to minimize code, but it ended up being double the length), it also obviously fixes compatibility issues with in-app-menu

- 5524a14 **:**
  - Declutter mainTemplate by moving advanced options to dedicated submenu, (also removed advanced options button from plugin submenu, since it already exists in another menu - and isn't directly related to plugins)

---

- a194046 **.**
  - using `win.once()` instead of `win.on` in multiple places fixes callbacks being called twice and frees up redundant listeners

- 2162052 **:**
  - Fixes bugs with css styling of in-app-menu and made scrollbar thumb gray to be more in theme
 
- 28d366a **Fix a critical bug ⚠️ :**
  it seems there is an unknown bug with in-app-menu causing did-fail-load to be called with [error code -3](https://source.chromium.org/chromium/chromium/src/+/master:net/base/net_error_list.h) , its seems to be some kind of false positive (it doesn't actually fail to load anything and also error code -3 means "ABORTED" )

so now it only loads `error.html` if not in this exact false-positive case (in-app-menu active + error code -3)
   > When a song started to play, `did-fail-load` was called - resulting in displaying error.html
   I don't understand why it happened it the first place, and how it works. but it's the only thing that fixed it

---

Sorry that it ended up being a long PR. I tried making it readable. 
If needed, I could split it


index.js
```diff
@@ -300,3 +322,28 @@ app.on("ready", () => {
...
+
+function showUnresponsiveDialog(win, details) {
+	if (!!details) {
+		console.log("Unresponsive Error!\n"+JSON.stringify(details, null, "\t"))
+	}
+	dialog.showMessageBox(win, {
+		type: "error",
+		title: "Window Unresponsive",
+		message: "The Application is Unresponsive",
+		details: "We are sorry for the inconveniance! please choose what to do:",
```
th-ch: nit: 
```suggestion
		details: "We are sorry for the inconvenience! please choose what to do:",
```

index.js
```diff
@@ -300,3 +322,28 @@ app.on("ready", () => {
...
+		cancelId: 0
+	}).then( result => {
+		switch (result.response) {
+			case 1: //if relaunch - relaunch+exit
+				app.relaunch();
+			case 2:
+				app.quit();
+				break;
+			case 0:
+			default:
```
th-ch: nit: if 0 has nothing special, no need to specify it:
```suggestion
			default:
```

index.js
```diff
@@ -2,6 +2,7 @@
 const path = require("path");
 
 const electron = require("electron");
+const  { dialog }  = require("electron");
```
th-ch: nit: electron being imported fully, dialog can be accessed with `electron.dialog` (or `const dialog = electron.dialog`)

tray.js
```diff
@@ -57,7 +56,13 @@ module.exports.setUpTray = (app, win) => {
 				win.show();
 			},
 		},
-		...mainMenuTemplate(win, true, true),
```
th-ch, line 60: How about adding a tray option to show/hide the menu? This way, when the option "Hide menu" is enabled, it is possible to show again the menu without editing the config (using the tray menu).

th-ch (approved):
Some small nits but otherwise looks good! Removing plugins from the tray menu is a breaking change but it simplifies the code and not sure it was very useful (it's still possible to show the app and the main menu when changing plugins).
Thanks for the PR, feel free to address the comments, otherwise I can merge it and make the changes as a follow up!

tray.js
Araxeus: The problem is that it wont be synced with the main menu checkbox. 
and you can already show the menu when its hidden with `alt` key (or `escape` with in-app-menu)

  Are you sure I should add it?

th-ch: Good point, should be good as is then!

plugins/taskbar-mediacontrol/back.js
```diff
@@ -2,57 +2,53 @@ const getSongControls = require('../../providers/song-controls');
...
+			click() { controls.next(win.webContents); }
+		}
+	]);
+}
+
 // Util
 function get(file) {
-	return path.join(__dirname,"assets", file);
+	return path.join(__dirname, "assets", file);
 }
```
Araxeus, line 54: This is actually the exact version in the main branch (no actual changes)
Was updated to solve messy merge conflict

index.js
```diff
@@ -152,13 +156,32 @@ function createMainWindow() {
...
+			errorCode,
+			errorDescription,
+			validatedURL,
+			isMainFrame,
+			frameProcessId,
+			frameRoutingId,
+		};
 		if (is.dev()) {
-			console.log("did fail load");
+			console.log(log.toString());
```
th-ch: Because `log` is an object, this will just log `[object Object]`, any reason not to call `JSON.stringify` directly? I also got `Error: Failed to serialize arguments` on line 184 so it might be better to pass it already stringified (it will end up serialized anyway!)

preload.js
```diff
@@ -32,7 +32,11 @@ document.addEventListener("DOMContentLoaded", () => {
 	// inject song-info provider
 	const songInfoProviderPath = path.join(__dirname, "providers", "song-info-front.js")
 	fileExists(songInfoProviderPath, require(songInfoProviderPath));
-	
+
+	// inject front logger
+	const loggerPath = path.join(__dirname, "providers", "front-logger.js")
+	fileExists(loggerPath, require(loggerPath));
```
th-ch: nit: contrary to plugins, we are sure this file exists so we can just require it directly!

providers/front-logger.js
```diff
@@ -0,0 +1,15 @@
+const { ipcRenderer } = require("electron");
+
+function logToString(log) {
```
th-ch, line 3: If we know it's either a log or a string, we can simplify it into `typeof log === "string" ? log : JSON.stringify(log, null, "\t")`

th-ch (commented):
Left a few additional comments after some testing, feel free to address them, otherwise I am fine merging as is and making changes as a follow up!

index.js
Araxeus: This actually a big oversight on my part, sorry about that..

Araxeus:
> Left a few additional comments after some testing, feel free to address them, otherwise I am fine merging as is and making changes as a follow up!

Thanks for all the feedback !

First error was kinda embarrassing, I don't know how I missed that 😅  
I think it should be ok now

th-ch (approved):
Looks good, thanks for the contribution, merging! ✅

PR #228, opened 2021-04-08
Interactive notifications for windows

Araxeus:
----

> Could be changed to 2nd option which might be better:

Unicode: [`\u{2770}`-previous]  [`\u{25B6}`-play] [`\u{2771}`-next]
(would need a subtle change of code since output for that play button [뿯▽] isn't the same as input for some reason)

----

The same lib could be used to setup interactive notification for mac, if anyone wants to work on that - [Node Notifier](https://www.npmjs.com/package/node-notifier)

- The standard pause button unicode wasn't used, because windows render those media buttons as blue

- Because of a bug with the delivery method, appID cannot be set without breaking the button functionality 
 (it makes returned `data` equal `undefined`)
so appID will it will always be the default "SnoreToast".

 - Song Icon is resized and cropped to keep original aspect ratio (resizing was really ugly with wide images before)

- Made `Notification Priority` menu item show only on linux (since it works only on linux)

- SongInfoProvider was parsing artist name based on video request , not on song - resulting in buggy artist names.
  changed it scrape artist name from playbar (native artist name)
  - could move functionality to notifications plugin instead of infoProvider
  - could be tweaked to wait a second on first run (first song always returns old VideoArtistName because playBar hasn't loaded by the time it queries it)

- `win.once()` instead of `win.on` fix multiple callbacks being registered

### DIFF
--- a/plugins/notifications/back.js
+++ b/plugins/notifications/back.js
@@ -3,7 +3,7 @@ const is = require("electron-is");
 const getSongInfo = require("../../providers/song-info");
 const { notificationImage } = require("./utils");
 
-const { setup, notifyInteractive } = require("./interactive")
+const { setupInteractive, notifyInteractive } = require("./interactive")
 
 const notify = (info, options) => {
 
@@ -24,9 +24,10 @@ const notify = (info, options) => {
 };
 
 module.exports = (win, options) => {
+	const isInteractive = is.windows() && options.interactive;
 	//setup interactive notifications for windows
-	if (is.windows()) {
-		setup(win);
+	if (isInteractive) {
+		setupInteractive(win);
 	}
 	const registerCallback = getSongInfo(win);
 	let oldNotification;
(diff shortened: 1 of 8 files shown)


plugins/notifications/back.js
```diff
@@ -39,13 +43,17 @@ module.exports = (win, options) => {
 				}
 				return;
 			}
-			// If url isn't the same as last one - send notification
+			// If url isn"t the same as last one - send notification
```
th-ch: nit: typo?

plugins/notifications/interactive.js
```diff
@@ -0,0 +1,92 @@
...
+    onPause = unpauseNotification;
+
+    win.webContents.once("closed", () => {
+        Delete()
+    });
+}
+
+//delete old notification
+let toDelete;
+function Delete() {
```
th-ch: nit: functions are generally written lower-case, capital case being kept for classes
```suggestion
function deleteNotification() {
```

th-ch (approved):
Looks ok, thanks for the contribution! 2 small nits (feel free to address them), otherwise it should be good to merge!

plugins/notifications/back.js
Araxeus: oh woupsie it must have happened when I replaced all `'` with `"` 😅

plugins/notifications/interactive.js
Araxeus: missed this one, thanks :)

th-ch (approved):

PR #231, opened 2021-04-12
Dependencies update

Araxeus:
- Removed `downloads-folder` dependency - use instead `app.getPath("downloads")`
- Updated *most* of the dependencies, notable important fixes:

  - Electron [11.4.2](https://github.com/electron/electron/releases/tag/v11.4.2) has [fixed ](https://github.com/electron/electron/pull/28366) `win.hide` breaking thumbar (because of my [feedback ](https://github.com/electron/electron/issues/28319) 😋 )
‎🔽 
 Updated `taskbar-mediacontrol` accordingly - it no longer overrides `win.hide()` + `win.show()`
+set taskbar on `win.show()` if songInfo is valid

  - YoutubeNonStop [v0.9.0](https://github.com/lawfx/YoutubeNonStop/releases/tag/v0.9.0) *probably* fixed a bug with `auto-confirm-when-paused` not allowing pausing through external buttons [Tray, Taskbar, Notifications]

### DIFF
--- a/plugins/downloader/utils.js
+++ b/plugins/downloader/utils.js
@@ -1,4 +1,4 @@
-const downloadsFolder = require("downloads-folder");
+const { app } = require("electron");
 
-module.exports.getFolder = (customFolder) => customFolder || downloadsFolder();
+module.exports.getFolder = (customFolder) => customFolder || app.getPath("downloads");
 module.exports.defaultMenuDownloadLabel = "Download playlist";
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
 		"browser-id3-writer": "^4.4.0",
 		"custom-electron-titlebar": "^3.2.6",
 		"discord-rpc": "^3.2.0",
-		"downloads-folder": "^3.0.1",
 		"electron-debug": "^3.2.0",
 		"electron-is": "^3.0.0",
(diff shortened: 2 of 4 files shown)


plugins/downloader/utils.js
```diff
@@ -1,4 +1,4 @@
-const downloadsFolder = require("downloads-folder");
+const { app } = require("electron");
 
-module.exports.getFolder = (customFolder) => customFolder || downloadsFolder();
+module.exports.getFolder = (customFolder) => customFolder || app.getPath("downloads");
```
th-ch, line 3: The `app` module from electron is only available in the [main process](https://github.com/electron/electron/blob/master/docs/api/app.md#app), not the renderer one, so it breaks the downloader - a [fix](https://stackoverflow.com/questions/47957479/how-to-use-electrons-app-getpath-to-store-data) for that is to use `(electron.app || electron.remote.app).getPath("downloads")`

th-ch (approved):
Looks good, thanks for the updates! Just one comment but the fix is simple, I will merge the PR then fix the `getFolder` method

plugins/downloader/utils.js
Araxeus: Woups sorry I missed that, good catch 🙂

PR #236, opened 2021-04-16
[Plugin] Precise volume control

Araxeus:
### `precise-volume` plugin

The original volume slider and shortcuts lets you control volume only in steps of 5% or 10%, 
this has always bothered me, there isn't enough control - especially around the lower percentages

So I made this small plugin to fix that (name could be changed):

### Features

* You can now have precise control over volume percentage by using mouse-wheel in play bar area (bottom bar)
each step add/subtract [1%] (default can be changed)

* Optional setting to change volume with local hotkey -  arrowkeys up/down (Like in native youtube)

* Advanced option in config to setup global hotkeys

* Advanced option to change volume steps (int 1-100 representing percentage change)
Will use prompt `type: "Counter"` to allow controlling this setting once the custom prompt from #224 gets implemented

* Plugin also makes the Volume slider + icon tooltip's always show the exact current volume percentage

* Precise volume is saved and restored between sessions

--- 

### Misc

To override the default mousewheel + arrowkey behavior of the volume bar, I had to do a trick before DOMContent loaded
(The only way to remove an eventListener created with an anonymous function is to catch it before its created)

Initially I just added the code to the main preload.js but later thought it would probably be better to enable all plugins to do some preload
so, I added an option for **Every Plugin** to add a preload.js to be executed from Renderer before DOMContent loaded

 ### Note on the *original* volume slider
 * Dragging it with the cursor still works the same way.
 * When changing the volume using this plugin - it gets **visually** updated only around every 5% volume
  For example, there is no visual difference between 5% and 7%
  (that's pretty much the main reason I added the tooltip with volume percentage)

I'm not too sure how to visually modify the slider itself without native youtube functionality breaking, if it's even possible

### DIFF
--- a/plugins/precise-volume/front.js
+++ b/plugins/precise-volume/front.js
@@ -1,6 +1,6 @@
 const { ipcRenderer } = require("electron");
 const { setOptions } = require("../../config/plugins");
- 
+
 function $(selector) { return document.querySelector(selector); }
 
 module.exports = (options) => {
@@ -54,7 +54,7 @@ function setPlaybarOnwheel(options) {
 function changeVolume(increase, options) {
 	// Need to change both the actual volume and the slider
 	const videoStream = $(".video-stream");
-
+	const slider = $("#volume-slider");
 	// Apply volume change if valid
 	const steps = options.steps / 100;
 	videoStream.volume = increase ?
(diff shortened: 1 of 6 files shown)


plugins/precise-volume/front.js
```diff
@@ -0,0 +1,207 @@
...
+	});
+}
+
+/** if (toIncrease = false) then volume decrease */
+function changeVolume(toIncrease, options) {
+	// Need to change both the actual volume and the slider
+	const videoStream = $(".video-stream");
+	const slider = $("#volume-slider");
+	// Apply volume change if valid
+	const steps = options.steps / 100;
```
th-ch: If I enable the plugin + `setArrowsShortcut`, `enabled` and `arrowsShortcut` are set to true but `options.steps` is undefined (it's in the defaults but there is no migration for existing configs), leading to an error -> we can have a defensive code here:
```suggestion
	const steps = (options.steps || 1) / 100;
```

th-ch (approved):
Looks mostly good and seems to work fine, thanks for the contribution! Just a comment to make the code defensive and then it should be good to merge 👍

th-ch (approved):

th-ch:
Looks ok, thanks for the contribution, merging! ✅

PR #243, opened 2021-04-28
Implement custom-electron-prompt

Araxeus:
## [**Custom Electron Prompt**](https://github.com/Araxeus/custom-electron-prompt#readme)

> I pretty much made this package to easily set settings in youtube-music

#### There are currently 4 available custom types:

- Input: Just a general text input field. 
 Currently used here for settings proxy option

- Keybind: Allow capturing single or multiple keybinds (`Electron Accelerator` format)
 Will be used for setting custom hotkeys in `precise-volume` and could also be implemented for `shortcuts` plugin

 -  Counter: Similar to input, but accept only numerical values and has [+] and [-] Buttons for ease of use
  Will be used for setting custom steps in `precise-volume`

  - Select: Create a dropdown select menu, currently no use in this app
  
----
 ### About the added Proxy Option
 - Currently just set the typed string as proxy option
 - Checbox Status is updated based on proxy status - `item.checked = input !== ""`
 - Could probably somehow add an input check to prevent locking users out of the application
  (As of now, if the proxy option is badly set - The app will crash on launch)

### DIFF
--- a/menu.js
+++ b/menu.js
@@ -6,6 +6,7 @@ const is = require("electron-is");
 
 const { getAllPlugins } = require("./plugins/utils");
 const config = require("./config");
+const prompt = require("custom-electron-prompt");
 
 const pluginEnabledMenu = (win, plugin, label = "", hasSubmenu = false) => ({
 	label: label || plugin,
@@ -103,30 +104,38 @@ const mainMenuTemplate = (win, withRoles = true, isTray = false) => [
 			},
 			...(is.windows() || is.linux()
 				? [
-						{
-							label: "Hide menu",
-							type: "checkbox",
-							checked: config.get("options.hideMenu"),
-							click: (item) => {
-								config.set("options.hideMenu", item.checked);
-							},
+					{
+						label: "Hide menu",
+						type: "checkbox",
+						checked: config.get("options.hideMenu"),
+						click: (item) => {
+							config.set("options.hideMenu", item.checked);
 						},
-				  ]
+					},
+				]
 				: []),
 			...(is.windows() || is.macOS()
 				? // Only works on Win/Mac
-				  // https://www.electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows
-				  [
-						{
-							label: "Start at login",
-							type: "checkbox",
-							checked: config.get("options.startAtLogin"),
-							click: (item) => {
-								config.set("options.startAtLogin", item.checked);
-							},
+				// https://www.electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows
+				[
+					{
+						label: "Start at login",
+						type: "checkbox",
+						checked: config.get("options.startAtLogin"),
+						click: (item) => {
+							config.set("options.startAtLogin", item.checked);
 						},
-				  ]
+					},
+				]
 				: []),
+			{
+				label: "Proxy",
+				type: "checkbox",
+				checked: !!config.get("options.proxy"),
+				click: (item) => {
+					setProxy(item, win);
+				}
+			},
 			{
 				label: "Tray",
 				submenu: [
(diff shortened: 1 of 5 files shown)


menu.js
```diff
@@ -300,3 +309,38 @@ module.exports.setApplicationMenu = (win) => {
...
+		type: 'input',
+		inputAttrs: {
+			type: 'url'
+		},
+		icon: iconPath,
+		customStylesheet: "dark",
+		width: 450,
+	};
+	//TODO: custom bar on prompt need testing on macOS
+	if (!is.macOS()) {
```
th-ch: For the record: tested (without the `if`) on MacOS, no difference (maybe due to the custom titlebar plugin not working properly)

menu.js
```diff
@@ -300,3 +309,38 @@ module.exports.setApplicationMenu = (win) => {
...
+		Object.assign(options, {
```
th-ch: nit: the more "modern JS" way:
```suggestion
		options = {
   		    ...options,
```

menu.js
```diff
@@ -300,3 +309,38 @@ module.exports.setApplicationMenu = (win) => {
...
+			frame: false,
+			customScript: path.join(__dirname, "plugins", "in-app-menu", "prompt-custom-titlebar.js"),
```
th-ch: We should probably only style the prompt with the custom title bar if the `in-app-menu` plugin is enabled

menu.js
```diff
@@ -300,3 +309,38 @@ module.exports.setApplicationMenu = (win) => {
...
 	Menu.setApplicationMenu(menu);
 };
+
+const iconPath = path.join(__dirname, "assets", "youtube-music-tray.png");
+const example = "Example: 'socks5://127.0.0.1:9999'";
+function setProxy(item, win) {
+	let options = {
+		title: 'Set Proxy',
+		label: 'Enter Proxy Address: (leave empty to disable)',
+		value: config.get("options.proxy") || example,
```
th-ch: Idea of feature for the prompt plugin: the ability to have a placeholder (would be better suited for `example` instead of putting it as value and checking later that `value !== example`)

menu.js
```diff
@@ -300,3 +309,38 @@ module.exports.setApplicationMenu = (win) => {
...
+	//TODO: custom bar on prompt need testing on macOS
+	if (!is.macOS()) {
+		Object.assign(options, {
+			frame: false,
+			customScript: path.join(__dirname, "plugins", "in-app-menu", "prompt-custom-titlebar.js"),
+			enableRemoteModule: true,			
+		});
+	}
+	prompt(options, win)
+		.then(output => {
```
th-ch: The promise can also be managed with async/await (feels simpler to read, and errors raised should be caught anyway by `electron-unhandled`)
```suggestion
		await prompt(options, win)
```

config/defaults.js
```diff
@@ -30,6 +30,7 @@ const defaultConfig = {
 		// Disabled plugins
 		shortcuts: {
 			enabled: false,
+			overrideMediaKeys: false,
```
th-ch, line 33: ~It would be better to default it to true for backwards-compatibility~
EDIT: given the issues mentioned in the PR description, ok to keep the change 👍

plugins/precise-volume/menu.js
```diff
@@ -1,19 +1,109 @@
...
+		]
+	};
+
+	setupPromptOptions(promptOptions);
+
+	prompt(promptOptions, win)
+		.then(output => {
+			if (output) {
+				for (const keybindObject of output) {
+					options.globalShortcuts[keybindObject.value] = keybindObject.accelerator;
```
th-ch: Similarly, doing it by destructuring options is easier to read (and does not require checking `if (output)`):
```suggestion
				options = {
				  ...options,
				  globalShortcuts: {
				     ...options.globalShortcuts,
				     ...output
				}   
```

th-ch (commented):
Left a few comments after a first review, looks promising 🤩 
nit: when the prompt window appears, the background is white, and it only gets black in a second time (and the change is noticeable), is it expected?

menu.js
Araxeus: Thats the way I did it initially, but I just think it looks massively better with the custom titlebar. (atleast on windows)
Also I don't really see any benefit in not enabling it (or drawbacks from enabling it)

menu.js
Araxeus: > For the record: tested (without the `if`) on MacOS, no difference

Ok thats weird but kinda expected since you said `in-app-menu` doesn't work on macOS

> when the prompt window appears, the background is white, and it only gets black in a second time (and the change is noticeable), is it expected?


sadly it is expected, since it :
 parse options + create a new electron window + load htmlPage + wait for `DOMContentLoaded` + **ipc.sendSync** to get the options to front + update and create DOM Elements accordingly.

which all sums up to a slower loading time

menu.js
Araxeus: there is already a placeholder which is the `value` field

> instead of putting it as value and checking later that value !== example

Are you talking about implementing this exact logic in `custom-electron-prompt` ?

I could make an boolean option `placeholder` which will mean "return null if `output=options.value`"
but it seems kinda unnecessary..

I think it might be best to let whoever implements the prompt decide what to do with the output, since there could be many different intentions when using the prompt,

plugins/precise-volume/menu.js
Araxeus: But output is in format `{ value: "volumeUp", accelerator: "Shift+A" }`
which isn't the same as options.globalShortcuts `{ volumeUp: "Shift+A" }`

I guess I could change the way the output work in the source level.
But only if you think that's really necessary, since people might be already using this library

> also this actually will make the code longer 😅

current version after destructuring:
```javascript
if (output) {
	for (const { value, accelerator } of output) {
		options.global[value] = accelerator;
	}
	setOption(options);
}
```
`if (output)` is here to make sure no code gets executed if prompt was canceled

menu.js
Araxeus: using spread operator now 👍🏼

menu.js
Araxeus: f910593 Switched to async await 👍🏼

plugins/shortcuts/back.js
```diff
@@ -19,31 +20,65 @@ function registerShortcuts(win, options) {
...
 
-		_registerGlobalShortcut(win.webContents, shortcut, songControls[action]);
-	});
-	(local || []).forEach(({ shortcut, action }) => {
-		console.debug("Registering local shortcut", shortcut, ":", action);
-		if (!action || !songControls[action]) {
-			console.warn("Invalid action", action);
-			return;
+/** Update options to new format if they are still an array (old format) */
+function updateOptions(options) {
```
th-ch: This looks like a one-shot operation that can be done with a [store migration](https://github.com/sindresorhus/electron-store#migrations) (there are examples in in `config/store.js`)

menu.js
th-ch: Yes, idea was to add a implement the placeholder logic/field in custom-electron-prompt! (main pro of a placeholder is to avoid having a fake value to remove when filling the field, contrary to a default value - here, it feels we have a placeholder used a default value 🙃)
Not a big deal though, does not block merging - just an improvement idea for the custom-prompt package :)

th-ch (commented):
For the `prompt-custom-titlebar`, looks ok to keep it all the time (but disabled for MacOS to avoid issues) but it should be moved to the `providers` folder in that case! (so that no `in-app-menu` code is executed if the plugin is disabled)

plugins/shortcuts/back.js
Araxeus: I moved the code to store migration in 002081b
Is it alright this way? should the version be higher?

(`prompt-custom-titlebar.js` was already moved to providers folder before 😝)

menu.js
Araxeus: 355f611 changed the input field "placeholder" attribute to the previous example as you suggested
(didn't need to change custom-electron-prompt itself)

I'm not sure its better though, since you can't copy paste or edit that text, and it disappears when typing. which can makes it harder for people that don't know the format

th-ch (approved):
Looks good, thanks for the contribution! Just need to fix the merge conflict and it should be good to merge! ✅

th-ch:
I am getting `TypeError: Cannot set property 'previous' of undefined` in `menu.js:45` when trying to register global song controls, not sure if it's expected?

Araxeus:
> I am getting `TypeError: Cannot set property 'previous' of undefined` in `shortcuts/menu.js:45` when trying to register global song controls, not sure if it's expected? (it seems the migration does not add the global/local keys in config when they are not present)

Indeed the object needs to be created before setting its properties 😅
should be fixed now

PR #252, opened 2021-05-05
Fix downloader metadata if not currently playing

### LINKED ISSUE #251: Max downloads from a playlist is 99?
I was downloading a playlist, and stopped when reached 99 downloads.
(Note: i tried downloading the same playlist 3 times and happened all the times. The playlist has 151 songs.)
image

Araxeus:
PR #221 allows downloading song that aren't playing, but the metadata wasn't captured - which also threw an error:
`"image.toPNG() is not a function"`. 
The same error was thrown once per song when downloading a playlist (since the metadata didn't contain a nativeImage)

* Error is avoided by checking if `metadata.image` exist before `writer.setFrame("APIC"...`

* Use `ytdl.getInfo(videoUrl)` to get metadata for songs that aren't playing (also for playlist-download items)
  this includes `metadata.image`

 * Allows downloading playlist of unlimited size if `options.playlistMaxItems` isn't specified. fixes #251

----

 `downloader-feedback` in `downloader/menu.js` results in constantly refreshing the menu 
```javascript
ipcMain.on("downloader-feedback", (_, feedback) => {
	downloadLabel = feedback;
	refreshMenu();
});
```
(Because every single song in the playlist sets a different label + refresh + restore original label + refresh)
This results in almost always crashing my application when I try to open the main menu while playlist is still downloading.

Instead uses the messageBox below To notify that playlist download started:

https://github.com/th-ch/youtube-music/blob/13fb686188b6a39c16e565e9b87bad14638f86ca/plugins/downloader/menu.js#L62-L68

And to update the download progress, 53bf7c5 adds a taskbar progress bar that gets updated by `chokidar` that watches the new playlist download directory

### DIFF
--- a/plugins/downloader/youtube-dl.js
+++ b/plugins/downloader/youtube-dl.js
@@ -184,19 +184,13 @@ module.exports = {
 ipcRenderer.on(
 	"downloader-download-playlist",
 	(_, url, playlistFolder, options) => {
-		const reinit = () =>
-			ipcRenderer.send("downloader-feedback", defaultMenuDownloadLabel);
-
 		downloadVideoToMP3(
 			url,
-			(feedback) => {
-				ipcRenderer.send("downloader-feedback", feedback);
-			},
+			() => {},
 			(error) => {
 				triggerAction(CHANNEL, ACTIONS.ERROR, error);
-				reinit();
 			},
-			reinit,
+			() => {},
 			options,
 			null,
 			playlistFolder
--- a/plugins/downloader/youtube-dl.js
+++ b/plugins/downloader/youtube-dl.js
@@ -152,7 +152,7 @@ const toMP3 = async (
 		ipcRenderer.send("add-metadata", filePath, fileBuffer, {
 			artist: metadata.artist,
 			title: metadata.title,
-			imageSrc: metadata.imageSrc || ""
+			imageSrc: metadata.imageSrc
 		});
 		ipcRenderer.once("add-metadata-done", reinit);
 	} catch (e) {
--- a/plugins/downloader/youtube-dl.js
+++ b/plugins/downloader/youtube-dl.js
@@ -24,7 +24,14 @@ const ffmpeg = createFFmpeg({
 });
 const ffmpegMutex = new Mutex();
 
-const downloadVideoToMP3 = (
+function noTopic(channelName) {
+	if (channelName && channelName.endsWith(" - Topic")) {
+		channelName = channelName.slice(0, -8);
+	}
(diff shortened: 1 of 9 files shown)


plugins/downloader/youtube-dl.js
```diff
@@ -24,7 +24,14 @@ const ffmpeg = createFFmpeg({
 });
 const ffmpegMutex = new Mutex();
 
-const downloadVideoToMP3 = (
+function noTopic(channelName) {
```
th-ch: Looks similar to what was done in the last.FM plugin (`cleanupArtistName` - other suffixes like `VEVO` can be removed too) and very useful! What do you think of moving that logic into the "song info" provider (and removing it from downloader/last.fm plugins) so that all plugins can benefit from it?

th-ch (approved):
Looks good, thanks for the improvement! Just a remark about sanitizing metadata which could be done in the provider instead of being done in the different plugins - I'm also fine to merge as is and improve it later!

plugins/downloader/youtube-dl.js
Araxeus: Great idea, I just implemented exactly what you said (one function in song-info, delete suffix `' - Topic' || 'VEVO'`)

> even though all this really only matter when downloading music that isn't playing (doesn't use song-info)
since song-info will 99% of the time return either an already formatted artist name from playbar)
https://github.com/th-ch/youtube-music/blob/8b471c07726f0b8317469b3b9426c860b088e47a/providers/song-info.js#L61

semvis123: Looks good to me, but I think that having the variable available in advanced settings could be useful (instead of hardcoding it).
btw. I wouldn't mix none semicolon with one semicolon in the same file, but I'm sure this wasn't intended.

Araxeus: I don't really see the point in allowing users to change that list, since it will almost never have any effect anyways..
its pretty much just a last resort safeguard at this point

But what did you mean by mixing semicolons? just what was fixed in 1140c3e or something else?

th-ch (approved):

plugins/downloader/youtube-dl.js
semvis123: Okay I agree, and if someone wants to add a suffix to remove they can just change the source. I just meant that small difference fixed by linting :)

th-ch: Didn't see the ongoing discussion before merging 🙈
Things look good but if needed, we can make a follow up PR!

PR #253, opened 2021-05-06
Fix precise volume listener override

Araxeus:
Primarily fix youtube-music search bar not registering keyboard inputs.
should also fix other interactions


th-ch (approved):
Looks ok, thanks for the fix! ✅

PR #259, opened 2021-05-06
fix playPause bugs by directly playPause video element

### LINKED ISSUE #129: Media Keys Broken with Auto-Unpause
When the plugin auto-confirm-when-paused is enabled, after a while the pause media key won't work anymore (when you push it nothing happens). I suspect this is because the app immediately unpauses thinking it's been paused by youtube. Pushing the buttons on the app still seem to work though

### LINKED ISSUE #145: Play/Pause menu item in taskbar icon Context Menu doesn't pause

Araxeus:
this PR makes song-controls.playPause() affect directly the video element in the page. (instead of just pressing space)
This fixes several bugs when pausing the video when it isn't focused
(from tray/taskbar/notifications)

fix #145 , fix #129

also sneaky change because of previous comments:
song-info-front is directly required instead of checking if file exist (since we already know it exist)

### DIFF
--- a/providers/song-controls-front.js
+++ b/providers/song-controls-front.js
@@ -10,7 +10,11 @@ module.exports = () => {
         if (videoStream.paused) {
             videoStream.play();
         } else {
-            videoStream.yns_pause();
+            if (videoStream.yns_pause) {
+                videoStream.yns_pause();
+            } else {
+                videoStream.pause();
+            }
         }
     });
 };
--- a/providers/song-controls-front.js
+++ b/providers/song-controls-front.js
@@ -1,9 +1,7 @@
 const { ipcRenderer } = require("electron");
 
-let videoStream;
+let videoStream = document.querySelector(".video-stream");
 module.exports = () => {
-    videoStream = document.querySelector(".video-stream");
-
     ipcRenderer.on("playPause", () => {
         if (!videoStream) {
             videoStream = document.querySelector(".video-stream");
--- a/providers/song-controls-front.js
+++ b/providers/song-controls-front.js
@@ -10,11 +10,9 @@ module.exports = () => {
         if (videoStream.paused) {
             videoStream.play();
         } else {
-            if (videoStream.yns_pause) {
-                videoStream.yns_pause();
-            } else {
+            videoStream.yns_pause ?
+                videoStream.yns_pause() :
                 videoStream.pause();
-            }
         }
     });
 };
--- a/providers/song-controls-front.js
+++ b/providers/song-controls-front.js
@@ -1,0 +1,18 @@
+const { ipcRenderer } = require("electron");
+
+let videoStream;
+module.exports = () => {
+    videoStream = document.querySelector(".video-stream");
+
+    ipcRenderer.on("playPause", () => {
+        if (!videoStream) {
+            videoStream = document.querySelector(".video-stream");
+        }
+
+        if (videoStream.paused) {
+            videoStream.play();
+        } else {
+            videoStream.yns_pause();
+        }
+    });
+}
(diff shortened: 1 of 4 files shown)


providers/song-info-front.js
```diff
@@ -10,17 +10,16 @@ ipcRenderer.on("update-song-info", async (_, extractedSongInfo) => {
...
 		realXHR.addEventListener(
 			"readystatechange",
 			() => {
-				if (realXHR.readyState == 4 && realXHR.status == 200) {
-					if (realXHR.responseURL.includes("/player")) {
+				if (realXHR.readyState === 4 && realXHR.status === 200
+					&& realXHR.responseURL.includes("/player")) {
 						// if the request contains the song info, send the response to ipcMain
 						ipcRenderer.send("song-info-request", realXHR.responseText);
-					}
```
Araxeus, line 23: this is a pretty random lint but xo was warning me about those so i decided why not

providers/song-controls-front.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = () => {
+    ipcRenderer.on("playPause", () => {
+        if (!videoStream) {
+            videoStream = document.querySelector(".video-stream");
+        }
+
+        if (videoStream.paused) {
+            videoStream.play();
+        } else {
+            videoStream.yns_pause();
```
th-ch: `Uncaught TypeError: videoStream.yns_pause is not a function` shouldn't it be
```suggestion
            videoStream.pause();
```
?

th-ch (commented):
The `auto-confirm-when-paused` unpauses the video each time because it is paused programmatically (instead of emulating the space key), is it expected?

providers/song-controls-front.js
Araxeus: ~~Thats weird... where did you get that error?
`pause()` doesn't work on the youtube video stream.
`yns_pause()` is a function that I extracted from the actual youtube video stream and it works in all my tests~~

~~I didn't test it with `auto-confirm-when-paused`, we'll have to see what to do about that~~

~~maybe use this new ipc("playPause") only if `auto-confirm-when-paused` isn't enabled 
and else use spacebar as before. but then it wont fix this issue for users that have that plugin enabled :(~~

th-ch: Maybe it depends on the platform 🤔  -> A workaround can be to check if `yns_pause` is an existing function, in which case it can be called, otherwise call `pause`.
Agreed about using it only if `auto-confirm-when-paused` isn't enabled, it's not perfect but this way it won't break stuff!

Araxeus: I just realized 2 things:
1. `auto-confirm-when-paused` was enabled for me all along while testing
2.  `pause()` works only when YoutubeNonStop isn't active (because YoutubeNonStop overrides the original function)
 `yns_pause()` is the original `pause()` function that YoutubeNonStop saved and registered

   

source: [YoutubeNonStop.overrideVideoPause()](https://github.com/lawfx/YoutubeNonStop/blob/7b6b97b31bb3fd2078179660db0fd3fcc7062259/autoconfirm.js#L115-L133)


With the latest changes, it should all work flawlessly for you right?

Araxeus: I Tested this with `auto-confirm-when-paused` and it seems to work flawlessly for me.

Is it maybe possible you didn't update your locally deployed version of `YoutubeNonStop`?
If not - I'm really confused as to how it can be that it works for me but not for you... platform difference again??

th-ch (approved):
Looks ok after some local testing ✅ 
Thanks for the improvement, merging!

PR #265, opened 2021-05-08
Downloader tweaks + taskbar progress bar

Araxeus:
* Use metadata.image from song-info provider only if not sending imgUrl from front (fixes out of sync artwork)
  https://github.com/th-ch/youtube-music/blob/6b88397f820eee0d94cbc4f4bef7f420f1ac2468/plugins/downloader/back.js#L51-L56
* Gets the best artwork possible using `UrlToJPG` + `cropMaxWidth` (in [downloader/utils](https://github.com/th-ch/youtube-music/blob/3831e61d106b5f9cfa61971ba3ca011a70b7bf3b/plugins/downloader/utils.js))
  (`UrlToJPG` is needed because `videoDetails?.thumbnails[-1]` is not always presented in jpg format (webp throw errors) + can be of different sizes. so this function guarantee the highest resolution of artwork)
  (`cropMaxWidth` gets pixel perfect cropping for official music artwork, without it the artwork usually have just big margins filled with solid color)
* Dodge possible thrown errors + more accurate query selector for the "start radio" button in the popup menu:
  https://github.com/th-ch/youtube-music/blob/a8ac2c3af988f299be85010e7fea541096b7e261/plugins/downloader/front.js#L44-L47
* Adds a taskbar progress bar when downloading single songs, this allows seeing download progress without the submenu open (already implemented for playlists)

### DIFF
--- a/plugins/downloader/utils.js
+++ b/plugins/downloader/utils.js
@@ -3,3 +3,33 @@ const electron = require("electron");
 module.exports.getFolder = (customFolder) =>
 	customFolder || (electron.app || electron.remote.app).getPath("downloads");
 module.exports.defaultMenuDownloadLabel = "Download playlist";
+
+module.exports.UrlToJPG = (imgUrl, videoId) => {
+	if (!imgUrl || imgUrl.includes(".jpg")) return imgUrl;
+	if (imgUrl.includes("maxresdefault")) {
+		return "https://img.youtube.com/vi/"+videoId+"/maxresdefault.jpg";
+	}
+	if (imgUrl.includes("hqdefault")) {
+		return "https://img.youtube.com/vi/"+videoId+"/hqdefault.jpg";
+	} //it will almost never get further than hq
+	if (imgUrl.includes("mqdefault")) {
+		return "https://img.youtube.com/vi/"+videoId+"/mqdefault.jpg";
+	}
+	if (imgUrl.includes("sdddefault")) {
+		return "https://img.youtube.com/vi/"+videoId+"/sdddefault.jpg";
+	}
+	return "https://img.youtube.com/vi/"+videoId+"/default.jpg";
+}
+
+module.exports.cropMaxWidth = (image) => {
+	const imageSize = image.getSize();
+	if (imageSize.width === 1280 && imageSize.height === 720) {
+		return image.crop({
+			x: 280,
+			y: 0,
+			width: 720,
+			height: 720
+		});
+	}
+	return image;
+}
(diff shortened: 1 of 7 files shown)


providers/song-info.js
```diff
@@ -57,8 +53,8 @@ const songInfo = {
 
 const handleData = async (responseText, win) => {
 	let data = JSON.parse(responseText);
-	songInfo.title = data.videoDetails?.media?.song || data?.videoDetails?.title;
-	songInfo.artist = data.videoDetails?.media?.artist || await getArtist(win) || cleanupArtistName(data?.videoDetails?.author);
```
Araxeus, line 61: `videoDetails.media` can only appear in data parsed from `ytpl.getInfo()`
(never from XHR, I shouldn't have added it here in the first place)

plugins/downloader/utils.js
```diff
@@ -3,3 +3,29 @@ const electron = require("electron");
...
+module.exports.UrlToJPG = (imgUrl, videoId) => {
+	if (!imgUrl || imgUrl.includes(".jpg")) return imgUrl;
+	//it will almost never get further than hqdefault
+	for (const quality of orderedQualityList) {
+		if (imgUrl.includes(quality)) {
+			return `https://img.youtube.com/vi/${videoId}/${quality}.jpg`;
+		}
+	}
+	return `https://img.youtube.com/vi/${videoId}/default.jpg`;
+}
```
Araxeus, line 17: this function will 99.9% of the time:
return a `maxresdefault` webp converted to jpg (quite frequent with newer videos)
or return the original url (is jpg) 

0.1% chance (might even be 0%)
to do more than 1 iteration over qualityList (will only happen if original doesn't contains `.jpg` or `maxresdefault`)

This was mostly coded this way to **100% guarantee** a jpg url output

> note that this function is only applicable to thumbnails from `ytpl.getInfo` and not from `XMLHttpRequest` (song-info provider), since they are stored in a different source and in completely different formats

plugins/downloader/utils.js
```diff
@@ -3,3 +3,29 @@ const electron = require("electron");
 module.exports.getFolder = (customFolder) =>
 	customFolder || (electron.app || electron.remote.app).getPath("downloads");
 module.exports.defaultMenuDownloadLabel = "Download playlist";
+
+const orderedQualityList = ["maxresdefault", "hqdefault", "mqdefault", "sdddefault"];
+module.exports.UrlToJPG = (imgUrl, videoId) => {
```
th-ch, line 8: nit:
```suggestion
module.exports.urlToJPG = (imgUrl, videoId) => {
```

th-ch (approved):

th-ch:
Just a nit but nothing blocking, otherwise looks good, thanks for the improvement, merging! ✅

PR #267, opened 2021-05-09
fix last-fm overwrite config on each start

Araxeus:
`config.suffixesToRemove` was removed, but this check was left in the code by accident


th-ch (approved):
Thanks for the fix! ✅

PR #269, opened 2021-05-10
fix song-info callback duplication

Araxeus:
## Old behavior
Every plugin had to setup SongInfo separately

Which causes duplication of listener registration `win.on("page-title-updated"...` and `ipcMain.on("song-info-request"...`

Which causes song-info callbacks to be called multiple times 
(because ipcMain callbacks were registered X times and sent X events for each "song-info-request")

## New behavior

songInfo is setup **once** from index.js before plugin load.
Each plugin can register a song-info callback by just requiring it. (without the need to set it up)
> achieved by separating `registerProvider()` and `registerCallback()` in song-info.js

registerCallback example:
```javascript
const registerCallback = require("../../providers/song-info");
registerCallback(songInfo => {
	console.log(songInfo.title);
})
```
> This fix also warranted a small refactor of notifications plugin (which was the most heavily affected by this bug)

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const { setApplicationMenu } = require("./menu");
 const { fileExists, injectCSS } = require("./plugins/utils");
 const { isTesting } = require("./utils/testing");
 const { setUpTray } = require("./tray");
+const { setupSongInfo } = require("./providers/song-info");
 
 // Catch errors and log them
 unhandled({
@@ -157,6 +158,7 @@ function createMainWindow() {
 }
 
 app.once("browser-window-created", (event, win) => {
+	setupSongInfo(win);
 	loadPlugins(win);
 
 	win.webContents.on("did-fail-load", (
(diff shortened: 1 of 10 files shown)


th-ch (commented):
Change looks good overall but is it expected that the downloader menu import is not updated as well?
https://github.com/th-ch/youtube-music/blob/master/plugins/downloader/menu.js#L11
(I get `TypeError: registerCallback is not a function` when launching the app)

Araxeus:
> Thanks for the fix! Change looks good overall but is it expected that the downloader menu import is not updated as well?
> [`master`/plugins/downloader/menu.js#L11](https://github.com/th-ch/youtube-music/blob/master/plugins/downloader/menu.js#L11)
> (I get `TypeError: registerCallback is not a function` when launching the app)

Sorry I don't know how I missed this one. it should be fine now

th-ch (approved):
Looks good, thanks for the improvement!

PR #270, opened 2021-05-10
fix notification showing appID instead of app name on windows

Araxeus:
## Before:
appID Unset
## After:
appID Fixed

This sets the appID +
if not in dev mode OR running from portable app -> check the windows startMenu shortcut and fix it if necessary, 
(Since this fix works only if shortcut is correctly registered in startMenu)

> This only fix **normal** notifications. fix for interactive notifications is [coming soon](https://github.com/mikaelbr/node-notifier/pull/375)

### DIFF
--- a/index.js
+++ b/index.js
@@ -255,6 +255,10 @@ app.on("ready", () => {
 		}, 20000);
 	}
 
+	if (is.windows()) {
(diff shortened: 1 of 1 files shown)


th-ch (approved):
Code change looks ok, thanks for the improvement! Merging ✅

PR #273, opened 2021-05-11
Upgrade electron to v12

th-ch:
This PR bumps electron to the latest version (v12).


th-ch:
App looks ok - remote module is deprecated but still there (it will be removed in a future version) - migrating to `@electron/remote` is not necessary for now but will have to be done at some point - merging

PR #277, opened 2021-05-13
[Need Feedback] check proxy before launching with it

Araxeus:
This PR aims to stop the application from crashing when `options.proxy` is set to a bad address
(shows a message on launch if it can't connect to proxy address)

This really need some testing before its merged.

Can anyone that have a proxy server test this to see if it actually works? 😅

> Its important that you test first without those changes to check that you can connect using your valid proxy address, and then see that this changes don't block that proxy

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,6 +40,9 @@ if (config.get("options.disableHardwareAcceleration")) {
 	app.disableHardwareAcceleration();
 }
 
+if (config.get("options.proxy")) {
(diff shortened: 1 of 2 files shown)


Araxeus:
Got feedback, this version doesn't work, also the whole thing is probably not needed since If you specify a bad proxy - you can now actually just change it easily in the menu and restart the app

PR #288, opened 2021-05-17
Audio compressor plugin

Thymue:
This plugin adds [DynamicsCompressorNode](https://developer.mozilla.org/en-US/docs/Web/API/DynamicsCompressorNode) onto the audio that is being played which should help with the case where there are two videos with different volume in the video's audio itself.

Example:
[vid1 (which is loud) compared to](https://music.youtube.com/watch?v=vz7EQTtzPo8&t=52)
[vid2](https://music.youtube.com/watch?v=n-4SVb8ZoIU&t=70)
without the plugin:
audacity screenshot of both without the plugin
with the plugin:
audacity screenshot of both with the plugin

### DIFF
--- a/plugins/audio-compressor/front.js
+++ b/plugins/audio-compressor/front.js
@@ -1,0 +1,32 @@
+const {
+	watchDOMElement
+} = require("../../providers/dom-elements");
+
+let videoElement;
+
+const applyCompressor = () => {
+	var audioContext = new AudioContext();
+	
+	var compressor = audioContext.createDynamicsCompressor();
+	compressor.threshold.value = -50;
+	compressor.ratio.value = 12;
+	compressor.knee.value = 40;
+	compressor.attack.value = 0;
+	compressor.release.value = 0.25;
+
+	var source = audioContext.createMediaElementSource(videoElement);
+
+	source.connect(compressor);
+	compressor.connect(audioContext.destination);
+};
+
+module.exports = () => {
+	watchDOMElement(
+		"video",
+		(document) => document.querySelector("video"),
+		(element) => {
+			videoElement = element;
+			applyCompressor();
+		}
+	);
+};


Araxeus:
It looks like you are watching any changes in the video element, including volume change / pause etc
is this actually necessary?

or can this be applied only once on load?
```js
document.addEventListener("DOMContentLoaded", ()=> applyCompressor());
```
if that doesn't work then maybe only when each song load?
```js
video.addEventListener("canplay" || "loadeddata", ()=> applyCompressor(this));

const applyCompressor = (videoElement) => { ... }
```

anyways nice job 🙂

Thymue:
> you should know that `front.js` is called after DOMContentLoaded

image
shouldn't this work then? I'm not very experienced so correct me if I'm doing something wrong.
but regardless, you're right that using `watchDOMElement` is probably unnecessary. I'll look into some other solution.

Araxeus:
You're right - DOMContentLoaded isn't very reliable in our case
It does actually takes a few miliseconds before the video is initially ready
the way i fixed it in `precise-volume` is waiting for `did-finish-load` and then having a timeout of 500milisecs if video is undefined.

if you end up using this, we should probably just send the did-finish-load ipc event from index.js
(its currently being sent only from precise-volume/back.js)

```js
//minimize the amount of times timeout is set
ipcRender.on("did-finish-load", checkVideoLoaded);

function checkVideoLoaded() {
	if (document.querySelector("video")) {
		ItsLoaded();
	} else {
		setTimeout(checkVideoLoaded, 500);
	}
}
```

Araxeus:
or the even shorter version that im using in the new Chromecast plugin:
```js
module.exports = function checkVideoLoaded() {
    const video = document.querySelector("video");
    video ? setup(video) : setTimeout(checkVideoLoaded, 500);
}
```

Thymue:
I personally don't like using `setTimeout` as a solution but it's certainly better than using `watchDOMElement` and creating unnecessary function calls.

Araxeus:
And I dont like watchDOMElement when you dont care about the changes 😝
Speaking of, I think there is a memory leak playback-speed/front because of it
https://github.com/th-ch/youtube-music/blob/61e7124516571a63c7155937e827ba95dc34b9cb/plugins/playback-speed/front.js#L43-L80

First dom watcher is just unnecessary, 
but the second? it looks like its creates a new mutation observer every time there is a change in the document?

I hope i'm reading it wrong somehow..

PR #296, opened 2021-05-23
[DEPRECATED] Chromecast Plugin

### LINKED ISSUE #289: [Feature Request] Chromecast support
Cast as YouTube Music (as smartphone version)
Or cast as "tab" (as Chrome/Edge etc.)

Araxeus:
Implement and closes #289

This plugin connects to all nearby Chromecast devices and stream youtube music with video

Sync Options: 
* Seek - sync current time on Chromecast when current time is changed in local video
* StartTime - wait for Chromecast to start playback before local playback begins
* Volume - self explanatory

Features Tested by @playday3008 - who originally requested this plugin

> I can't see why snyk failed the security test (access error - not a member of organization)

### DIFF
--- a/plugins/chromecast/back.js
+++ b/plugins/chromecast/back.js
@@ -11,7 +11,7 @@ module.exports = () => {
 function registerDevice(device) {
     let currentUrl;
     let isPaused;
-
+    console.log(`Registered a new device: ${device.friendlyName}`)
     registerCallback(songInfo => {
         if (!songInfo?.title) {
             return;
@@ -29,6 +29,7 @@ function registerDevice(device) {
             isPaused ?
                 device.pause() :
                 device.resume();
+            console.log(isPaused ? "Paused" : "Resumed" + "palyback on your chromecast device")
         }
 	});
 }
--- a/plugins/chromecast/back.js
+++ b/plugins/chromecast/back.js
@@ -1,4 +1,6 @@
 const getSongInfo = require("../../providers/song-info");
+const getSongControls = require('../../providers/song-controls');
+
 const { ipcMain } = require('electron')
 const ChromecastAPI = require('chromecast-api');
 const { setOptions } = require("../../config/plugins");
@@ -10,10 +12,16 @@ let registerCallback;
 
 let win;
 
+let play, pause;
+
 let options;
 
 module.exports = (winImport, initialOptions) => {
     win = winImport;
+    const { playPause } = getSongControls(win);
+    play = () => playPause(true);
+    pause = () => playPause(false);
+
     options = initialOptions;
     registerCallback = getSongInfo(win);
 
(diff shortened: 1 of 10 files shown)


plugins/chromecast/back.js
```diff
@@ -0,0 +1,105 @@
...
+                device.resume();
+            }
+        }
+    });
+}
+
+function transformURL(url) {// will not be needed after https://github.com/alxhotel/chromecast-api/pull/69 (chromecastAPI v0.3.5)
+    const videoId = url.match(/(?:http(?:s?):\/\/)?(?:www\.)?(?:music\.)?youtu(?:be\.com\/watch\?v=|\.be\/)([\w\-\_]*)/);
+    // videoId[1] should always be valid since regex should always be valid
+    return "https://youtube.com/watch?v=" + (videoId.length > 1 ? videoId[1] : "dQw4w9WgXcQ");
```
th-ch, line 89: Nice Rickroll 🙃 Maybe worth a comment in code, I wondered why this hardcoded ID was there at first 😅

providers/song-controls.js
```diff
@@ -12,8 +12,7 @@ module.exports = (win) => {
 		// Playback
 		previous: () => pressKey(win, "k"),
 		next: () => pressKey(win, "j"),
-		playPause: () => win.webContents.send("playPause"),
-		like: () => pressKey(win, "_"),
+		playPause: (toPlay = undefined) => win.webContents.send("playPause", toPlay),
```
th-ch: I'm getting a `Error: Failed to serialize arguments` (`song-controls.js:15:54`) when using the play/pause media key, is it expected?

providers/song-controls-front.js
```diff
@@ -1,18 +1,43 @@
...
+            break;
+        case false:
+            pause();
+    }
+
+    function play() {
+        video.play();
+    }
+    
+    function pause() {
```
th-ch: nit: it might not be necessary to define the 2 inner functions if they are a one-liner and only used once!

providers/song-controls-front.js
```diff
@@ -1,18 +1,43 @@
...
+
+function checkVideo() {
+    if (!video) {
+        video = document.querySelector("video");
+        if (!video) {
+            return false;
         }
-    });
-};
+    } return true;
```
th-ch: nit: it can be simplified into
```js
function checkVideo() {
    if (!video) {
        video = document.querySelector("video");
    }

    return !!video;
}
```

providers/song-controls.js
```diff
@@ -12,8 +12,7 @@ module.exports = (win) => {
 		// Playback
 		previous: () => pressKey(win, "k"),
 		next: () => pressKey(win, "j"),
-		playPause: () => win.webContents.send("playPause"),
-		like: () => pressKey(win, "_"),
```
th-ch: Is it expected this one is removed?

providers/song-controls-front.js
```diff
@@ -1,18 +1,43 @@
...
-            videoStream.play();
-        } else {
-            videoStream.yns_pause ?
-                videoStream.yns_pause() :
-                videoStream.pause();
+function playPause(toPlay = undefined) {
+    if (!checkVideo()) return;
+    
+    switch(toPlay) {
+        case undefined:
```
th-ch: nit: could also be the default case for the switch, to be defensive

th-ch (commented):
Thanks for the addition! Did not test it with a real chromecast device yet, but looks promising! Left a few first comments - about the security issue reported by Snyk, `chromecast-api@0.3.4` introduces `dns-packet@4.2.0` which has a [`Remote Memory Exposure` vulnerability](https://snyk.io/vuln/SNYK-JS-DNSPACKET-1293563), fixed in `dns-packet@5.2.4`. Until the chromecast package has an update for this, the version can be forced with a resolution `"dns-packet": "5.2.4",` in package.json:
https://github.com/th-ch/youtube-music/blob/1fc8f7ad7eeda8dd342f910e6df4a859c96fcd60/package.json#L101

_(note that it's not the same major version so worth manually testing that everything works correctly!)_

providers/song-controls-front.js
Araxeus: done 👍🏼

providers/song-controls.js
Araxeus: It actually has nothing to do with this pull - but the `like` function was written twice - why's that?
https://github.com/th-ch/youtube-music/blob/e0cb1326868594d172668712da8581b893c925ec/providers/song-controls.js#L28
https://github.com/th-ch/youtube-music/blob/e0cb1326868594d172668712da8581b893c925ec/providers/song-controls.js#L16

I just assumed it was a mistake since the first one works - so I removed the second one

providers/song-controls-front.js
Araxeus: done 👍🏼

providers/song-controls-front.js
Araxeus: I agree that `play()` is unnecessary, but `pause()` is called in 2 different places

(I originally added `play()` only because `pause()` was there - but I removed it as requested)

providers/song-controls.js
Araxeus: It happened because shortcut plugin call the method with `win.webContents` argument.
I just fixed it by making `play, pause` directly part of song controls
sorry about that

plugins/chromecast/back.js
Araxeus: There is a comment already 😅
I thought the module would update its version faster - and this method will be unneeded
but sadly it hasn't happened yet - the fix pr was merged but the npm version wasn't updated yet

(method should only be called when songInfo is updated which means the url regex is valid - which means you should never get rick-rolled 😝)

Araxeus:
> _(note that it's not the same major version so worth manually testing that everything works correctly!)_

Tested with latest changes

playday3008:
Well, we still need device list with optional auto connection

danir-de:
This feature would make the app feature complete for me, sadly I cannot help by coding. If there is anything I can do to help to refactor this for the latest version, please let me know!

PR #318, opened 2021-06-09
fix hidden webp thumbnail throwing MIME type error in downloader

### LINKED ISSUE #315: Error: Unknown picture MIME type
Hello,
when downloading music I came across an error.
The error says: Error: Unknown picture MIME type
pic
I am running the lastest version of this app.

Araxeus:
Fix #315

## The Problem

youtube sometimes hide the real picture filetype, for example in this link:

https://i.ytimg.com/vi/tmntIJSf0Dw/hqdefault.jpg?sqp=-oaymwEcCNACELwBSFXyq4qpAw4IARUAAIhCGAFwAcABBg==&rs=AOn4CLC5b9hkS4_861N7ooUq4fUGX7RLCw

the `.jpg` is a lie, this file is a webp - and Electron's `nativeImage` doesn't support this type.
(when trying to save the picture in a browser, it will auto suggest name `hqdefault.jpg.webp`..)

this results in an empty native image - and setting `APIC` frame to an empty buffer:
https://github.com/th-ch/youtube-music/blob/3485d26b111c708f0c0aac6a7dec34b49435fc5d/plugins/downloader/back.js#L66-L68
results in a thrown error: `Unknown picture MIME type`

## The Fix

fixed by removing everything that comes after the ".jpg" (only if needed - the nativeImage is empty)
in our example the result is: 
https://i.ytimg.com/vi/tmntIJSf0Dw/hqdefault.jpg
which is a valid jpg file

> Also check before creating coverBuffer that nativeImage isn't empty to avoid potential error throwing

### DIFF
--- a/plugins/downloader/back.js
+++ b/plugins/downloader/back.js
@@ -55,7 +55,9 @@ function handle(win) {
 			{ ...nowPlayingMetadata, ...currentMetadata };
 
 		try {
-			const coverBuffer = songMetadata.image ? songMetadata.image.toPNG() : null;
+			const coverBuffer = songMetadata.image && !songMetadata.image.isEmpty() ?
+				songMetadata.image.toPNG() : null;
+
 			const writer = new ID3Writer(songBuffer);
 
 			// Create the metadata tags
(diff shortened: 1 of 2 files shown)


th-ch (approved):
Looks good, thanks for the fix!

Issue #338, opened 2021-06-26
Downloader Stuck On Downloading


Rarab5534:
Hi its Rarab again

So i went to download some songs and then my Youtube Music started doing this strange thing that i have never saw before
![image](https://user-images.githubusercontent.com/86423147/123517505-46bc0680-d699-11eb-9623-8f75975538bf.png)

So is this a bug? or just known?

Bellisario:
This not a very bug... you watch the download only by text and not by icon, so, if the problem is that you see a download icon with a strange pixel, this is not a progress.
If this is not the problem, I can say you that you need only to wait some time to see downloading title, conversion title, etc., this is only because this has been configured not to see a very accurate progress but only action that the download is doing.

P.S.: if you don't want to see download button you can just disable the plugin.

th-ch:
Hey @Rarab5534 and @Bellisario , this issue is confirmed because some APIs have changed and upstream dependencies are being updated at the moment - https://github.com/fent/node-ytdl-core/pull/938 has now been merged, just need to check if everything works ok with the new version, keeping this issue opened in the meantime!

Rarab5534:
Well how do i download it? and install it to youtube music then?

Bellisario:
@Rarab5534, you can't... so, or you need to build yourself the executable as a dev, or you need to await the next release.

Rarab5534:
ok

th-ch:
Hey @Rarab5534 & @Bellisario, #339 has been merged and should fix the plugin, it will land in the next version! (which will probably be released soon)

PR #339, opened 2021-06-27
Fix downloader plugin

th-ch:
This PR fixes the downloader plugin and should close https://github.com/th-ch/youtube-music/issues/338.


Rarab5534:
Will this be in the new update or a fix?

PR #340, opened 2021-06-27
[DEPRECATED] Local Subscriptions Plugin

fdr-smnv:
As the title says this plugin allows the user to save subscriptions locally without google account. It adds 'Subscriptions' tab in the menu which opens new window with the list of user's local subscriptions that can be opened in the main window. I think this plugin would go hand in hand with no-google-login plugin. Also I'm fairly new to electron so I'm pushing this plugin early to hopefully get some feedback and advice.

What works:
- Subscribing to music channels with subscribe button
- Saving subscription in plugin options
- New menu tab button
- Tab button opens window with subscriptions list
- Subscription list item on click redirects main window to channel url

Issues/TODO: 
- Couldn't manage to call functions from prerender without contextBridge
- Initial load doesn't render sub button
- loadURL makes full page rerender. If possible it would be great if user could click channel in the list and get redirected without slow full page refresh
- No styles for subs list & buttons
- (Probably issue) Saving subscription data in plugin options
- Ability to sort and filter saved channels
- Export subscription data?


config/defaults.js
```diff
@@ -76,6 +76,10 @@ const defaultConfig = {
 				"music_offtopic",
 			],
 		},
+    "local-subscriptions": {
+      enabled: false,
+      subscriptions: {}
```
th-ch, line 81: That will require a store migration: when you enable it for the first time, your config will end up with:
```
local-subscriptions": {
  enabled: true
}
```
And the code does not support empty `subscriptions` :)

plugins/local-subscriptions/back.js
```diff
@@ -0,0 +1,57 @@
...
+          webPreferences: {
+            preload: path.join(__dirname, "popup.js"),
+            affinity: "main-window"
+          }})
+        subscriptionsWindow.loadFile(templatePath(__dirname, "subscriptions-menu.html"))
+        subscriptionsWindow.webContents.on('did-stop-loading', () => subscriptionsWindow.webContents.send('saved-subscriptions', getOptions('local-subscriptions').subscriptions))
+        break;
+      case ACTIONS.SUBS_LI_CLK:
+        subscriptionsWindow.close();
+        console.log(action)
```
th-ch, line 32: nit: only log that in dev mode (`if is.dev()`)

PR #380, opened 2021-08-15
Fix discord clearActivity, menu

cpiber:
The callback sends multiple events, in particular two pause when going to the next song, so the timeout wasn't properly cleared. I have documented this.

Added menu buttons for the two options. Currently for the timeout time I simply open the config, to allow the user to properly input a number here a library is needed.

### DIFF
--- a/plugins/discord/back.js
+++ b/plugins/discord/back.js
@@ -16,6 +16,10 @@ module.exports = (win, {activityTimoutEnabled, activityTimoutTime}) => {
 	win.once("ready-to-show", () => {
 		rpc.once("ready", () => {
 			// Register the callback
+			//
+			// We get multiple events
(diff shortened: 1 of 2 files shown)


config/store.js
```diff
@@ -3,6 +3,11 @@ const Store = require("electron-store");
 const defaults = require("./defaults");
 
 const migrations = {
+	">=1.13.0": (store) => {
+		if (store.get("plugins.discord.listenAlong") === undefined) {
+			store.set("plugins.discord.listenAlong", true);
+		}
+	},
```
cpiber, line 10: I added a new option to disable the listen along button (#426), I hope I did this migration correctly

Araxeus:
Could you open a small pr to clean this up? if you're busy I can do it np

Issue #353, opened 2021-07-05
[Feature Request] Discord - listen along.


thijsi123:
Hello, I just wanna ask if you guys may want to make the listen along a feature to the discord plugin like in PreMID.
![Naamloos](https://user-images.githubusercontent.com/64159148/124438278-4a761a00-dd78-11eb-82d2-c3969b5fe791.png)

konhi:
I think that's code for this is there: https://github.com/PreMiD/Presences/tree/master/websites/Y/YouTube%20Music

Listen Along is just a link to the song.
```js
if (buttons) {
      presenceData.buttons = [
        {
          label: "Listen Along",
          url: `https://music.youtube.com/watch?v=${watchID}`
        }
      ];
    }
```

cpiber:
Could you close this? The PR is already merged

PR #383, opened 2021-08-19
Add "Listen Along" button, solve #353

konhi:
image
Solves #353.


th-ch (approved):
Change looks good, thanks for the contribution! ✅

Issue #327, opened 2021-06-16
2 Factor Authentication on Google Account reports wrong device type on "New sign-in alerts" and Google Account security activity.


ESodikromo:
When logging in to the app, using 2FA will send a message to another device to confirm the login request. However, instead of displaying the correct info for the machine used to login (Windows, Laptop Name), all that was reported was a nondescript "Mac" machine. The reported location was correct, however. This same info is accessible from the Google Account/Security If this is intended, explain so on the download page or in the app to prevent confusion and worry over unspecified login attempts.

Bellisario:
Yes, in fact to prevent Google from blocking the login, there is Mac user agent, but the problem is that isn't explained correctly...
Here are two solutions:
- Advert the user the login is "made" by a Mac (a false Mac);
- Detect platform and send to Google Login an user agent that respects the platform detected (ex. Windows as Windows, etc. and not Windows as Mac)

Araxeus:
This issue should be fixed and can be closed

PR #384, opened 2021-08-19
Fix incorrect Google alert caused by changing user agent coresponding to current platform

konhi:
Solves #327

### DIFF
--- a/index.js
+++ b/index.js
@@ -211,19 +211,14 @@ app.once("browser-window-created", (event, win) => {
 			// User agents are from https://developers.whatismybrowser.com/useragents/explore/
 			const userAgents = {
 				mac: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:70.0) Gecko/20100101 Firefox/70.0",
-				windows: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.124 Safari/537.36",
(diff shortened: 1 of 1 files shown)


Araxeus (changes requested):
Could be slightly simplified using [electron-is](https://github.com/delvedor/electron-is) (it is already used throughout this program)
```js
const userAgent = 
	is.darwin() ? userAgents.mac :
	is.windows() ? userAgents.windows :
	userAgents.linux;
```
(also because you are currently trying to change the value of a const after initialization)
but anyway did you check that google OAuth works? there is a comment saying firefox must be forced for it to work

konhi:
> Could be slightly simplified using electron-is (it is already used throughout this program and imported in index.js line 

Thanks! This code looks much prettier! 💕

> (also because you were trying to change the value of a const after initialization)

Is it a bad practice? If so, could you please give some reference? I'm not really a javascript expert! 😅

> but anyways did you check that google OAuth works? there is a comment saying firefox must be forced for it to work

Good question! If you mean logging in, yes, it worked for me and Google didn't even alert me about anything suspicious! Anyway, I think it may be better option to still use Firefox user-agent.

Araxeus:
> Is it a bad practice? If so, could you please give some reference? I'm not really a javascript expert! 😅

Well const = constant which means immutable value, shouldn't be changed after giving it its initial value.
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Variables#constants_in_javascript

If the value is supposed to change later, you would probably use `let` instead of `const`
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Variables#declaring_a_variable

th-ch (approved):

Issue #343, opened 2021-06-28
Discord rich presence displays wrong song title


Keyboardsheep:
When I'm using this (Youtube Music) in my Discord status, it displays the song title of the next song in queue. It displays the correct artist though.
![image](https://user-images.githubusercontent.com/23726813/123669443-b3521500-d801-11eb-9a74-6a4e82a265ae.png)
![image](https://user-images.githubusercontent.com/23726813/123669437-b0efbb00-d801-11eb-8fe5-1a64ab6658fd.png)

Araxeus:
Try clearing app cache

Keyboardsheep:
I've tried that, unfortunately, it didn't work though I might have done something wrong. (I clicked `Reset app cache when app starts` then restarted the app. I also tried deleting the cache folder in `AppData\Roaming\YouTube Music\`)
Also, I've noticed that it will sometimes show the correct title if I skip a song, then go back to that song, also one time during the recording, Discord displayed a completely unrelated song later in the queue.

Grogyas:
I have the same problem even after a clean install of Windows it's the same

ofkalkan:
I have same problem as well.

After some tries I found out that it happens when you log-in your Google account. When you log-out, it displays the right title and time left. Hope this helps to the devs.

Araxeus:
> I have same problem as well.
> 
> After some tries I found out that it happens when you log-in your Google account. When you log-out, it displays the right title and time left. Hope this helps to the devs.

I have no idea what cause this, it worked perfectly fine for me when logged in/out

i can only recommend clearing app cache / uninstall + reinstall app (that fixed the problem for someone before)

Grogyas:
> I have same problem as well.
> 
> After some tries I found out that it happens when you log-in your Google account. When you log-out, it displays the right title and time left. Hope this helps to the devs.

Same, but it's happening only in version 1.12.0-2, 1.9-1.11 works fine

Grogyas:
Ok, got it fixed

Go to /providers/song-info.js

Paste this there
`const titleSelector = ".title.style-scope.ytmusic-player-bar";

const getTitle = (win) => {
	return win.webContents
		.executeJavaScript(
			"document.querySelector('" + titleSelector + "').innerText"
		)
		.catch((error) => {
			console.log(error);
		});
};`

And in const handleData
Replace 
`songInfo.title = data?.videoDetails?.title;`
With 
`songInfo.title = await getTitle(win);`

And build it with yarn

[...]

Issue #334, opened 2021-06-22
Reports wrong title and pic (and sometimes artists too)


FiestaLake:
I face similar issue with notification (and discord) like https://github.com/th-ch/youtube-music/issues/266. It shows something like this:
![image](https://user-images.githubusercontent.com/68979855/122870636-e2cbd200-d368-11eb-8c08-928104b2d654.png)

Using latest aur version.

Araxeus:
https://github.com/th-ch/youtube-music/issues/266#issuecomment-836076371
Try to do what he did

FiestaLake:
> [#266 (comment)](https://github.com/th-ch/youtube-music/issues/266#issuecomment-836076371)
> Try to do what he did
> 
> (delete all Youtube Music data from .config)

Already did many times. I also tried the PR but it doesn't allow me to play anything.

FiestaLake:
Should I try to get some log?

Rarab5534:
uninstall yt music and install yt music again

FiestaLake:
> uninstall yt music and install yt music again

= already done. And more important, it doesn't do anything.

SHU-red:
Same Problem here

marmichalski:
This happens in fe0f213919084f9e0bd95ce626ec711fb522d436 for me too. Displays current artist with next track track's title. For example, reports that I'm playing SIlent Theory - Animals, but it actually plays Silent Theory - Before the Storm and the next track that will play is Architects - Animals.

lephuongbg:
Still happening on 1.13.0

mesmerx:
as the same approach of the  https://github.com/th-ch/youtube-music/pull/268/files i did a pr that do almost the same thing https://github.com/th-ch/youtube-music/pull/439

PR #393, opened 2021-09-01
Fix wrong track title being reported

marmichalski:
#334, #343

Actually, everything in songInfo is being replaced by queued track, since there are 2 requests sent, the first one being current song and the second one is for the queued track.

I have created the same workaround that's done for artist, but this will need a closer look.

### DIFF
--- a/providers/song-info.js
+++ b/providers/song-info.js
@@ -32,6 +32,13 @@ const getPausedStatus = async (win) => {
 	return !title.includes("-");
 };
 
+const getTitle = async (win) => {
+	return win.webContents.executeJavaScript(`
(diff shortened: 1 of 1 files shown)


marmichalski:
Just found https://github.com/th-ch/youtube-music/pull/268. It might be a good idea to reopen that one instead of merging this. @Araxeus?

marmichalski:
This seems to no longer be needed as of 1.13.0. 🤔

SHU-red:
Still present for me and @lephuongbg

https://github.com/th-ch/youtube-music/issues/334#issuecomment-935132168

PR #431, opened 2021-10-13
fix: mpris instance not registering itself and media controls

itzmanish:
Sorry, I missed that MPRIS was not getting registered for youtube-music, and I was not wrong with the statement 
>  MPRIS is working in linux

because when youtube music is running there is also an chromium instance which register itself for MPRIS, and the playerctl command I used to test that was play, pause, next and prev but I did not listed my all player instance that time to see if it is really youtube-music mpris is registered or not. All the command worked that time because chromium instance was listening for those command.

So I fixed that and also tested again "**PROPERLY**" with playerctl -p youtube-music commands, and also with mediaplayer perl script which I am using in my i3blocks. 

Again **I am sorry I did not tested mpris support PR properly.**

### DIFF
--- a/plugins/shortcuts/back.js
+++ b/plugins/shortcuts/back.js
@@ -36,6 +36,9 @@ function registerShortcuts(win, options) {
 				'xesam:title': songInfo.title,
 				'xesam:artist': songInfo.artist
 			};
+			if (!songInfo.isPaused) {
+				player.playbackStatus = "Playing"
+			}
 		}
 	}
 	)


Araxeus:
Couldn't you just use `playPause()` instead of creating 2 new functions? is there a specific reason you did it this way?

also heads up - I'm going probably going to change auto-confirm-when-paused and there will be no more yns_pause,
which means song-control-front can be deleted and also playPause will go back to `playPause: () => pressKey(win, "space")`)

PR #439, opened 2021-10-17
Fix multiple songInfo calls 

mesmerx:
in the daily usage of tuna plugin that i created i get that sometimes the songs info are mixed with the currently one and the next one
debuging i get that the handleData are receiving or some reason the info for the currently and the next one, so i fix that setting functions to get currently info for what i can get from document itself

i set some verification for edge cases without info and a song-info crash that happens too

### DIFF
--- a/providers/song-info.js
+++ b/providers/song-info.js
@@ -62,6 +62,10 @@ const getSongDuration = async (win) => {
 	`);
 }
 
+const getSongURL = async (win) => {
+	return  win.webContents.getURL().split("&")[0];
+}
+
 
 // Fill songInfo with empty values
 const songInfo = {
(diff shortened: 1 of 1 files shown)


Araxeus:
Would probably be better at this point to get all the data from the same place instead of `executejs` on every field (get all song info from `song-info-front.js` and send it to `song-info.js`)

also am I missing something or imageSrc+songUrl weren't changed? they should come from the same place as the title to avoid out of sync stuff

(I actually was planning to do all this https://github.com/th-ch/youtube-music/issues/334#issuecomment-937112538 but you beat me to it)

mesmerx:
`onXHR request -> check if XHR.details.title !== document.querySelector(".title.ytmusic-player-bar")?.textContent `
thats a good idea
about the url, i didn't step much int his field because for the tuna i only need title and the artist, i just set the other fields that i know the tag where it come from
one thing that i'm currently thinging is about getting a event for update in progress bar, because as it's right now i dont get update for progress at all

mesmerx:
@Araxeus i did the song url too in the 
https://github.com/th-ch/youtube-music/pull/439/commits/4b0b03b87b576a6cd4346d48e9c5a2d1236f1a4c

mesmerx:
i'll code this tomorrow, so expect  one or twot days commit for that

mesmerx:
pr closed, i'll wait #443

Araxeus:
@mesmerx you wanna add the live updates yourself? or shall I just add this to my ever expanding list of pr's? 😝

PR #443, opened 2021-10-23
get songInfo from youtube API

### LINKED ISSUE #334: Reports wrong title and pic (and sometimes artists too)

### LINKED ISSUE #343: Discord rich presence displays wrong song title

### LINKED ISSUE #427: Last.fm sending wrong song information and only updates "Now Playing" not sending scrobble requests
I noticed Last.fm gets the wrong song information. If there is a queue/playlist then the app sends the next song's name instead of the current song, i.e. name of next song in queue + current song's album name. 

image

This behaviour doesn't happen if I manually go to a previously listened song in the queue and start listening from there, but happens again when it gets to a new song in the queue.

Also, Last.fm doesn't seem to scrobble anything, only updates Now Playing.

Araxeus:
Fix #334 , Fix #343 , Fix #427

Gets songInfo from `api.getPlayerResponse()` when the `video` element sends a `loadeddata` event

This eliminates the bug when the player received an XHR of the next song and it got cached in songInfo

----

small fixes:
* `url?.split("&")[0]` aims to remove playlist info etc.. from songInfo.url
* `suffixesToRemove` (from title & artist) now ignore case


Araxeus:
> this pr is a cleaner way of getting info
> just to known, this pr fixes the need for a if src in getImage? because sometimes src is empty as currently implementattion, and in terminal a error are triggered
> https://github.com/th-ch/youtube-music/pull/439/files#diff-183c64bf410feea83315a0ab22d83706800c59e7e028b8babf9b30d5cad3d21fR26
> the line that fixes that error in the old way

Indeed this way we always have a valid src
Also since we are using here the official youtube api - we always get the right artist name, no more need for scraping artistName from playbar.
(That fix was part of #228 ):
>SongInfoProvider was parsing artist name based on video request , not on song - resulting in buggy artist names.
changed it scrape artist name from playbar (native artist name)

PR #448, opened 2021-10-24
Video Toggle Plugin

### LINKED ISSUE #149: (Feature_request) Show song cover by genrating preview from video
When the video is hidden , player should show album cover either by using the default creator provided or from any frame from video of the song

### LINKED ISSUE #392: [Feature Request] Video and Album Art switcher
when i used hide-video-player plugin, the app shows only the playlists and nothing more. I think it will be more convenient if there is instant switcher like on mobile whenever we want to show the video player or not.
if the video has no album art, switch it to video thumbnail.

i think it will be great if the switcher placed on my mouse or above the video player
image

mobile example
image

Araxeus:
Resolves #392 , Resolves #149

Adds a button that allow switching between video and "song" mode (show thumbnail instead of video)

forceHide in options makes this plugin behave like the old hide-video-player (hide the whole main panel)

Screenshots of the button in action:
Video
Song

> Also refined some plugin's description


plugins/video-toggle/back.js
```diff
@@ -0,0 +1,11 @@
+const { injectCSS } = require("../utils");
+const path = require("path");
+const { urlToHttpOptions } = require("url");
+
+module.exports = (win, options) => {
+    if (options.forceHide) {
+		injectCSS(win.webContents, path.join(__dirname, "forceHide.css"));
```
th-ch: nit: lower case file names are often better because some filesystems are case sensitive (eg Linux) and others are not (eg MacOS).
```suggestion
		injectCSS(win.webContents, path.join(__dirname, "force_hide.css"));
```

plugins/video-toggle/back.js
```diff
@@ -0,0 +1,11 @@
...
+	} else {
+		injectCSS(win.webContents, path.join(__dirname, "buttonSwitcher.css"));
```
th-ch:
```suggestion
		injectCSS(win.webContents, path.join(__dirname, "button_switcher.css"));
```

config/defaults.js
```diff
@@ -77,6 +77,10 @@ const defaultConfig = {
 				"music_offtopic",
 			],
 		},
+		"video-toggle": {
+			enabled: false,
```
th-ch, line 81: nit: we could  add a migration so that people who had the old plugin enabled have this one enabled in the new version.

th-ch (commented):
Looks good overall, but when the video toggle is enabled ("song" mode), I get `TypeError: Only absolute URLs are supported` in https://github.com/th-ch/youtube-music/blob/713e005aa89dc314a223d1f8220122616fbf9b08/providers/song-info.js#L14 (`src` is undefined - possibly linked to the video being overridden?)

Araxeus:
> Looks good overall, but when the video toggle is enabled ("song" mode), I get `TypeError: Only absolute URLs are supported` in https://github.com/th-ch/youtube-music/blob/713e005aa89dc314a223d1f8220122616fbf9b08/providers/song-info.js#L14 (`src` is undefined - possibly linked to the video being overridden?)

How are you getting an error in song-info? This plugin doesnt even use it nor change any of its data 😵

Where is it originating from? Did that happen in every song or just a few?

Araxeus:
Ok I think I figured this one out too. you were playing a video with low resolution of artworks and so when it calls `thumbnails.pop()` it actually removes the last thumbnail...

plugins/video-toggle/front.js
```diff
@@ -0,0 +1,82 @@
+const { ElementFromFile, templatePath } = require("../utils");
+
+const { setOptions } = require("../../config/plugins");
+
+function $(selector) { return document.querySelector(selector); }
+
+let options;
+
+const switchButtonDiv = ElementFromFile(
+    templatePath(__dirname, "buttonTemplate.html")
```
th-ch: nit: same comment for lower-case file name
```suggestion
    templatePath(__dirname, "button_template.html")
```

th-ch (approved):
Looks good! ✅  - I cannot reproduce the `src` bug anymore so it looks fixed 🎉

plugins/video-toggle/front.js
Araxeus: done :)

th-ch (approved):

Araxeus:
@th-ch Oh I forgot to add `, { once: true, passive: true }` to the apiloaded listener

could you add the in please? :)

th-ch:
Sure! The listener has been fixed in https://github.com/th-ch/youtube-music/commit/cee2e066b95068b2c4ed172dc77ac17c04fbd107 and the plugin migration done in https://github.com/th-ch/youtube-music/commit/953d6fe3e40efe7fae480d3400ce338cae59e45f :)

PR #449, opened 2021-10-24
set resume on start url to songInfo.url

### LINKED ISSUE #444: "Resume last song when app starts" only working when queue pane is open
I've found on different machines and releases than the "Resume last song when app starts" option only works when the queue pane is open:

1. If the app is restarted whilst the queue pane is open (bottom right arrow pointing down and queue showing) the option works as intended.
2. If the app is restarted whilst the queue pane is closed (bottom right arrow pointing up and showing your library) nothing plays after the restart.

Araxeus:
Fix #444 by setting `config.url` to the current playing song url from songInfo
(including playlist info unlike `songInfo.url`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -197,13 +197,6 @@ app.once("browser-window-created", (event, win) => {
 		event.preventDefault();
 	});
 
-	win.webContents.on("did-navigate-in-page", () => {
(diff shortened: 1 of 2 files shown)


th-ch (approved):

PR #451, opened 2021-10-24
Blur plugin tweaks and integration with in-app-menu

Araxeus:
* Adds blur to header (in the screenshot below its the playlists|albums|songs|artists|subscriptions) and recently added div) 
image
 this was done for consistency because this header sticks to the navbar when scrolling the page down.

(only problem is that you can kinda see the separating line between both blurred elements which can be a bit weird but its better than the header being black when the top bar is blurred)

* Makes in-app-menu compatible with this plugin with some small adjustments to the css
* Dynamically adjust navbar position to compensate for the scrollbar which appears on all pages except the playerpage (it was weird when there was a 15px gap in the navbar when inside a colored playerpage)

> this follows up pr #442 by @cdaydreamer


th-ch (approved):

PR #456, opened 2021-10-26
alert on initial hide-menu enabled

### LINKED ISSUE #342: Alert user how to show menu after "Hide menu" enable
Now, if you enable "Hide menu", you have some problems to show it, because you are not informed how to do. Can be very useful show to user a popup that alerts how to un-hide menu at any time, after option enabled.

Araxeus:
Resolves #342 and hopefully all the questions about how to show the menu 😅

Many people don't know how to show menu after they hid it - so this should help

ps: I'm not sure about the added notes in the readme - you can change them or their positions (or delete them) if you want

### DIFF
--- a/index.js
+++ b/index.js
@@ -348,10 +348,10 @@ app.on("ready", () => {
 		});
 	}
 
-	if(config.get("options.hideMenu") && !config.get("options.hideMenuWarned")) {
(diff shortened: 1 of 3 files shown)


th-ch (approved):
Looks good, thanks for improving this! ✅

PR #458, opened 2021-10-26
use apiLoaded event in audio-compressor plugin

Araxeus:
simple optimization
(video element is always loaded when [`apiLoaded`](https://github.com/th-ch/youtube-music/blob/a47c906ab2ac47bd0a5589099b4e32bf1d3c4ae6/preload.js#L59-L78) is fired)


th-ch (approved):
✅

PR #462, opened 2021-10-30
fix playback speed plugin

Araxeus:
* Fixes memory leaks from observer within observer and unnecessary observers watching the whole document
  see https://github.com/th-ch/youtube-music/pull/288#pullrequestreview-787493434
  
* This also normalize the steps of the slider to be 0.125 (1/8)
  see here all available levels obtained by dragging the bar:
  ```js
  var x = []
  for(let i=0; i<=2; i+=0.125) x.push(i || 0.07)
  console.log(x)
  // (17) [0.07, 0.125, 0.25, 0.375, 0.5, 0.625, 0.75, 0.875, 1, 1.125, 1.25, 1.375, 1.5, 1.625, 1.75, 1.875, 2]
  ```
  

* This also adds a mousewheel listener to the playback speed bar, which allows more precise tuning of the current speed - 0.07 up to 16, in increments of 0.01 (good luck getting to 16 lol)


plugins/playback-speed/front.js
```diff
@@ -1,83 +1,87 @@
...
+	})
+	slider.addEventListener('wheel', e => {
+		e.preventDefault();
+		if (isNaN(playbackSpeed)) {
+			playbackSpeed = 1;
 		}
-	);
+		// e.deltaY < 0 means wheel-up
+		playbackSpeed = roundToTwo(e.deltaY < 0 ? 
+			math.min(playbackSpeed + 0.01, MAX_PLAYBACK_SPEED) :
```
th-ch:
```suggestion
			Math.min(playbackSpeed + 0.01, MAX_PLAYBACK_SPEED) :
```

th-ch (commented):
The mousewheel works fine but when clicking directly on the slider, the value updates in the UI but the playback rate is not changed anymore, is it expected?

Araxeus:
> The mousewheel works fine but when clicking directly on the slider, the value updates in the UI but the playback rate is not changed anymore, is it expected?

Thats weird... I tested this extensively and it worked great.

I'll look into it when i get home

Araxeus:
I figured it out - 'immediate-value-change' is a [polymer event](https://www.webcomponents.org/element/polymerElements/paper-slider/elements/paper-slider) which only fire on drag, not click. so I had to add a 'd' on there and use 'immediate-value-change**d**'

Except that event doesn't bubble up apparently so I have to set it up straight on the slider, which required waiting for its first load to find it in the dom. anyways it should be fixed now 41285ac

th-ch (approved):
Confirmed it now works fine on my side with both click and mousewheel 🎉 

PR #463, opened 2021-10-30
Sponsorblock fix + use new apiLoaded event

Araxeus:
1. Update Electron to v12.2.2 - Fixes sponsor block not working
  image
    * see https://github.com/node-fetch/node-fetch/issues/568#issuecomment-932200523
      (I tested this method and it works, but updating electron is better)
    * also see https://github.com/oliverschwendener/ueli/issues/807#issuecomment-932851049

> note for near future: electron v14 **remove** the remote module
will need some code changes when v13 is deprecated at `25.03.2022`
(electron v12 will be deprecated very soon - `16.11.2021`)

2. delete `providers/video-element` and use instead the new apiLoaded event in both `disable-autoplay` and `sponsorblock` plugins)

3. Change all existing apiLoaded event listeners to close on trigger (`{ once: true, passive: true }`)
   (api is loaded only once, page refresh re-register those listeners anyway)

4. Fix spamming video time override d35fef82feb81f305401b7d215504d8110775616
   image


th-ch (approved):
Looks good, thanks for the fix/improvements! ✅

PR #465, opened 2021-11-01
Fix sponsorblock skipping when not needed

Araxeus:
* Fix sponsorblock skipping when not needed by resetting the segments before next song load (using the "emptied" event)

  The bug originally occurred because "timeupdate" was called while the segments were still from the old song because there is a slight delay getting them from ipc->songInfo->**fetch**->json->sort->ipc etc..

* use `loadedmetadata` instead of `loadeddata` in song-info-front since the data is already available (its only a few milliseconds faster but whatever helps)

* small lint for in-app-menu


th-ch (approved):

Issue #483, opened 2021-11-11
ShortCut Key & System Tray Control Broken


criticalsoft:
Previous Track Is Pause Now
Next Track Is Previous 10 Second

Araxeus:
what are you talking about?

if you press `j` does it switch to the next track?
 `k` previous track?

criticalsoft:
What are you talking about YouTube Next Song shortcut key is Shift + n Previous shortcut key is Shift + p.
Ok, I Submited A Fix About This.
https://github.com/th-ch/youtube-music/pull/485

Araxeus:
obviously you have different keybinds for some reason...

can you say what OS are you on?

criticalsoft:
> 
> 
> obviously you have different keybinds for some reason...
> 
> can you say what OS are you on?

Windows 10

Araxeus:
Well that's weird since i'm only win10 too. no idea why you have different keybinds but ill make a pr that control the songs without using any keybinds

Araxeus:
any other currently existing shortcuts that don't work as intended on your machine?

criticalsoft:
> 
> 
> any other currently existing shortcuts that don't work as intended on your machine?

I Just Need Play Pause Next Previous So Far. Sorry I didn't test other keys.

criticalsoft:
> 
> 
> Well that's weird since i'm only win10 too. no idea why you have different keybinds
> 
> but anyways ill make a pr that control the songs without using any keybinds

What's the year ?

Araxeus:
> I Just Need Play Pause Next Previous So Far. Sorry I didn't test other keys.

Could you please test all shortcuts whenever you have time so that I know which actions to change?

> What's the year ? Are We in the same time zone ?

2021? what kind of question is that? O_O
timezone is gmt +2 but I don't see why that would be relevant either

criticalsoft:
> 
> 
> > I Just Need Play Pause Next Previous So Far. Sorry I didn't test other keys.
> 
> Could you please test all shortcuts whenever you have time so that I know which actions to change?
> 
> > What's the year ? Are We in the same time zone ?
> 
> 2021? what kind of question is that? O_O timezone is gmt +2 but I don't see why that would be relevant either

I mean are we in the year.

criticalsoft:
https://support.google.com/youtube/answer/7631406?hl=en

criticalsoft:
Normal YouTube ShortCut Key Is Shift + P / N
[...]

PR #486, opened 2021-11-12
FIX: Previous & Next Track Control

criticalsoft:


criticalsoft:
The shortcut key is the same https://github.com/th-ch/youtube-music/issues/483#issuecomment-967173706
So Reopen pull request.

criticalsoft:
> 
> 
> We already concluded this isn't needed, but I guess this shortcut should **also** work

Yes, Please Approve This Request.

PR #488, opened 2021-11-12
New Plugin: Exponential Volume

Rubecks:
Port of [Marco Pfeiffer's "Youtube Music fix volume ratio 0.4"](https://greasyfork.org/en/scripts/397686-youtube-music-fix-volume-ratio/) on Greasyfork. As stated in the script's description:

The volume slider of YouTube Music is linear by default which causes it to feel very ineffective until the very end where it mutes the music. The volume slider of YouTube (Video) is exponential to address that issue but YouTube music didn't get so much love.

This script addresses this issue by applying an exponent of 3 to the volume setting, which results in these settings:

    100% = 100%
    50% = ~12.5%
    20% = ~0.8%
    10% = ~0.1%
    5% = ~0.0125% (5% is the lowest selectable value by the UX)

### DIFF
--- a/plugins/exponential-volume/front.js
+++ b/plugins/exponential-volume/front.js
@@ -1,0 +1,47 @@
+// "Youtube Music fix volume ratio 0.4" by Marco Pfeiffer
+// https://greasyfork.org/en/scripts/397686-youtube-music-fix-volume-ratio/
+
+const exponentialVolume = () => {
+	// manipulation exponent, higher value = lower volume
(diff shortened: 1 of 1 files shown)


Araxeus:
Works good on my side 👍🏼 

Did you try it out precise-volume plugin? :)

Rubecks:
Yup, and it helped a little with the volume problem since I could select something besides the multiples of 5. But the range of comfortable volume levels on my audio setup is too short without this script. Using both of them together, in most cases I got no sound under 5%, but it still is way better than the default linear slider.

th-ch (approved):
Looks ok, thanks for the new plugin! ✅

PR #491, opened 2021-11-14
fix interaction between blur navbar & in-app-menu

### LINKED ISSUE #490: Fading blur navbar
Hi, i just love the app.

I love the blurry navbar but I am the only one that´s noticing that the blur fades out when going upwards?

image

Araxeus:
fix #490 by making the nav bar background sticky and moving it upwards which fixes the blur issues

(18px blur is now way too much, 8px is close but a bit lower than it was before - but I think it looks ok?)

this pr also fix the height of the search page item header when stuck (for in-app-menu) and include it in the blur plugin for consistency since it sticks to the navbar when scrolling down

<details>
  <summary>Click to see screenshots comparisons</summary>
  
 * Blur Before:
Blur_Before

* Blur After:
Blur_After

* Search page clip+blur before:
Search-Page_Before

* Search page after:
Search-Page_After

</details>

### DIFF
--- a/plugins/blur-nav-bar/style.css
+++ b/plugins/blur-nav-bar/style.css
@@ -1,6 +1,6 @@
 #nav-bar-background, #header.ytmusic-item-section-renderer {
     background: rgba(0, 0, 0, 0.3) !important;
-    backdrop-filter: blur(18px) !important;
+    backdrop-filter: blur(8px) !important;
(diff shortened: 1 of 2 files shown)


th-ch (approved):
Looks good, thanks for the style improvements! ✅

Issue #507, opened 2021-12-02
[Manjaro] Unsupported Browser


WhiteHusky:
Using the AUR package `youtube-music-bin` and cannot login due to the unsupported browser error.

So whatever fix mentioned in #17 didn't stick.

Araxeus:
maybe the user agent needs updated version

https://github.com/th-ch/youtube-music/blob/a14d27da70caba3fcdfc104f04aeef15646fa1b5/index.js#L207-L211

WhiteHusky:
@Araxeus 
Looks like it's gonna be a pain to fix. I did a patch in an attempt to circumvent but,
![image](https://user-images.githubusercontent.com/3196371/144471672-cdec76d1-769d-45b8-b9ed-37b50be87f1d.png)
```js
const userAgents = {
	mac: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:70.0) Gecko/20100101 Firefox/70.0",
	windows: "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:90.0) Gecko/20100101 Firefox/90.0",
	linux: "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0",
}

const userAgent = 
	is.macOS() ? userAgents.mac :
	is.windows() ? userAgents.windows :
	userAgents.linux;

// Also overwrite the user agent to prevent finding the actual user agent
win.webContents.userAgent = userAgent;
```

WhiteHusky:
Wait what
![image](https://user-images.githubusercontent.com/3196371/144472348-cb69d5c4-fbf8-4f18-9377-c1af688a932f.png)
Okay so _maybe_ my patch works. Sorta. I'll publish it in a PR and someone much more of a guru can figure out what's going on. Electron is not in my wheelhouse.

mimaketi:
On Arch, opening the developer tools, and running the below in the console allowed me to login no problem.

`Object.defineProperty(navigator, 'userAgent', {
    get: function () { return 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0'; }
});`

obvionaoe:
Please fix this and #17 , as the app is completely unusable for someone who only uses the library part of Youtube Music

Araxeus:
[...]

PR #508, opened 2021-12-02
Allow Google Signin

WhiteHusky:
This doesn't truely fix the problem. A weird combination of refreshing several times and...generally spastastic navigation behavior seems to get around the signin. This is mostly here to show what I tried.

### DIFF
--- a/index.js
+++ b/index.js
@@ -166,6 +166,8 @@ app.once("browser-window-created", (event, win) => {
 	setupSongInfo(win);
 	loadPlugins(win);
 
+	let actualUserAgent = win.webContents.userAgent;
+
 	win.webContents.on("did-fail-load", (
 		_event,
 		errorCode,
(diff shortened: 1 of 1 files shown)


WhiteHusky:
Related to #507

WhiteHusky:
Gets weirder:

So after doing the weird navigation dance it kinda..._doesn't_ sign in correctly. It appears if you are, but you get a signed out front page. Removing the patch, signing in, then signing back in by clicking your signed out account it works.

Obviously this is broken.

WhiteHusky:
More testing and trying to find a concise path to it working finds it's just a coinflip. Trying going back, forward, and several other navigations seems to be random when it'll just work or not. So I'm stumped.

I also tried just updating the user agent but no dice.

Skaifer:
Hello! 
So I'm facing the same issue on Fedora using appimage. Just in case this info would help, I could not sign in to google account on Vivaldi until I turned on google's _**crypto token** extension_ in privacy settings. 

Cheers!

Araxeus:
Hey could you try my PR? #515 

I think I might have fixed the problem 🙂

Araxeus:
@WhiteHusky #515 was merged and should have fixed the problem

So this PR could be closed

th-ch:
Closing as it should be fixed and the PR is stable - feel free to re-open a PR if needed!

PR #480, opened 2021-11-10
Mpris + obs-tuna fixes

Araxeus:
I've been testing on discord with @JoeJoeTV (many thanks for all the hours invested 😃 ) and we managed to fix all mpris issues (play/pause not working + update progress attribute + seek *from* mpris + seek *to* mpris)
should close #372

original comment with mpris bugs: https://github.com/th-ch/youtube-music/issues/65#issuecomment-964235549

> docs:
https://www.npmjs.com/package/mpris-service
https://github.com/dbusjs/mpris-service/blob/master/src/index.js#L26-L132
https://github.com/dbusjs/mpris-service/search?q=seeked
https://github.com/dbusjs/mpris-service/blob/master/examples/player.js

For obs-tuna this implements a live update to the progress

also fix #487 by changing status from `Playing`/`Paused` to `playing`/`stopped` 

this pr also adds a songInfo.album for both mpris and tuna (when available, never is available if current song has a video)

> ps. mpris could maybe be moved to its own plugin instead of 'camping' inside the shortcuts plugin

PR #509, opened 2021-12-03
Fix mpris bugs - follows #480

Araxeus:
I sadly couldn't test the changes on linux after the latest merge conflict solving on PR #480 since I use windows, and some stupid bugs made their way to the main branch because of that.

Thankfully @JoeJoeTV helped test mpris on linux once again, and all the little bugs have been fixed in this PR

Sorry about those!

-fix missing songControls (forgot to re-add it after moving the code from `shortcuts/back.js` to `shortcuts/mpris.js` 🤦🏼‍♂️)
-fix e.target instead of e.detail in apiLoaded event (to access the api)
-fix document.querySelector('video') being called before apiLoaded
-fix 'seeked' event listener
-setup timeChange Listener if on linux+shortcuts enabled

(I also added a sneaky change to the youtube-music.css that hides the chromecast button which doesn't work because electron doesn't support[[1](https://github.com/electron/electron/issues/869)][[2](https://github.com/electron/electron/issues/7024)] the `chrome.cast` api)

### DIFF
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -1,6 +1,7 @@
 const mpris = require("mpris-service");
 const { ipcMain } = require("electron");
 const registerCallback = require("../../providers/song-info");
+const getSongControls = require("../../providers/song-controls");
 
 function setupMPRIS() {
 	const player = mpris({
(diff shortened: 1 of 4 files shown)


th-ch (approved):
Looks good, thanks for the improvements! ✅

PR #510, opened 2021-12-04
Aligned lyric design

MiepHD:
I aligned the lyric to the normal lyrics so that the lyric isn't that tiny.

### DIFF
--- a/plugins/lyrics-genius/style.css
+++ b/plugins/lyrics-genius/style.css
@@ -7,6 +7,9 @@
 }
 
 #contents.genius-lyrics {
-    font-size: initial;
+    font-size: 14px;
     opacity: 0.9;
+}
+.style-scope ytmusic-tab-renderer {
+	display: block;
 }
--- a/plugins/lyrics-genius/style.css
+++ b/plugins/lyrics-genius/style.css
@@ -7,9 +7,6 @@
 }
 
 #contents.genius-lyrics {
-    font-size: 14px;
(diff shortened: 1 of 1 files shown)


Araxeus:
Can't you just do `font-size: medium;` ? is that actually smaller than 14px for you? (because thats what `initial` should be anyways

also those lines are not needed here, don't really see why you added them
```css
.style-scope ytmusic-tab-renderer {
	display: block;
}
```

MiepHD:
The original looks like that for me:
image
When I add display:block; it looks like:
image
And when adding font-size:14px; it looks how it should:
image
The LyricFind lyrics have font-size 14px

Araxeus:
Funny thing is that on my pc (1440p monitor), 14px is actually smaller than 'medium' and it looks really tiny, can't even read it at all

Can you try using one of those sizes?
```css
/* <absolute-size> values */
font-size: xx-small;
font-size: x-small;
font-size: small;
font-size: medium;
font-size: large;
font-size: x-large;
font-size: xx-large;
font-size: xxx-large;
```

Also this display-block is not needed like I said, for 2 reasons:
1. The lyrics actually looks better when centered (especially on smaller window sizes)
2. Your current selector `.style-scope ytmusic-tab-renderer` **selects other elements** which are not the lyrics tab (its not a unique selector)

Araxeus:
The viewport size works perfectly 👌🏼

th-ch (approved):

PR #513, opened 2021-12-10
fixes mpris bug in snap

markbaas:
Seems that to make mpris work properly in snap, we need to add the slot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,9 @@
 				"deb",
 				"rpm"
 			]
+		},
+		"snap": {
+			"slots": [{ "mpris": { "interface": "mpris" }}]
 		}
 	},
 	"scripts": {


Araxeus:
Could you show the documentation you used to find this?

Because I can't find any `slots` attribute in the snap key doc of electron-builder
https://www.electron.build/configuration/snap.html

markbaas:
@Araxeus I couldn't find any docs on the slots, but the snap documentation clearly says slots have to be defined. So I found this PR in electron-builder project: https://github.com/electron-userland/electron-builder/pull/5313

The link you are passing me basically already answeres your question as it is the docstring for the slot property.

Plugs is the other way around, this would be required when another app wants to connect to youtube-music through mpris interface. Only the slot lifts the apparmor error:
```
DBusError: Connection ":1.100" is not allowed to own the service "org.mpris.MediaPlayer2.youtube-music" due to AppArmor policy
```
How about submitting the project to snapcraft? Need any help with this?

th-ch (approved):
Thanks for the contribution - diff looks ok but I did not test it personally, fine with merging if it's tested and works fine ✅

PR #515, opened 2021-12-12
update electron & remote & user agents

### LINKED ISSUE #8: Sign-in doesn't work
Video playback works but sign-in currently doesn't work for me in version 1.1.5. It doesn't work for me in version 1.1.4 either, but it previously worked for me in version 1.1.4. GUI message says "Cannot load YouTube Music... internet disconnected?"

I also can't find anything in my Google account security settings to suggest it's blocked a sign-in from an insecure device or application.

Don't know if it's related but also ad-blocker plugin doesn't block ads.

### LINKED ISSUE #17: No login
Error to login account google

This is error message
"This browser or app may not be secure. Learn more

Try using a different browser. If you’re already using a supported browser, you can refresh your screen and try again to sign in."

### LINKED ISSUE #327: 2 Factor Authentication on Google Account reports wrong device type on "New sign-in alerts" and Google Account security activity.

### LINKED ISSUE #507: [Manjaro] Unsupported Browser

### LINKED ISSUE #526: Unsupported Browser Issue On Mac
Screen Shot 2021-12-24 at 2 37 44 AM

Araxeus:
* Update electron https://www.electronjs.org/blog/electron-16-0#end-of-support-for-12xy
which requires updating the remote package in this app as well as in custom-electron-prompt and custom-electron-titlebar
which caused some downloader bugs that needed fixing in commit baeebd1

* Fix for logging in when account is "shadowbanned" (browser not secure error)

fix #507, fix #526, fix #17, fix #8, fix #327


WhiteHusky:
Logging in is still pretty jank. I had to clear my data (or as much as I could), then try to sign in, then switch into my own account, then hard refresh. The home page is, for some reason, still showing the generic home page but now I get access to my library.

I swore there was Chromecast support but I guess I might've dreamt it up.

WhiteHusky:
And now opening and closing it breaks it again. Google really doesn't want us embedding YTM huh?

Araxeus:
It really should work now, did you try building and installing?

KimBum:
Login worked for me instantly. Although there were no navigation items except for search, after returning back to home page.
Also login screens looked a bit different (old?) compared to what I am currently used to with Google/Youtube.
The package versions you updated here, are the newest ones, right? Or are there any other parameters that might've been missed?

KimBum:
Okay nvm. They are having the same issue now. Seems like this is a common problem suddenly appearing with Electron apps, maybe?

KimBum:
Over there the fix also included opening their "guest" login separately without actually using it. That one does in fact look like the one that comes out after this PR here :)

image

KimBum:
Is it possible to force a restart after successful login in the second case? If so, I would say that it is an acceptable workaround.

On another note regarding this PR: a273f6f73c057453e88a72825cb133a4bcdfebd8 is out of scope in my opinion. Or does it have anything to do with the initial problem or resulting issues? I am guessing the simultaneous update to electron was needed for the remote package to work? (4 major versions is pretty big)

Araxeus:
> Is it possible to force a restart after successful login in the second case? If so, I would say that it is an acceptable workaround.

Latest version actually works good for me when built and installed (still buggy in dev mode)

----

>  a273f6f is out of scope in my opinion.

Well yes but its just that suddenly with the new electron version `player.videoMode_` sometimes doesn't reflect the real value when `videoStarted()` is called, so I just switched to the old way of checking

-----

> update to electron was needed for the remote package to work?

More like the update to the remote package was needed because of the electron update.
That update is for mainly for the 2 following reasons:
* Electron 12 will be obsolete at 16/12/2021 which is in 2 days (and with it the remote module)
* Update electron 16 since there shouldn't be any breaking changes affecting this project except the remote module and that way we get the latest chromium which performs better & more secure & less bugs

Araxeus:
@KimBum here are binaries of the latest changes (for windows)
https://www.mediafire.com/folder/fvjiihmu3ka2z/Youtube-Music+useragents-patch+v3

try installing via the setup this time

package.json
```diff
@@ -88,7 +89,7 @@
 		"ytpl": "^2.2.3"
 	},
 	"devDependencies": {
-		"electron": "^12.2.2",
+		"electron": "^16.0.2",
```
th-ch, line 93: This seems to break the downloader:
```
Error: unrecognized flag --experimental-wasm-bulk-memory
Try --help for options
[74932:1214/212918.545255:ERROR:node_bindings.cc(276)] Most NODE_OPTIONs are not supported in packaged apps. See documentation for more details.
```
And `SharedArrayBuffer is not defined` when converting - it should be fixable but might need additional work, maybe we can split the PR between fixing the user agents and the electron update?

KimBum:
Seems to work some of the time when the tooltip for "press to play" appears, but even though I just reinstalled it, the tooltip is not shown for the first startup 🤨

index.js
```diff
@@ -163,6 +165,30 @@ function createMainWindow() {
 }
 
 app.once("browser-window-created", (event, win) => {
+	// User agents are from https://developers.whatismybrowser.com/useragents/explore/
+	const originalUserAgent = win.webContents.userAgent;
+	const userAgents = {
+		mac: "Mozilla/5.0 (Macintosh; Intel Mac OS X 12.1; rv:95.0) Gecko/20100101 Firefox/95.0",
```
th-ch, line 172: Login seems to work fine on Mac with the fix and this user-agent: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.93 Safari/537.36`

th-ch (commented):
Thanks for the fix, tested on MacOS, looks promising! But I would separate the electron upgrade which might require more work

index.js
keliel: Would be willing to test a version with the electron upgrade reverted on Windows. As far as I understood that would also revert the remote package setup, right? So would like to re-check it once.

index.js
```diff
@@ -163,6 +165,30 @@ function createMainWindow() {
 }
 
 app.once("browser-window-created", (event, win) => {
+	// User agents are from https://developers.whatismybrowser.com/useragents/explore/
+	const originalUserAgent = win.webContents.userAgent;
+	const userAgents = {
```
th-ch, line 171: It might also be worth allowing an override in config for more flexibility :)

Araxeus:
@th-ch Sorry but I really can't separate it now since remote doesn't work for me anymore no matter what I do
-which means I can't even test another PR without the updated electron version

Whenever I'll have time I will try look into the bugs:
* disable-autoplay
* downloader plugins (`unrecognized flag --experimental-wasm-bulk-memory` +` SharedArrayBuffer is not defined` when converting)

and allow override of useragent in config for more flexibility

Would be nice if someone else could also work on this since im kinda swamped with work right now, can message me on discord whenever

package.json
Araxeus: baeebd1 Fixed both of those issues and 1 more :)

index.js
Araxeus: see my comment below

> this useragent isn't really doing much, the actual fix is that when it fails (some accounts are shadowbanned) - it tries again with the original useragent which does work currently if login initially fail
https://github.com/th-ch/youtube-music/blob/8f2ed3039a5f30eeaf372afc4bcb1105b080bee9/index.js#L185-L191

Araxeus:
Downloader issues are fixed

I've also fixed and updated both custom-electron-prompt&titlebar packages

All thats left is that disable-autoplay bug that I can't personally replicate. can anyone look into it?

index.js
Araxeus: @th-ch You can't really test this useragent fix unless you are shadowbanned

(which means you will see a message saying the login failed due to unsecure browser, then press retry and it will redirect to a different login page which should work)

th-ch (approved):
Looks good, thanks for the fix and the upgrade! Merging, will release a new version with all the recent changes

PR #519, opened 2021-12-14
Add "Skip silences" plugin

th-ch:
This PR adds a new plugin to skip silences in videos.


Araxeus:
This actually doesn't work for me.
`stopped_speaking` is never called

For example the last 10 seconds of this song https://music.youtube.com/watch?v=08jzDDxTO_4 aren't skipped

th-ch:
Weird, skipping logic is still experimental but the silence events are correctly emitted for me 🤔

Araxeus:
I have no idea what caused this issue for me but the events are emitted just fine now

But I've found and fixed 2 other bugs in #521

PR #521, opened 2021-12-18
fix skip-silences plugin

Araxeus:
This PR fixes 2 problems:

* Threshold was too high, making it skip quieter parts of songs
* it would always skip forward on song start / seeking


th-ch (approved):

PR #525, opened 2021-12-24
fix precise-volume options sync

Araxeus:
As title says, this PR mainly aims to fix an issue with precise-volume options sync between main and renderer
(for example changing keybinds (menu.js) and then changing the volume (front.js) would reset the keybind)

those 2 fixes below are also included:
* fixed an issue where the custom volume steps was interpreted as a string instead of a number
* slightly reduced the "cooldown" for saving the volume


th-ch (approved):

PR #537, opened 2022-01-03
fix notifications showing thumbnail of last song

Araxeus:
because of the delay in `await getImage()` in songInfo.js, if the video sends a pause/play event before `getImage()` finishes, it will send callbacks with the new songInfo - but with the old songInfo.image + everything that comes after that `await` in the code 

and thats because `handleData()` function updates *some* of the songInfo, then tries to download the image which takes a few miliseconds. so if the callbacks from playPause are triggered meanwhile, it will have new data from before the `await getImage()` but not the rest since it wasn't updated yet.

I hope this explanation was clear enough. if not then I can try explaining more clearly

### DIFF
--- a/providers/song-info.js
+++ b/providers/song-info.js
@@ -86,10 +86,14 @@ const registerCallback = (callback) => {
 	callbacks.push(callback);
 };
 
+let handlingData = false;
+
 const registerProvider = (win) => {
 	// This will be called when the song-info-front finds a new request with song data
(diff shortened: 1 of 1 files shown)


Araxeus:
On another note, for some reason this commit 4d4ac56 breaks the app for me
image

deleting the `NODE_OPTIONS= ` part fixes this

th-ch (approved):
For the `NODE_OPTIONS=` bug, it seems it only works on Unix systems, I'll give it a look!

PR #539, opened 2022-01-04
xesam:artist should be a list

markbaas:
According to mpris specs artists should be a list of strings:

https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#xesam:artist

Thus this was not working in gnome-shell :(


th-ch (approved):

PR #547, opened 2022-01-10
Fix the downloaded music file length

semvis123:
Uses ffmpeg to copy the file which fixes the music file.
This fixes the wrong length of music files as described in this issue, https://github.com/th-ch/youtube-music/issues/151 (at least when shown in Apple Music)
Doesn't re-encode the file so it should be quick, but it could also be implemented as an optional step (as not everyone experiences this problem).


Araxeus:
> could also be implemented as an optional step (as not everyone experiences this problem).

this should probably be done before merging (really easy to do too)

semvis123:
> this should probably be done before merging (really easy to do too)

Yeah true, but I was wondering if it should be optional or just always enabled (as it doesn't impact the speed to much).

Araxeus:
IMO since this is only needed to fix a bug for certain people it should be optional  + disabled by default

also any idea why copying the file fix the problem? seems weird

semvis123:
> IMO since this is only needed to fix a bug for certain people it should be optional + disabled by default

Alright I will make it optional.

> also any idea why copying the file fix the problem? seems weird

Don't exactly know why, but all (?) default codecs produce this length/duration issue on macOS (might be a bug in ffmpeg). lib-opus doesn't but that can't be played using iTunes/Apple-Music so I'm not sure. Also tried adding the duration metatag, but that didn't seem to fix the issue. 

Some related posts.
https://askubuntu.com/a/1000020
https://stackoverflow.com/questions/10437750/how-to-get-the-real-actual-duration-of-an-mp3-file-vbr-or-cbr-server-side


> also next time might be better to create a new branch from upstream/master for new changes (commit history looks weird + it says this branch is from 9 months ago)

oh definitely, didn't notice it at first.

semvis123:
Looks like this is a bug in Apple Music, the metadata is parsed incorrectly.
If I open the file with any other music player it's correct (even when using QuickTime or quick look)


Broken file information (shows double length in Apple Music):
```
Input #0, mp3, from ‘broken.mp3':
  Metadata:
    title           : song-title
    artist          : song-artist
  Duration: 00:06:04.01, start: 0.023021, bitrate: 134 kb/s
  Stream #0:0: Audio: mp3, 48000 Hz, stereo, fltp, 128 kb/s
    Metadata:
      encoder         : Lavc58.91
  Stream #0:1: Video: png, rgb24(pc), 720x720, 90k tbr, 90k tbn, 90k tbc (attached pic)
    Metadata:
      comment         : Cover (front)
```


File information after being fixed with ffmpeg (shows the correct length in Apple Music):
```
Input #0, mp3, from ‘fixed.mp3':
  Metadata:
    title           : song-title
    artist          : song-artist
    encoder         : Lavf58.76.100
  Duration: 00:06:04.01, start: 0.023021, bitrate: 135 kb/s
  Stream #0:0: Audio: mp3, 48000 Hz, stereo, fltp, 128 kb/s
    Metadata:
      encoder         : Lavc58.91
  Stream #0:1: Video: png, rgb24(pc), 720x720, 90k tbr, 90k tbn, 90k tbc (attached pic)
    Metadata:
      comment         : Cover (front)
```

Since this looks like a bug in Apple Music, I guess we don't need to include a patch in YouTube Music right?

Issue #542, opened 2022-01-04
App not showing when you run it


Starboy345:
I run the .exe file and previously, it worked fine. But now when i try open it, the icon shows up on my task bar but the app wont show on my screen. When i right click the tray icon and click show, it does not do anything. The app is playing music and i right click the tray icon and use the pause/play, previous, skip etc buttons.

I have tried deleting and downloading it again. The issue seems to be the same for v1.14 and v1.15 as well.

Any suggestions? Thanks

Araxeus:
try making a backup of your `config.json` and deleting that file

> config.json folder path:
> * Windows: `%AppData%/Youtube Music` (MyUser/AppData/Roaming/Youtube Music)
> * Linux: `~/.config/Youtube Music`
> * MacOS: `~/Library/Application\ Support/YouTube\ Music/config.json`

also note that v1.15 should be installed via setup.exe

Starboy345:
how do you create a backup of the config.json file (sorry, im not the most experienced in this kinda stuff)?

Araxeus:
just rename it to `config_BACKUP.json` and that should be enough.

you dont actually have to delete it sorry 😅

Starboy345:
i didnt delete it. I thought i would wait until i knew exactly what i was doing

but it did work! i renamed to config_BACKUP.json and the issue was sorted. thanks for your help!

also, is the download playlist function being fixed for v1.15 or is there a fix for it in v1.14?

Araxeus:
can you please post here this file? config_BACKUP.json?

please try v1.15 and tell if playlist downloading works for you, because it does for me :)

Starboy345:
it won't let me upload that file type unfortunately...

and you said i need to download the setup .exe for 1.15 and then run the .exe for the normal application?

Araxeus:
For future reference:
This problem was caused because `window-position` (x value in particular) in config.json was set to a negative value that was offscreen

 (reasons unkown but probably not related to this app since it never changes the window position on its own)

Issue #193, opened 2021-03-17
[Bug] No GUI/window appearing


IdmiDwebe:
I'm not sure exactly when it happened but no window is displayed anymore. 

Now, when I launch, the application shows in the taskbar and a standard window in the windows task switcher (the alt+tab one) with all the GUI as it should, but clicking on either of these doesn't bring up the window, although other windows do lose focus as they should. I can still control audio with the media keys and I can see what's playing through the task switcher or if I hover over it's icon on the taskbar.

I only have one Windows "desktop" active, and it doesnt seem related to the fact that I'm using dual monitors. Its worked fine on my system for months, and it wasn't caused by the latest update, since it was working at one point with this version.

It happened around the time that I installed the Vulkan SDK (latest version, haven't installed any Vulkan SDK before) on my system, but I'm not sure exactly of when it happened so it may be unrelated, I also updated and installed some Visual Studio components but apart from that I can't think of what the cause could be. Renaming the vulkan SDK folder then re-launching or using the portable version doesnt change anything.

I've tested on the latest version 1.11.0 (installer and portable)

IdmiDwebe:
For unrelated reasons I did a fresh install of Windows and re-installed the Vulkan SDK and it's working now.

IdmiDwebe:
Without logs or anything else to go on or proof that it wasn't something I did, I guess this issue is kind of useless, I'll just close it.

IdmiDwebe:
This issue has occurred again and it persists after installing on 1.15.1 and 1.14.1

Araxeus:
check out #542 

(might be a problem with config)

IdmiDwebe:
Thanks! 

Yep, 'x' in 'window-position' in 'config.json' was set to around 4000 (I have a 1080p monitor). Setting it between -1920 and 1920 made it visible on again.

Araxeus:
@th-ch could maybe add some check that x,y is inside the [primary display](https://www.electronjs.org/docs/latest/api/screen#screengetprimarydisplay) ? https://github.com/th-ch/youtube-music/blob/ec4c2e92af70b9e4cab180e26f65cf9f072773fb/index.js#L125-L128

PR #548, opened 2022-01-11
fix app starting offscreen

### LINKED ISSUE #517: Now the app itself is not opening
I submitted an issue saying that it's not downloading, now the app is closed and it's not opening. I tried restarting my laptop but it's not opening.
What do I do?

Araxeus:
Aims to fix issues such as #193 #542, probably fix #517

checks if the app is completely offscreen (x / width) or if the titlebar is offscreen (y / height) 
> checks if window is inside the display closest to the given point

**Tested on multiple monitor setup and works perfectly**

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,12 +122,16 @@ function createMainWindow() {
 		autoHideMenuBar: config.get("options.hideMenu"),
 	});
 	remote.enable(win.webContents);
+
 	if (windowPosition) {
 		const { x, y } = windowPosition;
-		if(x + win.getSize()[0] < 0 || x - win.getSize()[0] > electron.screen.getDisplayNearestPoint({x, y}).bounds.width) {
+		const winSize = win.getSize();
+		const displaySize = electron.screen.getDisplayNearestPoint(windowPosition).bounds;
+		if((x + winSize[0] < 0 || x -winSize[0] > displaySize.width) ||
+		   (y < 0 || y > displaySize.height)) {
 			//Window is offscreen
 			if (is.dev()) {
-				console.log(`Window tried to render offscreen, Width=${win.getSize()[0]}, nearestDisplayPointer.bounds=${electron.screen.getDisplayNearestPoint({x, y}).bounds}, position=${{x, y}}`);
+				console.log(`Window tried to render offscreen, windowSize=${winSize}, displaySize=${displaySize}, position=${windowPosition}`);
 			}
 		} else { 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -124,7 +124,14 @@ function createMainWindow() {
 	remote.enable(win.webContents);
 	if (windowPosition) {
 		const { x, y } = windowPosition;
-		win.setPosition(x, y);
+		if(x + win.getSize()[0] < 0 || x - win.getSize()[0] > electron.screen.getDisplayNearestPoint({x, y}).bounds.width) {
```
th-ch: No need to do the same with y/height?

Araxeus: Well usually the error occurs because x value is offscreen but just to make sure ill add it

Araxeus:
now also check if the titlebar y coordinate is offscreen (if its offscreen then there isn't a way to move the app)

Araxeus:
f5e22b5 made it so that the app validate the coordinates if they are inside the coordinates of the closest display to the given point

* get saved absolute coordinates (grid start point (0,0)  is the top left corner of **the primary monitor**)
* get monitor closest to said point
* check if given coordinates is inside the rectangle of the closest monitor

index.js
```diff
@@ -122,9 +122,20 @@ function createMainWindow() {
...
 	});
 	remote.enable(win.webContents);
+
 	if (windowPosition) {
 		const { x, y } = windowPosition;
-		win.setPosition(x, y);
+		const winSize = win.getSize();
+		const displaySize = electron.screen.getDisplayNearestPoint(windowPosition).bounds;
+		if((x + winSize[0] < displaySize.x - 8 || x - winSize[0] > displaySize.x + displaySize.width) ||
+		   (y < displaySize.y - 8 || y > displaySize.y + displaySize.height)) {
```
th-ch, line 131: Shouldn't it be `y + winSize[1]` similar to `x + winSize[0]`? (+  nit:the condition only contains `or` operators so inner parentheses can be omitted)

th-ch (commented):
Looks good, just one question before merging!

index.js
Araxeus: if y < displaySize.y then the titlebar is out of the screen and the app can't be controlled

y + winSize[1] would refer to the bottom corner, and its an error if the top corner is out of the screen (just y value)

also the parentheses could be indeed omitted, if you want

th-ch (approved):

PR #551, opened 2022-01-13
Fix duplicate media session on linux

### LINKED ISSUE #514: [Feature request] Is it possible to implement sharing the progress bar/album art using MPRIS?
I use KDE Connect (usually) to control audio players on my computer. So it would be very nice if there was a way to implement the sharing of the progress bar/album art via the MPRIS plugin.

Additionally, the music player is listed as "Chromium" when there already is a "YouTube Music" music player which seems to be a dummy audio player that is listed.

One music player which has this implemented is [MellowPlayer](https://gitlab.com/ColinDuquesnoy/MellowPlayer) for reference.

Araxeus:
On linux chromium shouldn't register its own mpris service if the app does it (currently bundled in the `shortcuts` plugin)

Fix #514 (the half that wasn't fixed in the last PR)

> thanks @JoeJoeTV for testing

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,6 +40,11 @@ if (config.get("options.disableHardwareAcceleration")) {
 	app.disableHardwareAcceleration();
 }
 
+if (is.linux() && config.plugins.isEnabled("shortcuts")) {
+	//stops chromium from launching it's own mpris service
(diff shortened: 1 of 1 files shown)


th-ch (approved):

PR #553, opened 2022-01-17
Build command for Apple (m1) silicon macs 

arunim2405:


th-ch (approved):
Looks good, I'll see to update CI to also release for arm64 :)

Issue #556, opened 2022-01-17
Some songs will not download


valkyrienyanko:
Some songs just say "Downloading..." forever, but other songs download just fine.

Steps to Reproduce
1. Search for `"big shot deltarune drum and bass version"`
2. Try downloading the first top result

OS: Windows 64 Bit

Araxeus:
There are actually 2 bugs here

1. The first search result which is "Top Result" has a very weird url
   * menu doesn't contain the song url like all other menus - instead containing some random playlist url
   * if opening the menu from **any other** place, the url contains the song id and doesn't include `?playlist=`

2. playlist folder has no name sanitation, meaning that if the playlist name has invalid characters, `mkdirSync()` will throw an error at https://github.com/th-ch/youtube-music/blob/44aa62c9c8e05cf07226433eb842ec2a1ebe7602/plugins/downloader/menu.js#L108

PR #557, opened 2022-01-18
filenamify playlist folder name

Araxeus:
Bug: playlist folder had no name sanitation, causing a silent error on [`mkdirSync()`](https://github.com/th-ch/youtube-music/blob/44aa62c9c8e05cf07226433eb842ec2a1ebe7602/plugins/downloader/menu.js#L108) when downloading playlists with invalid characters in the name

fix part of #556

> note: replacement *could be* `_` instead of ` `


th-ch (approved):

PR #561, opened 2022-01-20
fix "restart app on config change" option

### LINKED ISSUE #560: plugin conflicting "Video Playback"
Hey, sorry for last time (forgot to close the issue😅)

I found out that the "Video playback" plugin conflicts with "restart on config changes"

Switching between video and song works normal but when I select next song/video the app restarts with different playlist.
Everything works great when I turn off "restart on config changes"

I added a gif with that shows the program restarting. (the pink is just a background nothing to do with yt music : ) )

I tried looking into it myself but I can't seem to fully understand how everything works since the only langue I can understand a bit is lua (zero to nothing of scripting experience) and don't know where the location of "restart on config changes" is.

ezgif com-gif-maker (5)

Araxeus:
Fix #560 (see https://github.com/th-ch/youtube-music/issues/560#issuecomment-1017758481)

config was changing for many reasons that didn't warrant a restart (position, url, precise-volume, video-toggle)

so I made it so that only config change originating from a menu can trigger a restart

This was done by changing all current menu implementation which is abit messy, but I can't currently think of a cleaner way to fix this..

> ⚠️ this pr needs conflict solving with the other open prs (rename `setOptions` to `setMenuOptions` in #578 & #580)


th-ch (approved):
Looks good, thanks for the improvement on the option! ✅

PR #562, opened 2022-01-20
fix window position save spam

Araxeus:
Previously when moving/resizing the window, the position would get spam saved to config, this PR introduce a timer (1sec) to keep it from spamming,

this also fix window position being' forgotten' on maximizing (happened because a `move` event is triggered to the top left of the screen when maximizing the window, atleast on windows)

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,7 +172,6 @@ function createMainWindow() {
 		savedTimeouts[key] = setTimeout(() => {
 			config.set(key, value);
 			savedTimeouts[key] = undefined;
-
 		}, 1000)
 	}
 
--- a/index.js
+++ b/index.js
@@ -143,20 +143,36 @@ function createMainWindow() {
 
 	win.on("move", () => {
 		let position = win.getPosition();
-		config.set("window-position", { x: position[0], y: position[1] });
+		lateSave("window-position", { x: position[0], y: position[1] });
 	});
+
+	let winWasMaximized;
 
 	win.on("resize", () => {
 		const windowSize = win.getSize();
 
-		config.set("window-maximized", win.isMaximized());
-		if (!win.isMaximized()) {
-			config.set("window-size", {
+		const isMaximized = win.isMaximized();
+		if (winWasMaximized !== isMaximized) {
+			winWasMaximized = isMaximized;
+			config.set("window-maximized", isMaximized);
+		}
+		if (!isMaximized) {
+			lateSave("window-size", {
 				width: windowSize[0],
 				height: windowSize[1],
 			});
 		}
 	});
+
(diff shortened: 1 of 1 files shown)


Araxeus:
Sorry about the messy commit history, I fucked up and can't squash

index.js
```diff
@@ -142,22 +142,39 @@ function createMainWindow() {
...
+		if (!isMaximized) {
+			lateSave("window-size", {
 				width: windowSize[0],
 				height: windowSize[1],
 			});
 		}
 	});
 
+	let savedTimeouts = {};
+	function lateSave(key, value) {
```
th-ch, line 169: nit: the debouncing can be improved by applying immediately the save then waiting before another save occurs:
```suggestion
function debouncedSave(key, value, timeout = 1000){
    if (!savedTimeouts[key]) {
      config.set(key, value);
    }
    clearTimeout(savedTimeouts[key]);
    delete savedTimeouts[key];
    …
  };
}
```

th-ch (approved):

index.js
Araxeus: I don't understand how would that work?

how would the function know if it should immediately save or timeout save?

from what I understand from your proposition, it would always immediately saves which defeats the purpose of this

---- 
### & no benefits?
this pr was made because only the **last** move/resize event is important IMO

resize/move events are fired on every pixel that was moved/resized so I don't think its really useful to immediately save instead of waiting 1 second and then saving - since the first change isn't even important (currentPosition/size +- 1px)

Say you want to move the window somewhere / resize it to a certain size - hundreds of move/resize events could be fired but the only one that matter is the one which indicates the final position

if you think the delay is too much it could just be lowered (1 full second might be too much)

th-ch: The idea of debouncing is to save immediately if no save has been performed in the last second, otherwise wait (based on the timeout being set or not)  - here, unless I missed something, waiting each time has edge cases - eg if you quit the app just after having resized it, it won't be saved because the timeout did not finish.

Though I don't expect edge cases to be too much of a problem so I'm fine with merging as is :)

Araxeus: I think you might be misunderstanding how the move/resize events are fired

If a user resize his window by 50px - it will send 50 resize events.
if you immediately save the size on the first event you get, it will just be currentSize+- 1px (useless)
and the last event will **not** be saved immediately (because there were *49* events before it)
-so that edge case you talked about will still apply..

Anyways I don't think that edge case is very realistic, its pretty hard to resize/move the window and then close it in less than one second (also its a weird thing to do, even if you could)

also like I said above, the delay could be lowered to something like 750/500milisecs so that there is even less chance that this edge case will happen

PR #567, opened 2022-01-22
fix precise-volume hud positioning

Araxeus:
This change was needed for 2 reasons:
* Video.clientHeight update delay on `srcChanged` event:
https://github.com/th-ch/youtube-music/blob/89ea66ba2b425a3a797a9429746a83532bf548b2/plugins/video-toggle/front.js#L43
the following code was supposed to position the hud, but `video.clientHeight` was 0 (because it wasn't updated yet)
https://github.com/th-ch/youtube-music/blob/89ea66ba2b425a3a797a9429746a83532bf548b2/plugins/video-toggle/front.js#L96
fixed by setting a tiny delay (250ms) before changing the hud position

----

* hud sometimes mispositioned when video-toggle is off:
the active hud positioning was only inside the video-toggle plugin, so I moved it to the precise-volume plugin.
now if video-toggle is enabled - it will handle all hud positioning, and if it isn't, then precise-volume will handle it itself

### DIFF
--- a/plugins/precise-volume/front.js
+++ b/plugins/precise-volume/front.js
@@ -1,7 +1,7 @@
 const { ipcRenderer } = require("electron");
 const { globalShortcut } = require('@electron/remote');
 
-const { setOptions } = require("../../config/plugins");
+const { setOptions, isEnabled } = require("../../config/plugins");
 
 function $(selector) { return document.querySelector(selector); }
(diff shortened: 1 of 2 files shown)


th-ch (approved):
Looks ok, thanks for the improvement! ✅

PR #573, opened 2022-01-26
Use `center` alignment for lyrics text

lazerl0rd:
This seems to be the case for most music applications [that I've used] and also provides the benefit of "fixing" how lyrics appear in RTL languages (such as Arabic).


Araxeus:
Hey I took inspiration from your pr and fixed a few more things in lyrics plugin in #584

(I also included this center alignment change - that now also apply to the vanilla lyrics (that are generated automatically by youtube)

th-ch (approved):

PR #578, opened 2022-01-27
add single instance lock option

### LINKED ISSUE #576: Show window if app already running
I have the tray icon enabled. I sometimes close the main window by habit (clicking the x) and then click the icon in my launcher. This launches another instance of the application.  Idealy the currenly running instance is shown instead of opening a new one.

This could be solved by checking if an instance is running and showing it, or by allowing a "minimize on close" option.  If the window only minimized when closing the window, then clicking the app in the laucher would bring it forward (instead of launching a new instance)

Araxeus:
resolves #576

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,22 @@ unhandled({
 process.env.NODE_OPTIONS = "";
 
 const app = electron.app;
+// Prevent window being garbage collected
+let mainWindow;
+autoUpdater.autoDownload = false;
+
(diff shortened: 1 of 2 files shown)


Araxeus:
the label "Single instance lock" could be replace with "Single instance mode" or "Allow only one instance" or something simillar

bkanuka:
:pray: @th-ch

th-ch (approved):
Looks good, thanks for the option! ✅

PR #580, opened 2022-01-27
discord set inactivity timeout prompt

Araxeus:
replace the button "Set timeout time in config" that just opened config with a text editor with "Set inactivity timeout" which opens a prompt where users can input timeout in seconds

(custom-electron-prompt v1.4.1 fixed a bug that happened when the default value was 0)


th-ch (approved):
Looks ok ✅ (fixed the merge conflict by applying `setMenuOptions` when the timeout is set)

PR #586, opened 2022-01-30
Polish Discord RPC

xn-oah:
Removes the unnecessary date and amount of views on the largeImage, adds a new playing icon, and changes the old paused icon.


Araxeus:
bruh next time check to see if your changes even works? because this literally causes error on load...

* `songInfo.isPlaying` doesn't exist? 😮
* module.exports inside other exports obviously don't work which crashes the app...

also changing all the arrow function `x => y()` to `(x) => y()`  is completely pointless imo, you should add parentheses only if there are 2+ arguments

PR #587, opened 2022-01-30
Add album title to largeImage and change paused icon

xn-oah:
paused

### DIFF
--- a/plugins/discord/back.js
+++ b/plugins/discord/back.js
@@ -30,13 +30,14 @@ const resetInfo = () => {
 	info.ready = false;
 	clearTimeout(clearActivity);
 	if (dev()) console.log("discord disconnected");
-	refreshCallbacks.forEach((cb) => cb());
+	refreshCallbacks.forEach(cb => cb());
 };
 
 let window;
 const connect = (showErr = false) => {
 	if (info.rpc) {
-		if (dev()) console.log("Attempted to connect with active RPC object");
+		if (dev())
+			console.log('Attempted to connect with active RPC object');
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


Araxeus:
Will only correctly display album when available after #588 is merged

plugins/discord/back.js
```diff
@@ -4,8 +4,8 @@ const { dialog, app } = require("electron");
 
 const registerCallback = require("../../providers/song-info");
 
-// Application ID registered by @semvis123
-const clientId = "790655993809338398";
+// Application ID registered by @xn-oah
+const clientId = "937234785716285471";
```
th-ch, line 8: Does that need to be updated/can it change something for people relying on the plugin? If it's not necessary (which seems to be the case), better not updating it - a good alternative if you want to use another app ID is to make it a parameter in the plugin config, so that everyone is free to choose what they want to use :)

Araxeus: to use  a custom icon on discord presence - you need to register an app and upload assets
(see https://discord.com/developers/docs/topics/gateway#activity-object-activity-asset-image)

he uploaded a new icon (paused icon) to the app's assets on discord

`activityInfo.smallImageKey = "paused";` is the key of the paused icon he uploaded to the dashboard

and since he doesn't have access to the app registered by semvis, he registered a new app so that he can change the paused icon

th-ch (approved):
✅

Araxeus:
this clientId seems to not work anymore..
image

switching to the old clientId registered by semvis instantly works

th-ch:
@xn-oah is your app still correctly registered, or should we switch back to the previous client ID?

xn-oah:
> @xn-oah is your app still correctly registered, or should we switch back to the previous client ID?

Fin in #602, really sorry about that.

PR #588, opened 2022-01-30
get album name from DOM

Araxeus:
### Problem
`$('ytmusic-player-page').__data.playerPageWatchMetadata` isn't reliable at all

### Solution
Grabs album name from dom (available only if in "song" mode, else will be `undefined`)

* `$$(".byline.ytmusic-player-bar > .yt-simple-endpoint")` 
  gets the links in the footer which will be either [channel] for video mode or [artist, album] for song mode
image

* `.find(e => e.href?.includes("browse"))`
  this should only match album links
   * artist/channel link is `https://music.youtube.com/channel/UCxxxxxxxx` 
   * album link is `https://music.youtube.com/browse/MPREb_xxxxxxx`

> grabbing straight from DOM because I can't find the album name anywhere else now

### DIFF
--- a/providers/song-info-front.js
+++ b/providers/song-info-front.js
@@ -6,7 +6,8 @@ const config = require("../config");
 
 global.songInfo = {};
 
-function $(selector) { return document.querySelector(selector); }
+const $ = s => document.querySelector(s);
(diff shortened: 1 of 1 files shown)


th-ch (approved):
Looks good! ✅

PR #591, opened 2022-02-01
fix interactive notifications icon + exclude platform specific plugins from build

### LINKED ISSUE #479: Interactive Notifications not registering as Youtube Music
Using the regular Notification plugin works the same as before, appearing with the song name as a Youtube Music Notification in Windows 11.

Irregular: Checking the box for interactive notifications works, however the notification is registered under "SnoreToast" instead of Youtube Music, which seems to be stemming from the application not correctly creating the notification. Unsure if this was an issue before the latest version, as I only recently started using the notification plugin.

Araxeus:
### pass appID to node notifier in release mode (assuming release has updated binaries from #600)
Fix #479 (see my comment there for more info about upstream pr's that are almost one year old)

old behavior:
old behavior

new behavior:
new behavior

The problem was that setting an `appID` (which makes the icon and appName show up) would make the buttons not work (returning undefined instead of index of clicked button)
This was caused because of a problem in a dependency of a dependency (node-notifier->snoretoast)
and the pr with the fix has still yet to be merged many months later, so we now use a custom compiled snoretoast that include said fix

## package.json changes

* `taskbar-mediacontrol` plugin added only to windows distributions
* `touchbar` plugin added only to macOS distributions


plugins/notifications/snoretoast/interactive.js
```diff
@@ -1,7 +1,13 @@
...
-const notifier = require("node-notifier");
+const { notificationImage, icons } = require("../utils");
+const getSongControls = require('../../../providers/song-controls');
+const registerCallback = require("../../../providers/song-info");
+const WindowsToaster = require('node-notifier').WindowsToaster;
+
+const notifier = new WindowsToaster({
+    withFallback: true,
+    // see https://github.com/th-ch/youtube-music/pull/591
+    customPath: require("path").join(__dirname, 'snoretoast-x64.exe')
```
th-ch: The idea makes sense but I am not a fan of including arbitrary binaries directly in the source code of this app because they can't be verified + unless I missed something, that won't work on 32 bits, no? :)

Until the mentioned upstream PRs are merged, a workaround is to generate the custom builds in CI from a given SnoreToast version - this gives people a way to check and compile them themselves if they want - I opened https://github.com/th-ch/youtube-music/pull/600 for that - here are the generated files if you want to test with them:

[snoretoast-patched-0.8.0.zip](https://github.com/th-ch/youtube-music/files/8056235/snoretoast-patched-0.8.0.zip)

th-ch (commented):
Definitely looks good, just a comment about the binary file - a possible workaround is proposed in https://github.com/th-ch/youtube-music/pull/600 :)

th-ch (approved):
Looks good, thanks! Merging both this one and #600 ✅

Araxeus:
@th-ch 
Sorry I made a mistake when I moved `interactive.js` *after* testing

the commit below is the fix, could you apply it to the main branch? 😟

https://github.com/th-ch/youtube-music/commit/bd374d129328d070d7390091706d6bf1609fdb25

th-ch:
Done! https://github.com/th-ch/youtube-music/commit/6cad6871bfe8e1d94d63e74f6d28c49436865577 🚀

PR #595, opened 2022-02-05
make useragent override optional

### LINKED ISSUE #594: Cannot log in, "This browser or app may not be secure"
This is with 1.15.0, completely unable to log in no matter what I do.
image
image

Araxeus:
fix #594

make useragent override *optional* from advanced-options

could maybe be made enabled by default since it works for most users (store/defaults + migration)


th-ch (approved):
Looks good, thanks for adding the option ✅

PR #596, opened 2022-02-06
update in-app-menu

Araxeus:
* Big performance improvement
* Doesn't require electron@remote anwhere anymore
* Radio buttons now have their own icon (finally... 😅)
  image

tested with the changes in https://github.com/AlexTorresSk/custom-electron-titlebar/pull/175

----

also:
* adds a full screen button to the view submenu (which allows using the f11 shortcut)
* lighten up selected menuItem background color (better contrast/visibility against the mostly black menu)


Araxeus:
custom-electron-titlebar v4.1.0 was released, this pr should be safe to merge

th-ch (approved):
Looks good, thanks! ✅

PR #600, opened 2022-02-13
Add snoretoast custom compile script

th-ch:
This PR follows up on https://github.com/th-ch/youtube-music/pull/591 and patches SnoreToast to avoid including arbitrary binaries that cannot be verified in the source code - with caching, only the time to fetch from cache is necessary (a few seconds) - once https://github.com/KDE/snoretoast/pull/15 then https://github.com/mikaelbr/node-notifier/pull/375 get merged, this patching won't be necessary anymore.

todo: the Windows build command (`yarn build:win`) does not include these patched versions!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,60 @@ jobs:
       - name: Install dependencies
         run: yarn --frozen-lockfile
 
+      ######################
+      # Patch SnoreToast to fix App ID - see https://github.com/th-ch/youtube-music/issues/479#issuecomment-965473559
+      - name: SnoreToast - parameters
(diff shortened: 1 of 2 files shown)


Araxeus:
Is there any way to check if it works in development?

I'm not sure how this all works exactly but seems to me like `./node_modules/node-notifier/vendor/snoreToast/snoretoast` gets overwritten only when GitHub actions build the dist files?

th-ch:
Indeed! The CI build should also be ported to local `yarn` commands if people need it, in the meantime I linked them in https://github.com/th-ch/youtube-music/pull/591#discussion_r805392356 :)

PR #605, opened 2022-02-15
migrate from remote to ipc

Araxeus:
can only be after #596 (which removes remote from in-app-menu)


th-ch (approved):

Issue #613, opened 2022-02-20
(Plugin) Custom song volume if needed


Bellisario:
Recently I found a very beautiful song, but that it's too low of sound, so, would be great a plugin that permits a custom volume if needed to a song (for example, using a slider on the three dot menu, or something similar).

If you like my idea, please like  :+1:  below!!!

thijsi123:
You mean something like an amplifier?

AuracleTech:
Oh my gosh please yes the sound control sometimes jump around, its not precise, also it's impossible to go like 2% minimum is probably like 10%, youtube volume is so bad

Araxeus:
There are already 3 plugins for this...

+ `precise-volume` Control the volume precisely 

+ `exponential-volume`  Make volume slider exponential

+ `audio-compressor`  Apply compression to audio (similar to exponential-volume)

AuracleTech:
https://user-images.githubusercontent.com/40531871/155345671-7a20808e-eed9-4eef-9ee4-418e703b1822.mp4

AuracleTech:
Indeed it now seems to work, thank you 
![image](https://user-images.githubusercontent.com/40531871/155359474-b87e5b72-defc-4865-9254-ed3a3299ed0e.png)

PR #619, opened 2022-02-23
fix custom titlebar in prompt options

Araxeus:
fix https://github.com/th-ch/youtube-music/issues/613#issuecomment-1048876106
(prompt not working on windows/linux if in-app-menu isn't enabled)

Oversight on my part, new version of `custom-electron-titlebar` requires `setupMain()` to be executed before the titlebar can be used

Fixed by enabling the custom-titlebar in prompt options only if `in-app-menu` is enabled

(macOS check is still there because i'm not sure if it works ok there)

### DIFF
--- a/providers/prompt-options.js
+++ b/providers/prompt-options.js
@@ -1,17 +1,18 @@
 const path = require("path");
 const is = require("electron-is");
+const { isEnabled } = require("../config/plugins");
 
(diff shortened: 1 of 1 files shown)


th-ch (approved):
Looks good, at least on MacoS ✅

PR #670, opened 2022-04-06
[precise-volume] fix expand-volume-slider not updating its value

### LINKED ISSUE #669: Precise volume doesn't update the slider if it is hidden
Enabling precise-volume and using the scroll wheel to change volume doesn't update the volume slider if the window is small enough (horizontally) that the volume slider is hidden. I've attached a video so that it's a bit more clear.

https://user-images.githubusercontent.com/24532624/161965267-4744cb4c-f834-4112-9559-65f39efdf1da.mp4

Araxeus:
`#expand-volume-slider` wasn't getting updated when volume programmatically changed
fix #669

notes:
* no changes needed in `setupSliderObserver()` because manually moving the `#expand-volume-slider` automatically update the main `#volume-slider` which triggers the observer

* `showVolumeSlider()` wasn't updated because it would look weird and I see no point in automatically expanding it,

### DIFF
--- a/plugins/precise-volume/front.js
+++ b/plugins/precise-volume/front.js
@@ -189,8 +189,12 @@ function changeVolume(toIncrease, options) {
 
 function updateVolumeSlider(options) {
 	// Slider value automatically rounds to multiples of 5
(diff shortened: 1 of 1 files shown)


th-ch (approved):

PR #671, opened 2022-04-06
Fix lyrics genius missing parts

### LINKED ISSUE #664: Lyrics from Genius are incomplete
Comparing the lyrics in app to those on the Genius website, it seems they're cut off by the featured video. Everything after the video is missing within the app.

Araxeus:
Previously lyrics would come from `querySelector('[class^="Lyrics__Container"]')`, but genius sometimes have multiple lyrics div's for the same song, for example https://genius.com/Built-to-spill-velvet-waltz-lyrics

this PR change the lyrics to come from querySelectorAll and join their inner html with `<br>`

fix #664 

notes:

* `lyricsSelector1 = Array.from(querySelectorAll(...))....join(...)` is safe since
  * `querySelectorAll` returns an empty `NodeList` if none are found
  *  `Array.from(EmptyNodeList)` returns an empty array (switched to spread syntax but same behavior)
  * `EmptyArray.join(..)` returns empty string

* commit history is messy and should be squashed

### DIFF
--- a/plugins/lyrics-genius/back.js
+++ b/plugins/lyrics-genius/back.js
@@ -16,37 +16,35 @@ module.exports = async (win) => {
 			metadata.title
 		)}`;
 
-		let response = await fetch(
-			`https://genius.com/api/search/multi?per_page=5&q=${encodeURI(
-				queryString
-			)}`
-		);
-		if (!response.ok) {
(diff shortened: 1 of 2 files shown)


th-ch (approved):
Looks ok, thanks for the fix! ✅

Issue #656, opened 2022-03-24
[feature request] picture-in-picture support


ghalibansari:
please bring pip(picture-in-picture) or pop-up player support.

PR #674, opened 2022-04-07
Add "Picture in picture" plugin

th-ch:
This PR implements a raw version of PiP to resolve https://github.com/th-ch/youtube-music/issues/656 - first version re-uses the main window as having a new one requires way more work.

### DIFF
--- a/plugins/picture-in-picture/back.js
+++ b/plugins/picture-in-picture/back.js
@@ -23,7 +23,13 @@ const togglePiP = async (win) => {
 		win.setFullScreenable(false);
 		await win.webContents.executeJavaScript(
 			// Go fullscreen
-			`document.querySelector(".fullscreen-button").click()`
+			`
+			if (!document.querySelector("ytmusic-player-page").playerPageOpen_) {
+  				document.querySelector(".toggle-player-page-button").click();
+			}
+			document.querySelector(".fullscreen-button").click();
+			document.querySelector("ytmusic-player-bar").classList.add("pip");
+			`
 		);
 		win.setFullScreenable(true);
 
@@ -36,7 +42,10 @@ const togglePiP = async (win) => {
 	} else {
 		await win.webContents.executeJavaScript(
 			// Exit fullscreen
-			`document.querySelector(".exit-fullscreen-button").click()`
+			`
+			document.querySelector(".exit-fullscreen-button").click();
+			document.querySelector("ytmusic-player-bar").classList.remove("pip");
+			`
 		);
 
 		win.setVisibleOnAllWorkspaces(false);
--- a/plugins/picture-in-picture/back.js
+++ b/plugins/picture-in-picture/back.js
@@ -10,7 +10,7 @@ let originalPosition;
 let originalSize;
 
 const pipPosition = [10, 10];
-const pipSize = [400, 220];
+const pipSize = [450, 275];
 
 const togglePiP = async (win) => {
 	isInPiPMode = !isInPiPMode;
--- a/plugins/picture-in-picture/back.js
+++ b/plugins/picture-in-picture/back.js
@@ -2,6 +2,7 @@ const path = require("path");
 
 const { app, ipcMain } = require("electron");
 
+const { setOptions } = require("../../config/plugins");
 const { injectCSS } = require("../utils");
 
 let isInPiPMode = false;
@@ -13,6 +14,7 @@ const pipSize = [400, 220];
 
 const togglePiP = async (win) => {
 	isInPiPMode = !isInPiPMode;
+	setOptions("picture-in-picture", { isInPiP: isInPiPMode });
 
 	if (isInPiPMode) {
 		originalPosition = win.getPosition();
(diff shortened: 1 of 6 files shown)


Araxeus:
Doesn't really look like PiP to me, it looks like the native fullscreen mode (`f`) except its windowed

also it seems like pressing the "PiP" button cause the fullscreen mode to be windowed (but with the fullscreen ui)

so maybe this should just be an option (like "Windowed Fullscreen Mode") in the main menu? (Its more of a global toggle so I don't think it should be in the song menu)

There's Also an error/bug: ⚠️ 
trying to log in after having used "PiP" causes the app to freeze, console says
`TypeError: Cannot read properties of undefined (reading 'hide')
    at togglePiP (C:\Git\youtube-music\plugins\picture-in-picture\back.js:29:12)`

plugins/picture-in-picture/back.js
```diff
@@ -0,0 +1,58 @@
...
+		originalPosition = win.getPosition();
+		originalSize = win.getSize();
+
+		win.setFullScreenable(false);
+		await win.webContents.executeJavaScript(
+			// Go fullscreen
+			`document.querySelector(".fullscreen-button").click()`
+		);
+
+		app.dock.hide();
```
Araxeus:
```suggestion
		app.dock?.hide();
```

plugins/picture-in-picture/back.js
```diff
@@ -0,0 +1,58 @@
...
+		win.setVisibleOnAllWorkspaces(true, {
+			visibleOnFullScreen: true,
+		});
+		app.dock.show();
```
Araxeus:
```suggestion
		app.dock?.show();
```

plugins/picture-in-picture/back.js
```diff
@@ -0,0 +1,58 @@
...
+		win.setVisibleOnAllWorkspaces(false);
+		win.setAlwaysOnTop(false);
+	}
+
+	const [x, y] = isInPiPMode ? pipPosition : originalPosition;
+	const [w, h] = isInPiPMode ? pipSize : originalSize;
+	win.setPosition(x, y);
+	win.setSize(w, h);
+
+	win.setWindowButtonVisibility(!isInPiPMode);
```
Araxeus:
```suggestion
	win.setWindowButtonVisibility?.(!isInPiPMode);
```

Araxeus (changes requested):

Araxeus:
Ok sorry i'm just immediately writing as I'm thinking of stuff but ignore what I said about the size and name and all that

It just needs to behave correctly when entering fullscreen mode because right now thats buggy

also there should be a plugin option to set a hotkey for PiP

Good work 😄

Araxeus:
2 more changes related to other plugins that should maybe come in follow up PR?
* precise-volume hud position needs to be tweaked if both pip AND in-app-menu are active
  (the volume hud is positioned inside the in-app-titlebar somehow)
* since in-app-menu is contained within the window, it shows only the topmost options in this mode, thus the most important menu options should maybe be moved up (View-exit/restart) (Options-AlwaysOnTop)

Araxeus:
- [ ] right now its pretty annoying to exit the pip mode. there are 2 good solutions that I think could be both implemented:
  1. in `togglePiP` when PiP is activated ,register click event for t
   `$('.exit-fullscreen-button').addEvent -> onClick = togglePiP()` / `togglePictureInPicture()`
  and when its deactivated restore it
  2. like I wrote above, a custom keybind to toggle PiP

- [ ] on PiP start: if browserWindow is maximized, unmaximize it to avoid weird interaction

- [ ] some more css changes for more visibility in the playerbar:
  ```css
  ytmusic-player-bar.pip svg,
  ytmusic-player-bar.pip .time-info,
  ytmusic-player-bar.pip yt-formatted-string,
  ytmusic-player-bar.pip .yt-formatted-string {
	  filter: drop-shadow(2px 4px 6px black);
	  color: white !important;
	  fill: white !important;
  }
  ```
   - fix playerbar's artist/album string which instead of being `yt-formatted-string`  element they are links that have that as class.
   - fix `.timeInfo` which is a span that shows the current time (only when window is bigger than default PiP but still)
   - fix add `!important` to color to fix it not applying to artist name
   - fix `fill: white !important` mainly to fix `#button.dropdown-trigger` that didn't get filled by using `color

th-ch:
Interesting feedback, thanks for the involvement! The PR is getting big and it gives a first version of the plugin (not perfect but good enough for a first iteration) so I'll merge it as is, and we can follow up with the remaining improvements (which make sense)! 🚀

Araxeus:
>  I don't think that alwaysontop should be enabled by default since you can just enable it in the options
(maybe a plugin option to automatically enable alwaysontop but it shouldn't force this behavior)

^ do you agree with that?

should I open PR's about stuff I wrote since you closed this?

th-ch:
definitely, making it a user choice is better than forcing it!

Araxeus:
also im thinking maybe its better if the PiP button is inserted on the video overlay (like the "quality changer" button )
image
which would be way easier to access

(and then to exit that you could press the exit-fullscreen button i mentioned above / use the custom keybind)

PR #676, opened 2022-04-08
Add lyrics in metadata

th-ch:
This PR is a first version of having lyrics in metadata - right now, the lyrics are only set if the lyrics tab has been selected, otherwise the field is just empty (can be improved in the future). It also adds lyrics to downloaded files.

### DIFF
--- a/plugins/downloader/back.js
+++ b/plugins/downloader/back.js
@@ -71,6 +71,12 @@ function handle(win) {
 					description: ""
 				});
 			}
+			if (songMetadata.lyrics){
+				writer.setFrame("USLT", {
+					description: songMetadata.lyrics,
+					lyrics: songMetadata.lyrics,
+				});
+			}
 			writer.addTag();
 			fileBuffer = Buffer.from(writer.arrayBuffer);
 		} catch (error) {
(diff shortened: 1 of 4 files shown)


Araxeus:
I don't like the additional overhead on songInfo and the fact that the callbacks are called for that

I propose that instead of keeping the lyrics in songInfo we get them from lyrics-genius if active
```js
if (plugins.isEnabled('lyrics-genius')){
	const lyrics = await fetchFromGenius(`${metadata.artist} ${metadata.title}`); // needs a refactor first
	if (lyrics) {
		writer.setFrame("USLT", {
			description: songMetadata.lyrics,
			lyrics: songMetadata.lyrics,
		});
	}
}
```

th-ch:
Closing in favor of https://github.com/th-ch/youtube-music/pull/679

cc-gm:
I'm not a programmer, but I'm seeing that the lyrics that are downloaded with lyrics-genius (which are from a page of the same name) also contain the track number of the song (Track Number) and the year of the album (Year), yes You can consider that I add them to the metadata (ID3)

PR #679, opened 2022-04-09
Set lyrics metadata from Genius

th-ch:
Alternative implementation of https://github.com/th-ch/youtube-music/pull/676 - lighter but does not include lyrics in general metadata.


Araxeus:
Looks good to me 🥇

PR #680, opened 2022-04-09
[in-app-menu] add hide icon option

Araxeus:
* [feat: in-app-menu hideIcon option](https://github.com/th-ch/youtube-music/pull/680/commits/b4b785d77313a0ec5529f150171b316b675bfeab) :
 allows hiding the app icon from the titlebar (this + hideMenu option makes PiP mode really smooth looking)

* [fix: in-app-menu bg color edge case](https://github.com/th-ch/youtube-music/pull/680/commits/8000a8326fb77f81880818b4ce9c854a6ccbd2e3) :
  when elements weren't initially rendered because the window was too small, but were then accessed via the arrow keys - backgrounds were missing for the newly rendered elements

### DIFF
--- a/plugins/in-app-menu/front.js
+++ b/plugins/in-app-menu/front.js
@@ -3,7 +3,7 @@ const config = require("../../config");
 const { Titlebar, Color } = require("custom-electron-titlebar");
 function $(selector) { return document.querySelector(selector); }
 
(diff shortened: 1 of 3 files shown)


th-ch (approved):
Looks good, thanks!

PR #655, opened 2022-03-21
add always-on-top option

Araxeus:
Resolves #654

PR #685, opened 2022-04-10
Picture in Picture v2

### LINKED ISSUE #155: [Feature Request] Add Miniplayer as a feature
First off I want to say that this project is simply amazing. 

Before I stumbled upon this repo I downloaded the original YouTube Music Desktop application and saw that it had a Miniplayer feature. I think that adding such a capability to this already immensely better version of YouTube Music would make it incomparable to pretty much any other music application out there.

Araxeus:
continues #674
resolves #155

## Changes ✨ 
* enter PiP when clicking the miniplayer button on the video overlay (replace the native functionality since you can already go to miniplayer when clicking the arrow at the bottom left of the screen)

* Exit PiP when clicking the unmaximize button in the playerbar or pressing the <kbd>escape</kbd> key

* close expanding menu when mouse leave event is fired by removing the
  `no-cancel-on-outside-click` attribute and `$('#expanding-menu').onmouseleave = $('.middle-controls').click()`
  
 * exit fullscreen / maximized state and restore them on PiP exit
 
 * Remember PiP position/size options
 
 * disable PiP mode in config on app launch to avoid a bug where window position wasn't saved even tho PiP was disabled
   (happened if app was shut down while in pip mode, the config would still say pip is on)
   
 * disable double-clicking the video when in PiP mode (it quit fullscreen which looked weird)
 
 *  some more css changes to make the playerbar items more visible
  
 * add always on Top option 
 
 * add adaptors: (wasn't sure where to put this, might refactor)
   * in-app-menu: minify the menu while in PiP mode (custom mini menu since space is limited)
     
 * pretty awkward to use big code blocks inside `webContents.executeJavaScript` so it was all moved to front.js using ipcRenderer
   
 * added keybind option

 * fix precise-volume volumeHud position when in PiP using CSS

 * hide the video-toggle button while in PiP mode

 * add defaults and migration:
   ```js
   {
     "alwaysOnTop": true,
     "savePosition": true,
     "hotkey": "P"
   }
   ```

## Feedback Needed ❓ 

* should the PiP button be removed from the songMenu ?
  (because it's already super cluttered and there is an easy way of getting in and out now)

* should the PiP/AlwaysOnTop option replace the global option added in https://github.com/th-ch/youtube-music/pull/655 ?
  the plugin option works only in PiP mode, and the other one is globally active so I'm thinking of maybe leaving both, but its kinda weird


plugins/picture-in-picture/menu.js
```diff
@@ -0,0 +1,61 @@
...
+                        value: "hotkey",
+                        label: "Hotkey",
+                        default: options.hotkey
+                }],
+                ...promptOptions()
+            }, win)
+        
+            if (output) {
+                const { value, accelerator } = output[0];
+                console.log({ [value]: accelerator }); // DELETE
```
th-ch: nit: debug line?

th-ch (approved):

plugins/picture-in-picture/menu.js
Araxeus: 😅

PR #693, opened 2022-04-17
fix injectCSS `did-finish-load` listener overload

Araxeus:
Fix the following error message:
image


th-ch (approved):

PR #697, opened 2022-04-20
Video Toggle: Fix video display

MiepHD:
Changed margin-left to 25% to display it in the middle on top of the video.
Hide video toggle from youtube music.

### DIFF
--- a/plugins/video-toggle/button-switcher.css
+++ b/plugins/video-toggle/button-switcher.css
@@ -1,7 +1,6 @@
 #main-panel.ytmusic-player-page {
     align-items: unset !important;
 }
-
 .video-switch-button {
     z-index: 999;
     box-sizing: border-box;
@@ -77,4 +76,4 @@
 }
 #av-id {
 	display: none;
-}
+} 
--- a/plugins/video-toggle/button-switcher.css
+++ b/plugins/video-toggle/button-switcher.css
@@ -7,7 +7,7 @@
     box-sizing: border-box;
     padding: 0;
     margin-top: 20px;
-    margin-left: 10px;
+    margin-left: 25%;
     background: rgba(33, 33, 33, 0.4);
     border-radius: 30px;
     overflow: hidden;
@@ -74,4 +74,7 @@
     border-radius: 30px;
     transform: translateX(0);
     transition: transform 300ms;
+}
+#av-id {
+	display: none;
 }
--- a/plugins/video-toggle/button-switcher.css
+++ b/plugins/video-toggle/button-switcher.css
@@ -6,7 +6,7 @@
     box-sizing: border-box;
     padding: 0;
     margin-top: 20px;
-    margin-left: 25%;
+    margin-left: 10px;
     background: rgba(33, 33, 33, 0.4);
     border-radius: 30px;
     overflow: hidden;
--- a/plugins/video-toggle/front.js
+++ b/plugins/video-toggle/front.js
@@ -50,6 +50,9 @@ function changeDisplay(showVideo) {
     player.setAttribute('playback-mode', showVideo ? 'OMV_PREFERRED' : 'ATV_PREFERRED');
     $('#song-video.ytmusic-player').style.display = showVideo ? 'block' : 'none';
 	$('#song-image').style.display = showVideo ? 'none' : 'block';
+    if (showVideo && !video.style.top) {
+        video.style.top = `${(player.clientHeight - video.clientHeight) / 2}px`;
+    }
     moveVolumeHud(showVideo);
 }
 
--- a/plugins/video-toggle/front.js
+++ b/plugins/video-toggle/front.js
@@ -25,8 +25,6 @@ function setup(e) {
 
     $('ytmusic-player-page').prepend(switchButtonDiv);
 
-    $('#song-image.ytmusic-player').style.display = "block";
-
     if (options.hideVideo) {
         $('.video-switch-button-checkbox').checked = false;
         changeDisplay(false);
@@ -50,7 +48,8 @@ function setup(e) {
 function changeDisplay(showVideo) {
     player.style.margin = showVideo ? '' : 'auto 0px';
     player.setAttribute('playback-mode', showVideo ? 'OMV_PREFERRED' : 'ATV_PREFERRED');
-    $('#song-video.ytmusic-player').style.display = showVideo ? 'unset' : 'none';
+    $('#song-video.ytmusic-player').style.display = showVideo ? 'block' : 'none';
+	$('#song-image').style.display = showVideo ? 'none' : 'block';
     if (showVideo && !video.style.top) {
         video.style.top = `${(player.clientHeight - video.clientHeight) / 2}px`;
     }
--- a/plugins/video-toggle/front.js
+++ b/plugins/video-toggle/front.js
@@ -50,9 +50,6 @@ function changeDisplay(showVideo) {
     player.setAttribute('playback-mode', showVideo ? 'OMV_PREFERRED' : 'ATV_PREFERRED');
     $('#song-video.ytmusic-player').style.display = showVideo ? 'block' : 'none';
 	$('#song-image').style.display = showVideo ? 'none' : 'block';
-    if (showVideo && !video.style.top) {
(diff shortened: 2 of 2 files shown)


Araxeus:
25% isn't always the middle of the video, it depends on the window size

Here's a better way to center it:
in front.js:
```diff
- $('ytmusic-player-page').prepend(switchButtonDiv);
+ $('#main-panel').prepend(switchButtonDiv);
```
in button-switcher.css:
```diff
#main-panel {
    align-items: unset !important;
+   justify-content: center;
}
...
- margin-left: 25%;
```

MiepHD:
> also maybe I ask why would you want to use the video-toggle plugin if you already have a premium account and access to the native feature?
> is this plugin better than the native feature somehow?
The button is also displayed when you don't have youtube premium

Araxeus:
> The button is also displayed when you don't have youtube premium

That's weird... I've personally never seen this button 🤔

MiepHD:
I'm not sure if I did something wrong but when trying your way it gets like this.
image

MiepHD:
I'm not sure if I did something wrong but when trying your way it gets like this.
image

Araxeus:
hard to know what you did wrong when I can't see your code 😅 
is the video-toggle div inside `#main-panel` and
```css
#main-panel {
    align-items: unset !important;
    justify-content: center;
}
```
?
btw margin-top should also probably be removed at this point

MiepHD:
bug
For me, it somehow aligns on the wrong angle.

Araxeus:
Ok so I figured out what's happening here:

I don't have the `has-av-switcher` so my flex-direction is `row`
image
You do have the `av-switcher` (not sure why you do have it and I don't...)  so your flex-direction is `column`
image

can be probably fixed by adding to button-switcher.css:
```diff
#main-panel  {
+  flex-direction: row !important;
}
```

Araxeus:
your latest commit https://github.com/th-ch/youtube-music/pull/697/commits/f1cf94d8fa01de4ebf790e360d4e3bc487c3ba9a
breaks a fix that was made, why did you do it?

its a fix for a bug that is easy to reproduce:
change to song mode and restart the app, then change to video mode - the position of the video player is not correct

also did you look at what wrote in my previous comment? does it work for you?

MiepHD:
> this can probably be fixed by adding to button-switcher.css:

#main-panel  {
+  flex-direction: row !important;
}

if I add this the button is displayed like it's supposed to but I can't see the video.

Araxeus:
> bug For me, it somehow aligns on the wrong angle.

ok then instead of that, try adding the following in front.js
```js
$('ytmusic-player-page').removeAttribute('has-av-switcher');
```

note: I'm not sure if this attribute is dynamically updated - if it works try navigating/refreshing the page and see if breaks
(might have to add a mutationObserver)

MiepHD:
> your latest commit https://github.com/th-ch/youtube-music/commit/f1cf94d8fa01de4ebf790e360d4e3bc487c3ba9a
> breaks a fix that was made, why did you do it?
> 
> its a fix for a bug that is easy to reproduce:
> change to song mode -> restart the app -> change to video mode -> the position of the video player is not correct

I thought that this is no longer needed since it was displayed correctly. I added that the #song-image is hidden when changing to the video and is displayed again when changing to the image. Because there's no longer the #song-image in the div the size of the video is correctly displayed and don't has to be changed.
The bug is generated because  $('#song-image.ytmusic-player').style.display = "block"; set the son-image to block everytime you start the app. Then the video isn't displayed correctly because the video is together in the div with the image.

Araxeus:
> > your latest commit [f1cf94d](https://github.com/th-ch/youtube-music/commit/f1cf94d8fa01de4ebf790e360d4e3bc487c3ba9a)
> > breaks a fix that was made, why did you do it?
> > its a fix for a bug that is easy to reproduce:
> > change to song mode -> restart the app -> change to video mode -> the position of the video player is not correct
> 
> I thought that this is no longer needed since it was displayed correctly. I added that the #song-image is hidden when changing to the video and is displayed again when changing to the image. Because there's no longer the #song-image in the div the size of the video is correctly displayed and don't has to be changed. The bug is generated because $('#song-image.ytmusic-player').style.display = "block"; set the song-image to block everytime you start the app. Then the video isn't displayed correctly because the video is together in the div with the image.
> 
> (I also tested it in my version and couldn't reproduce this bug)

I wrote this after testing your changes and experiencing the bug myself, tho now after further testing it seems to happen only when the devtools is open inside the window
image

> I thought that this is no longer needed

so the only reason you removed it was because it wasn't needed, not because it broke something right?
if so then maybe just put it back in since it does fix some edge case?
if it does break something then I'm fine with removing it since it only affect an edge case

MiepHD:
I added it back in

MiepHD:
Now only the black bars are removed, so the video is displayed normally because the image gets hidden correctly.
And you can't see the original toggle anymore.

Araxeus:
Actually, it wasn't even about the devtools, it just made it more pronounceable (bigger height difference) but the bug occurred even with devtools off.
so good thing you added that back in

did you try the solution I wrote in https://github.com/th-ch/youtube-music/pull/697#issuecomment-1103920990 ?
if it works for you then please change in CSS:
```diff
.video-switch-button {
-  margin-top: 20px;
+  top: 20px;
}
```

MiepHD:
> did you try the solution I wrote in https://github.com/th-ch/youtube-music/pull/697#issuecomment-1103920990 ?

Yes I did, didn't work

Araxeus:
> Yes I did, didn't work

Why didn't it work? pretty sure its related to that attribute, is it still not showing the video?
if so can you check if both `$('ytmusic-player-page')` and `$('ytmusic-player')` indeed don't have the attribute?

I was really sure that its related to the inline CSS related to that attribute..
```css
ytmusic-player[video-mode_]:not([has-av-switcher]) #song-image.ytmusic-player,ytmusic-player:not([video-mode_]):not([has-av-switcher]) #song-video.ytmusic-player,ytmusic-player:not([playable_]) #song-image.ytmusic-player,ytmusic-player:not([playable_]) #song-video.ytmusic-player,ytmusic-player:not([playable_]) .song-media-controls.ytmusic-player {
  display: none;
}

ytmusic-player[has-av-switcher][playback-mode=OMV_PREFERRED] #song-image.ytmusic-player,ytmusic-player[has-av-switcher][playback-mode=ATV_PREFERRED] #song-video.ytmusic-player {
  display: none;
}

ytmusic-player-page:not([has-av-switcher]) #main-panel.ytmusic-player-page {
  flex-direction: row;align-items: center;
}

ytmusic-player-page:not([has-av-switcher]) .av.ytmusic-player-page {
  display: none;
}

ytmusic-player-page[has-av-switcher]:not([player-fullscreened_]) #player.ytmusic-player-page {
  max-height: calc(56.25vw - (var(--ytmusic-player-page-vertical-padding) + 32));
}

ytmusic-player-page[has-av-switcher] #player.ytmusic-player-page {
  flex: none;margin: auto 0;
}
```

Araxeus:
Aaaaand I just realized that THE WHOLE PLUGIN CAN BE REPLACED WITH 2 LINES
```js
$('ytmusic-player-page').setAttribute('has-av-switcher');
$('ytmusic-player').setAttribute('has-av-switcher');
```

and that's it. done, video toggle enabled...

Araxeus:
There isn't any reason not to use this instead of the current video-toggle plugin right?

kinda feels bad for all the hours I invested in this plugin, but why use a special plugin when the native implementation can be enabled?

MiepHD:
> There isn't any reason not to use this instead of the current video-toggle plugin right?

> kinda feels bad for all the hours I invested in this plugin, but why use a special plugin when the native implementation can be enabled?

For me, only the plugin works. The native implementation is displayed but you can't actually do anything. It just chooses the video if available if not it shows the picture.

Araxeus:
> For me, only the plugin works. The native implementation is displayed but you can't actually do anything. It just chooses the video if available if not it shows the picture.

Thanks for your input !

Because of it, I've decided against deleting the video-toggle plugin

#700 adds a mode selection to video-toggle plugin and also uses the fixes you added here (thanks for that too 🙂)

Issue #692, opened 2022-04-16
Video Toggle Issue


thijsi123:
The Video Toggle plugin interferes with the new feature baked into youtube music to toggle video or song only.
![image](https://user-images.githubusercontent.com/64159148/163676632-952ab4bb-624c-440c-b341-7aefa715355f.png)
Edit: The feature from youtube music itself doesn't work when the video toggle is enabled or disabled, So I have to use the video toggle plugin, which causes the video to be lowered.

Araxeus:
From what I understand this native feature is only available to Premium users.

So you can disable `video-toggle` plugin if you have a premium account

thijsi123:
> From what I understand this native feature is only available to Premium users.
> 
> So you can disable `video-toggle` plugin if you have a premium account

Ahh that's why it didn't work, thanks.

thijsi123:
Thanks for working on it : ).
I'll now close this issue.

Araxeus:
Hey, would you prefer the plugin or the native implementation?

Araxeus:
Because I'm pretty sure I can enable the default native implementation for everyone. 
which makes the `video-toggle` plugin kinda redundant?

thijsi123:
I'll prefer native.

thijsi123:
I think it's better that way now with youtube their own implementation.

Araxeus:
😭 all the hours I spent coding this plugin 😭

Araxeus:
Ok I decided against deleting the plugin, but there will now be an option called "Mode" which lets you choose between
* "Custom" - the plugin version
* "Native" - the native youtube implementation'
* "Disabled" - disable all video toggles

That pr will also fix this issue by automatically hiding native button if the custom button is enabled

PR #700, opened 2022-04-20
add different modes to video-toggle plugin

### LINKED ISSUE #631: Video Player UI messed up
The video panel has this massive black space above it and the video is being pushed down by it and then cut off at the bottom of the field. it looks like a video on a vertical screen where its being centered in the middle of the screen but the bottom half of the "screen" cant be shown This issue started with v1.15 and is the same in 1.16. I'm on Windows 10

Araxeus:
Modes:
* Custom: like before but disabled the native video-toggle + fixed the video position

* Native: use the native video-toggle

* Disabled: force disable the native video-toggle

fix #631
see https://github.com/th-ch/youtube-music/issues/692
thanks to @MiepHD in #697 for the fixes to the "custom" mode

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -80,6 +80,7 @@ const defaultConfig = {
 		},
 		"video-toggle": {
 			enabled: false,
+			mode: "custom",
 			forceHide: false,
 		},
 	},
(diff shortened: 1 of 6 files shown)


th-ch (approved):
Looks good, thanks for the improvement!

PR #749, opened 2022-06-14
Support MPRIS loop and volume change

foonathan:
This PR supports MPRIS loop status on linux, enabling `playerctl loop` to fetch and `playerctl loop None/Track/Playlist` to set.

### DIFF
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -19,7 +19,7 @@ function setupMPRIS() {
 
 function registerMPRIS(win) {
 	const songControls = getSongControls(win);
-	const { playPause, next, previous } = songControls;
+	const { playPause, next, previous, volumeMinus10, volumePlus10 } = songControls;
 	try {
 		const secToMicro = n => Math.round(Number(n) * 1e6);
 		const microToSec = n => Math.round(Number(n) / 1e6);
@@ -92,6 +92,22 @@ function registerMPRIS(win) {
 
 		player.on('seek', seekBy);
 		player.on('position', seekTo);
+
+		ipcMain.on('volumeChanged', (_, value) => {
+			player.volume = value;
+		});
+		player.on('volume', (newVolume) => {
+			// With keyboard shortcuts we can only change the volume in increments of 10, so round it.
+			const deltaVolume = Math.round((newVolume - player.volume) / 10);
+
+			if (deltaVolume > 0) {
+				for (let i = 0; i < deltaVolume; i++)
+					volumePlus10();
+			} else {
+				for (let i = 0; i < -deltaVolume; i++)
+					volumeMinus10();
+			}
+		});
 
 		registerCallback(songInfo => {
 			if (player) {
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -52,11 +52,11 @@ function registerMPRIS(win) {
 		player.on("loopStatus", (status) => {
 			// switchRepeat cycles between states in that order
 			const switches = ["None", "Playlist", "Track"];
-			const curIdx = switches.indexOf(currentLoopStatus);
-			const newIdx = switches.indexOf(status);
+			const currentIndex = switches.indexOf(currentLoopStatus);
+			const targetIndex = switches.indexOf(status);
 
 			// Get a delta in the range [0,2]
-			const delta = ((newIdx - curIdx) % 3 + 3) % 3;
+			const delta = (targetIndex - currentIndex + 3) % 3;
 
 			manuallySwitchingStatus = true;
 			songControls.switchRepeat(delta);
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -83,7 +83,7 @@ function registerMPRIS(win) {
 			}
 		});
 		player.on("playpause", () => {
-			player.playbackStatus = player.playbackStatus == 'Playing' ? "Paused" : "Playing";
+			player.playbackStatus = player.playbackStatus === 'Playing' ? "Paused" : "Playing";
 			playPause();
 		});
 
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -40,11 +40,11 @@ function registerMPRIS(win) {
 			if (manuallySwitchingStatus)
 				return;
 
-			if (mode == "Repeat off")
+			if (mode === "Repeat off")
 				currentLoopStatus = "None";
-			else if (mode == "Repeat one")
+			else if (mode === "Repeat one")
 				currentLoopStatus = "Track";
-			else if (mode == "Repeat all")
+			else if (mode === "Repeat all")
 				currentLoopStatus = "Playlist";
 
 			player.loopStatus = currentLoopStatus;
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -82,8 +82,11 @@ function registerMPRIS(win) {
 				playPause()
 			}
 		});
+		player.on("playpause", () => {
+			player.playbackStatus = player.playbackStatus == 'Playing' ? "Paused" : "Playing";
+			playPause();
+		});
 
-		player.on("playpause", playPause);
 		player.on("next", next);
 		player.on("previous", previous);
 
--- a/plugins/shortcuts/mpris.js
+++ b/plugins/shortcuts/mpris.js
@@ -2,6 +2,7 @@ const mpris = require("mpris-service");
 const { ipcMain } = require("electron");
 const registerCallback = require("../../providers/song-info");
 const getSongControls = require("../../providers/song-controls");
+const config = require("../../config");
 
 function setupMPRIS() {
 	const player = mpris({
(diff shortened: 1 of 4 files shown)


plugins/shortcuts/mpris.js
```diff
@@ -34,6 +34,35 @@ function registerMPRIS(win) {
...
+			// switchRepeat cycles between states in that order
+			const switches = ["None", "Playlist", "Track"];
+			const curIdx = switches.indexOf(currentLoopStatus);
+			const newIdx = switches.indexOf(status);
+
+			// Get a delta in the range [0,2]
+			const delta = ((newIdx - curIdx) % 3 + 3) % 3;
+
+			manuallySwitchingStatus = true;
+			songControls.switchRepeat(delta);
```
foonathan, line 63: This approach is a bit of a hack (it presses the button the appropriate number of times to go to the desired state), but I can't see a better way.

plugins/shortcuts/mpris.js
```diff
@@ -34,6 +34,35 @@ function registerMPRIS(win) {
...
+			player.loopStatus = currentLoopStatus;
+		});
+		player.on("loopStatus", (status) => {
+			// switchRepeat cycles between states in that order
+			const switches = ["None", "Playlist", "Track"];
+			const curIdx = switches.indexOf(currentLoopStatus);
+			const newIdx = switches.indexOf(status);
+
+			// Get a delta in the range [0,2]
+			const delta = ((newIdx - curIdx) % 3 + 3) % 3;
```
Araxeus: What is the point of the first modulo? `(newIdx - curIdx)` can only be in the range (-2, 2) then adding 3 is the exact result we need

Can be simplified to just
```js
const delta = (newIdx - curIdx + 3) % 3;
```

<details>
    <summary>See calc</summary>

0 = (0 - 0) % 3 + 3) % 3
0 = (0 - 0 + 3) % 3
---
2 = (0 - 1) % 3 + 3) % 3
2 = (0 - 1 + 3) % 3
---
1 = (0 - 2) % 3 + 3) % 3
1 = (0 - 2 + 3) % 3
---
1 = (1 - 0) % 3 + 3) % 3
1 = (1 - 0 + 3) % 3
---
0 = (1 - 1) % 3 + 3) % 3
0 = (1 - 1 + 3) % 3
---
2 = (1 - 2) % 3 + 3) % 3
2 = (1 - 2 + 3) % 3
---
2 = (2 - 0) % 3 + 3) % 3
2 = (2 - 0 + 3) % 3
---
1 = (2 - 1) % 3 + 3) % 3
1 = (2 - 1 + 3) % 3
---
0 = (2 - 2) % 3 + 3) % 3
0 = (2 - 2 + 3) % 3
---

```js
for (let a = 0; a <= 2; a++) {
    for (let b = 0; b <= 2; b++) {
        console.log(`${((a - b) % 3 + 3) % 3} = (${a} - ${b}) % 3 + 3) % 3`);
        console.log(`${(a - b + 3) % 3} = (${a} - ${b} + 3) % 3`);
    }
}
```

</details>

----

also, I personally think that short variable names aren't helpful, its better if they are clear and informative
could maybe be named `currentIndex` and `targetIndex` instead?

providers/song-controls.js
```diff
@@ -20,7 +20,10 @@ module.exports = (win) => {
 		go1sBack: () => pressKey(win, "h", ["shift"]),
 		go1sForward: () => pressKey(win, "l", ["shift"]),
 		shuffle: () => pressKey(win, "s"),
-		switchRepeat: () => pressKey(win, "r"),
+		switchRepeat: (n = 1) => {
+			for (let i = 0; i != n; ++i)
```
Araxeus: `++i` is useless here since the value isn't used
`i++` is more appropriate in this scenario (tho ultimately it's the same, the syntax is more clear)

also instead of `i != n` you should use `i < n` to avoid edge cases (this is standard)

plugins/shortcuts/mpris.js
Araxeus: I've tried making it work using the youtube api, but its weird and doesn't visually change the button state

So your solution is probably the best bet

plugins/shortcuts/mpris.js
foonathan: > What is the point of the first modulo? 

I originally had just `(newIdx - curIdx) % 3`, but that has the wrong sign. I then simply added the general code. Good catch.

providers/song-controls.js
foonathan: Funny. I come from a C++ background, and if someone wrote `for (auto i = 0; i < n; i++)` I would have requested it changed to the way I wrote it, as that's the standard there :D

Araxeus:
LGTM (but can't personally test it since I don't have access to Linux currently)

foonathan:
Anything I could do to help it get merged?

Araxeus:
@th-ch

foonathan:
I've just pushed another fix for a bug when using `playerctl play-pause`.

Araxeus:
In javascript we always use `===` instead of `==` 😋

th-ch (approved):
Looks ok overall, could be worth applying Araxeus' suggestion on using `===` - feel free to apply the change, otherwise I can merge and do the fix afterwards

foonathan:
I've changed it. Do you want me to clean up the commits and force push?

Araxeus:
you should also change it in plugins/shortcuts/mpris.js lines 43-45-47

foonathan:
Volume change support is now implemented as well.

providers/song-info-front.js
```diff
@@ -63,3 +65,23 @@ function setupTimeChangeListener() {
...
+
+	// Emit the initial value as well; as it's persistent between launches.
+	ipcRenderer.send('repeatChanged', $('#right-controls .repeat').title);
+}
+
+function setupVolumeChangeListener() {
+	const volumeObserver = new MutationObserver(mutations => {
+		ipcRenderer.send('volumeChanged', mutations[0].target.value);
+	});
+	volumeObserver.observe($('#right-controls .volume-slider'), { attributeFilter: ["value"] });
```
Araxeus: you should probably observe `$('video').volume` which is a more accurate representation of the current volume
(video volume could be changed directly without the slider moving which would cause bugs)

foonathan: That doesn't appear to give the correct values?

Araxeus: > That doesn't appear to give the correct values?

what makes you say that?

I'm pretty sure that `$('#right-controls .volume-slider').value` doesn't give the right volume and `$('video').volume` is the correct one

you can easily test this by changing `$('video').volume` and you'll see the video volume changes, while manually changing the volume slider value wont change the video volume

maybe you got confused because its a different format? 0.01/1 instead of 1/100

(`$('video')` is the actual video element in the page, so it makes no sense for its volume to have an 'incorrect' value)

foonathan: > maybe you got confused because its a different format? 0.01/1 instead of 1/100

Yea, in my quick test I got 0.01 instead of 100. I'll look into it again.

Araxeus: haha yeah so you can just `*100` it

foonathan: I can use `$('video').volume` to get the initial volume, but can't get events to trigger when it fires. I've tried the `MutationObserver` with a filter for `volume`, and the `volumechanged` event handler on the video element.

Is there anything that changes the volume without affecting the slider? Because there shouldn't be, right?

Araxeus: > Is there anything that changes the volume without affecting the slider? Because there shouldn't be, right?

No and there shouldn't be, but there can be, so best be avoided if possible
(for example, an earlier version of precise-volume changed the volume without affecting the slider)

> volumechanged event handler on the video element

that should work, I just tested on my side and it works, can you show me your code?

if you write this in devtools:
```js
$('video').addEventListener('volumechange', console.log);
```
doesn't it write a message on every change?

maybe you have a typo in the event name? https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/volumechange_event

foonathan: Okay, I don't know why it didn't work; it does now.

```js
$('video').addEventListener('volumechange', (_) => {
    ipcRenderer.send('volumeChanged', $('video').volume * 100);
});
ipcRenderer.send('volumeChanged', $('video').volume * 100);
```

However, instead of 100 or 30, I'll get values like 98.855310546902 or 29.656593164070603

Araxeus: ah yes the infamous rounding error... have to use `Math.round`
```js
$('video').addEventListener('volumechange', (_) => {
    ipcRenderer.send('volumeChanged', Math.round($('video').volume * 100));
});
```

foonathan: Then it's still 99 instead of 100? Why is this more inaccurate?

Araxeus: I dunno since I can't reproduce this... actually doing ```$('video').volume * 100``` always give me a round number without needing a rounding function

anyways there is another way to do this using `api.getVolume()`
```js
..
if (config.plugins.isEnabled('tuna-obs') ||
			(is.linux() && config.plugins.isEnabled('shortcuts'))) {
			setupTimeChangeListener();
			setupRepeatChangeListener();
			setupVolumeChangeListener(apiEvent.detail);
		}
...
function setupVolumeChangeListener(api) {
	 $('video').addEventListener('volumechange', (_) => {
	    ipcRenderer.send('volumeChanged', api.getVolume());
	});
	// Emit the initial value as well; as it's persistent between launches.
   ipcRenderer.send('volumeChanged', api.getVolume());
}
```

foonathan: Okay, I've implemented it that way; works.

foonathan:
Can this be merged?

th-ch (approved):

PR #767, opened 2022-06-28
[DEPRECATED] [Proposal] API for control and get informations

mesmerx:
its just a proposal for a internal API for yt-music

tomorrow i'll write better the approach and the details for this idea

(its just the basic of basic and just a  draw of the ideia, fell free to comment and discuss)


JellyBrick:
#1909

Issue #785, opened 2022-07-21
[Feature request] Config option for css styling


DrunkenToast:
I know you can create a plugin but it would be convenient if you can simply point to a css file(s) in the config file.
This would also make it easier for users would don't want to build it.

th-ch:
Hey @DrunkenToast, thanks for the feedback! The feature should actually be straightforward to implement (other parts of the code already inject CSS), I opened https://github.com/th-ch/youtube-music/pull/800 to allow users to pass custom CSS files:
```json
"options": {
    …
    "cssFiles": [
        "path-to-file-1.css",
        "path-to-file-2.css"
    ]
}
```

th-ch:
I renamed the option to `themes`, and there is now a menu entry to set it! PR is merged, it will ship in the next version - closing the issue, feel free to re-open if needed :)

PR #800, opened 2022-08-21
Allow user to pass custom CSS file

th-ch:
Simple approach to implement https://github.com/th-ch/youtube-music/issues/785 (allowing the user to pass custom CSS) - an improvement could be to allow passing the CSS as a string directly in config.


DrunkenToast:
Looks great.
Don't think a string is necessary, however a default css folder with the option to open said folder inside the app could be good too.

However this would be perfect enough for me.

PR #802, opened 2022-08-23
proposal: Adding an option to hide duration before the song ends

amsyarasyiq:
It bothers me sometimes when the countdown gets out of the album's "horizontal area", and I think it's more cleaner to remove it.
Therefore, I'm proposing an option to hide the countdown (disabled by default).

image
image


th-ch (approved):
Looks good, thanks!

PR #807, opened 2022-08-31
Update README.md with a new theme repo

kerichdev:
The repo referenced is currently unmaintained, so I made a fork with some fixes and improvements, with more to come. Maybe a good idea to reference it as well / replace it?


th-ch (approved):
Looks good, thanks!

PR #817, opened 2022-09-05
fix malformed json in tuna-obs

anytarseir67:
the tuna web server uses `cover` for the songs cover art, where as we are sending it under `cover_url`.

this lead to tuna failing to find the cover art.

i'm not 100% certain, but i believe this was changed in a somewhat recent version of tuna, so it might require a more indepth fix to stay compatible with older versions.

### DIFF
--- a/plugins/tuna-obs/back.js
+++ b/plugins/tuna-obs/back.js
@@ -6,6 +6,7 @@ const registerCallback = require("../../providers/song-info");
 const secToMilisec = t => Math.round(Number(t) * 1e3);
 const data = {
 	cover: '',
+	cover_url: '',
 	title: '',
 	artists: [],
 	status: '',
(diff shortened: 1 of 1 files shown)


plugins/tuna-obs/back.js
```diff
@@ -41,7 +41,7 @@ module.exports = async (win) => {
 
 		data.duration = secToMilisec(songInfo.songDuration)
 		data.progress = secToMilisec(songInfo.elapsedSeconds)
-		data.cover_url = songInfo.imageSrc;
+		data.cover = songInfo.imageSrc;
```
th-ch, line 45: Could we set both cover and cover_url? This way, it would support all versions (I guess additional properties are not an issue?)

Liknox (approved):

th-ch (approved):

PR #822, opened 2022-09-09
Fix likes on touchbar (they were inverted)

andrew-mathieu:
When we tap on "👍", it doesn't leave a like but does the opposite

### DIFF
--- a/plugins/touchbar/back.js
+++ b/plugins/touchbar/back.js
@@ -59,7 +59,7 @@ const touchBar = new TouchBar({
 });
 
 module.exports = (win) => {
-	const { playPause, next, previous, like, dislike } = getSongControls(win);
+	const { playPause, next, previous, dislike, like } = getSongControls(win);
 
 	// If the page is ready, register the callback
 	win.once("ready-to-show", () => {


anytarseir67:
> When we tap on "👍", it doesn't leave a like but does the opposite

can't reproduce

th-ch (approved):
Looks good, thanks!

l-laziz:
The issue still exists in v.19 - buttons still inverted.

afairb:
Looks like the change needed is on [line 66](https://github.com/andrew-mathieu/youtube-music/blob/b4ec6a791d5adfef6738cf0f60575c6491d43543/plugins/touchbar/back.js#L66) of the file (`plugins/touchbar/back.js`) instead:

```diff
-		controls = [previous, playPause, next, like, dislike];
+		controls = [previous, playPause, next, dislike, like];
```

I don't have time to make a PR to this repo at the moment sorry, but thought I'd highlight it in case anyone else wants to.

Araxeus:
Lol this PR only changed the import deconstruction order which is essentially only changing how the code looks 😂

Zo-Bro-23:
Done (#968)!

PR #840, opened 2022-09-22
fix listen along plugin

Dushuav2:
(from what I understand this would require the app to register itself as a handler for all https link which is a problem)

https://electronjs.org/docs/latest/api/protocol

https://electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app


th-ch (commented):
There seems to be a lot of commits unrelated to the change, can you rebase your PR?

th-ch:
Closing as this PR has drifted and now includes many commits from other branches - feel free to re-open another one with the relevant changes!

Issue #863, opened 2022-10-17
[Feature Request] Option to change captions


Marxcx:
Sometimes captions are on by default, but I'd like the option to turn them off or change the language as you can on the native youtube player.

btw I love the fact I stumbled upon this app, big thank you to all who contribute <3

Marxcx:
Thank you so much!

PR #866, opened 2022-10-22
Add Captions selector

LetrixZ:
#863 

Adds ability to select and disable captions, and an option to disable them by default.
Uses IFrame API module `captions`.

image

image


Marxcx:
Thank you so much!

plugins/captions-selector/front.js
```diff
@@ -0,0 +1,60 @@
+const { ElementFromFile, templatePath } = require("../utils");
+const { ipcRenderer } = require("electron");
+
+function $(selector) { return document.querySelector(selector); }
```
th-ch, line 4: Note for later (no change needed): this function is used in multiple plugins and can probably be refactored

plugins/captions-selector/front.js
```diff
@@ -0,0 +1,60 @@
...
+        if (captionTrackList?.length) {
+            const currentCaptionTrack = api.getOption("captions", "track");
+            const currentIndex = captionTrackList.indexOf(captionTrackList.find(track => track.languageCode === currentCaptionTrack.languageCode));
+
+            const captionLabels = [
+                ...captionTrackList.map(track => track.displayName),
+                'None'
+            ];
+
+            ipcRenderer.invoke('captionsSelector', captionLabels, currentIndex).then(promise => {
```
th-ch, line 45: nit: this could use async/await

plugins/captions-selector/templates/captionsSettingsTemplate.html
```diff
@@ -0,0 +1,13 @@
+<tp-yt-paper-icon-button class="player-captions-button style-scope ytmusic-player" icon="yt-icons:subtitles"
```
th-ch, line 1: note on naming: it's generally advised to use non-capital letters for file names (eg `captions-settings-template.html`) to avoid bugs on case insensitive OSes.

th-ch (approved):
Some minor comments but otherwise looks good, thanks!

th-ch:
Merging as is, will make a follow up PR for the nits

Araxeus:
Hmm this doesn't seem to work for me. Clicking the button doesn't do anything.

LetrixZ:
> Hmm this doesn't seem to work for me, clicking the button doesn't do anything?

Could you tell me what platform are you testing on (Windows, macOS, Linux) and, if possible, the video you used? Thanks.

Araxeus:
> If the video doesn't have captions, the button shouldn't open anything.

That's probably the issue here. Maybe the button shouldn't show if there aren't any captions available.

Also could you please post an example url that has captions? Because all the songs I tried don't have any

LetrixZ:
> Also could you please post an example url that has captions? Because all the songs I tried don't have any

Only YouTube videos have captions. Music "videos" do not and it isn't possible to bring those from the linked YouTube video because:
1 - There is not a way to easly get the linked video.
2 - You would need to modify the YouTube player to be able to inject captions that aren't there. This plugin relies on the player's built-in functions.

Take Kikuo - Yoru no Uta for example:
Video with captions: https://music.youtube.com/watch?v=Kv5-7kTp9m8
Music "video": https://music.youtube.com/watch?v=9mWMcy6CBKE

Araxeus:
Thanks, those examples helped me fix stuff in #1063

PR #873, opened 2022-11-06
docs: Added winget install instructions

Nixxen:


Nixxen:
Note: Windows might block the installation since it is from an unknown publisher. This is also true for the manual installation when trying to run the executable(.exe) after a manual download here on github (same file). Not sure if there is any way to bypass this without the users consent.
Due to this, this might not be worth adding to the readme, as it could cause confusion.

readme.md
```diff
@@ -46,6 +46,12 @@ scoop bucket add extras
 scoop install extras/youtube-music
 ```
 
+Alternately you can use [Winget](https://learn.microsoft.com/en-us/windows/package-manager/winget/), Windows 11s official CLI package manager to install the `th-ch.YouTubeMusic` package.
```
th-ch, line 49: Change looks good - should we add the warning you mentioned in the PR as well?

Nixxen: It might be worth mentioning for anyone with trouble installing on Windows. 

On the other hand, it might be better to attempt to get the files verified by Microsoft. I briefly looked into it, and it seems there are four main ways to get verified. https://stackoverflow.com/questions/48946680/how-to-avoid-the-windows-defender-smartscreen-prevented-an-unrecognized-app-fro

From the top SO response:
- submit your app for malware analysis to Microsoft,
- buy an "Extended Validation" code signing certificate,
- buy a standard code signing certificate, or
. just wait for a long time.

The SO response goes into detail about the different steps. I'd suggest using the file submitter as a software developer.

Adding the note to the readme is a decent temporary solution until the releases gets verified.

Nixxen: Added the note to the readme, so it's ready for merge.

th-ch (approved):

PR #888, opened 2022-11-20
Discord Plugin RPC Fix

Zo-Bro-23:
Update to new client ID so that the Discord Plugin works properly
Thanks to @zoeeechu for the fix!

### DIFF
--- a/plugins/discord/back.js
+++ b/plugins/discord/back.js
@@ -4,7 +4,7 @@ const { dialog, app } = require("electron");
 
 const registerCallback = require("../../providers/song-info");
 
-// Application ID registered by @xn-oah
+// Application ID registered by @Zo-Bro-23
 const clientId = "1043858434585526382";
 
 /**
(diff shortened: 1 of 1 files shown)


th-ch (commented):
Change looks good, but for context, why is it needed to change it? (also `Application ID registered by @xn-oah` is it still the case?)

Zo-Bro-23:
The application ID registered by @xn-oah doesn't seem to work anymore, so this is an application ID registered by @zoeeechu. It doesn't matter who registers the application ID as long as it's a valid ID, so it's fine either way, but you can edit my commit with your own application ID if you want more control over the Discord plugin (picture, application name, etc). Just head over to the [Discord Developer Portal](https://discord.com/developers) and register an application (and paste the client ID here).

Zo-Bro-23:
On second thoughts, I've added a commit where I've updated the application ID to an application that I created, so that the image and name syntax reflect the convention used for the application (@zoeeechu had lowercase with no image). I've also updated the comments to reflect this change. Let me know if anything else needs to be changed!

th-ch (approved):

PR #894, opened 2022-11-22
Center toggle of video-toggle

MiepHD:
Appended video-toggle to `#main-panel` so it can be centered over the video.
Centered it through setting left to `50% - 120px` and set position of `#main-panel` to relative that it uses this to center.

### DIFF
--- a/plugins/video-toggle/button-switcher.css
+++ b/plugins/video-toggle/button-switcher.css
@@ -2,6 +2,10 @@
     align-items: unset !important;
 }
 
+#main-panel {
+    position: relative;
+}
+
 .video-switch-button {
     z-index: 999;
     box-sizing: border-box;
@@ -18,6 +22,7 @@
     color: #fff;
     padding-right: 120px;
     position: absolute;
+    left: calc(50% - 120px);
 }
 
 .video-switch-button:before {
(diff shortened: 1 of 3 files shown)


plugins/video-toggle/button-switcher.css
```diff
@@ -18,6 +22,7 @@
     color: #fff;
     padding-right: 120px;
     position: absolute;
+    left: calc(50% - 120px);
```
th-ch: It works but (opinionated) I preferred having it on the side, it covers less the video 😄 Maybe we can have it as an additional css file, and make it a plugin option to center or not the toggle? This way users have a choice

MiepHD:
You can now choose between left, middle and right

th-ch (approved):

Araxeus:
This can possibly break the plugin (and it did for me, multiple time
`Cannot read properties of undefined.`
`mainpanel.style.setProperty("--align", "0px");`

this whole code block needs to be inside an `apiLoaded` closure, we can't guarantee that `$('#main-panel')` exists when this code is called

In other words, instead of being inside the default export function, it should be inside the `setup` function which is called only when the page is loaded

@MiepHD @th-ch

MiepHD:
Hm.. that's right. I'll correct that.

Issue #559, opened 2022-01-20
Unable to build app using Yarn


VBtheFirbolg:
OS: Ubuntu 20.04 LTS

I tried to build the app from the master branch using Yarn, but I am unable to do so. I get an error that states:

```
• electron-builder  version=22.11.7 os=5.4.0-94-generic
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=dist/builder-effective-config.yaml
  ⨯ Cannot find module 'dmg-license'
Require stack:
- /home/user/youtube-music/node_modules/dmg-builder/out/dmgLicense.js
- /home/user/youtube-music/node_modules/dmg-builder/out/dmg.js
- /home/user/youtube-music/node_modules/dmg-builder/out/dmgUtil.js
- /home/user/youtube-music/node_modules/app-builder-lib/out/macPackager.js
- /home/user/youtube-music/node_modules/app-builder-lib/out/packager.js
- /home/user/youtube-music/node_modules/app-builder-lib/out/index.js
- /home/user/youtube-music/node_modules/electron-builder/out/builder.js
- /home/user/youtube-music/node_modules/electron-builder/out/cli/cli.js
- /home/user/youtube-music/node_modules/electron-builder/cli.js  failedTask=build stackTrace=Error: Cannot find module 'dmg-license'
Require stack:
- /home/user/youtube-music/node_modules/dmg-builder/out/dmgLicense.js
- /home/user/youtube-music/node_modules/dmg-builder/out/dmg.js
- /home/user/youtube-music/node_modules/dmg-builder/out/dmgUtil.js
- /home/user/youtube-music/node_modules/app-builder-lib/out/macPackager.js
- /home/user/youtube-music/node_modules/app-builder-lib/out/packager.js
- /home/user/youtube-music/node_modules/app-builder-lib/out/index.js
- /home/user/youtube-music/node_modules/electron-builder/out/builder.js
- /home/user/youtube-music/node_modules/electron-builder/out/cli/cli.js
- /home/user/youtube-music/node_modules/electron-builder/cli.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
[...]

PR #905, opened 2022-12-03
Update readme.md

Zo-Bro-23:
Updated README with more precise build instructions (#559)


Zo-Bro-23:
Also added a GitHub Actions workflow to build the app. I'm only familiar with Windows, so it only supports Windows at the moment, but feel free to add jobs for Mac and Linux too!

readme.md
```diff
@@ -176,10 +176,13 @@ module.exports = () => {
 - communicating between the front and back: can be done using the ipcMain module from electron. See `utils.js` file and example in `navigation` plugin.
 
 ## Build
-
-```sh
-yarn build
-```
+1. Clone the repo
+2. Run ```yarn```
+3. Run ```npm i rimraf```
```
th-ch: rimraf is already a dependency so this step should not be needed!

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
+on:
```
th-ch: It seems this will do the same as the build workflow, no? (it already takes care of publishing the artifacts to GH releases)

th-ch (commented):
Readme change looks good (apart from one line), the rest should not be needed, unless I missed something?

Zo-Bro-23:
There seems to be an error regarding `rimraf`, which I wasn't able to quite pinpoint. Any build is failing, and removing `rimraf` from the build script seems to fix it. Sometimes, running `npm i rimraf` fixes the issue, but other times, it still persists. This issue is the reason for both the changes you mentioned. The GH Actions script I wrote doesn't use `rimraf`, because it's not necessary to delete the folder `dist` when running on Actions (because the folder doesn't exist). The `build.yml` workflow also has some other bugs, which means that it fails. Shall I replace the `build.yml` workflow with my own code from the `release.yml` workflow (Linux and Windows at least)?

Also, can you approve the change to `package.json` adding the two new scripts (`build:winactions` and `build:linuxactions`)? Those run without `rimraf`, so if you approve those, then I can remove the `npm i rimraf` line from the README.

DaveThuet:
**Hi...**
I have an idea or suggestion for the app.
How about if there was a shortcut for liking and disliking the playback.
**Best Regards**
_(With Google Translate from German to English)_

Zo-Bro-23:
For the time being, if you don't approve of the changes to `release.yml` and `package.json`, can you cherrypick the changes to `readme.md` and merge only that? Thanks!

readme.md
Zo-Bro-23: > rimraf is already a dependency so this step should not be needed!

Removed this line!

PR #936, opened 2022-12-30
Automatic release by CI when version is updated

th-ch:
This PR makes releases automatic by CI on the master branch when the version is updated, making the draft release a production one, and updating the changelog at the same time.


Zo-Bro-23:
Awesome!

Issue #945, opened 2023-01-07
[Feature Request]: Fade in/out (crossfade)


Zo-Bro-23:
Spotify has an _**awesome**_ feature for their mobile apps where you can choose to "crossfade" between songs. The last x seconds of a song and the first x seconds of the next one gets mixed together, and the effect is really pleasing and awesome when you're listening (I personally love it). Anyone willing to add this feature as a plugin to YTM? I don't think I have the knowledge to do this myself, but I might be able to achieve this if I get help with ideas and tips. I'll really appreciate this feature if added 🙃

Zo-Bro-23:
Awesome 🚀

Issue #909, opened 2022-12-06
[Feature Request]: Provide option to play song with search query


Zo-Bro-23:
It would be great if an option could be added to `./providers/` such that the best result for a particular search query will be played. That is, perform a search for that query and play the first option. This can be achieved with keyboard shortcuts, but that will be messy, and that needs the app to be in the foreground. This feature can be used for many useful plugins like voice control, remote control, API control, etc.

JellyBrick:
Related: https://github.com/th-ch/youtube-music/issues/2506#issuecomment-2561520071

Issue #831, opened 2022-09-15
[Feature Request] Like all songs in an Album


Zo-Bro-23:
I think the title says it all. I think it would be great to have a "like all songs" option for an album, instead of the "add to library" option which is currently there. I like to shuffle my "Liked songs" every once in a while, so I think this will be better than to play the albums I like one-by-one.

MiepHD:
`for (like of document.querySelectorAll("#button-shape-like")) {
    if (like.getAttribute("aria-pressed") === "false") {
        like.children[0].click();
    }
}
`
Add a button that runs this

Zo-Bro-23:
> ```
> for (like of document.querySelectorAll("#button-shape-like")) { //Gets all buttons
>     if (like.getAttribute("aria-pressed") === "false") { //Checks if it's pressed
>         like.children[0].click(); //Clicks the button
>     }
> }
> ```
> 
> Add a button that runs this

Thanks. I'll try it out when I have some free time and get back to you!

PR #946, opened 2023-01-07
[Notifications] [Windows] Native interactive notifications

Araxeus:
This is a pretty big PR whose main goal is to replace SnoreToast/node-notifier with native toast notifications using XML
— along the way there were some other changes detailed below

Feel free to ask questions about code or its purpose in this PR.

### Changes related only to this plugin:
* Made config local to the plugin, which means it is synced between the menu and the other files

* Different presets and option to hide text on buttons (screenshots below)

* Adaptive text size in the `Banner Centered` toasts + Time widget in the one custom preset

* Option to open notification on tray click, which can act as a very cool quick music control from everywhere
  (This feature is a personal favorite of mine 🤩)

### Other Changes:

-----
* Added a Protocol Handler (can handle all `song-controls` actions)
  ```
  youtubemusic://pause
  ```
  Should maybe add to README?
  could capitalize on this for other plugins/features
  For example, users could execute the URI with custom remote control apps

----

* Made Single Instance Lock enabled on app launch, can disable it on a specific session via the menu
  Multiple Instances break the Protocol Handler, (and already broke a few other features), and also resources cost is unreasonable

----

* Moved the icons from taskbar-mediacontrol to a global folder (these toasts use the same icons)
  Should probably recolor them to white at some point

----

* The special listeners in `song-info-front` are now activated on demand instead of depending on options
  (Easier to use it from plugins now, don't have to modify song-info-front logic every time, don't have to think about if other plugins requested it since it's a singleton)

----

* SongInfo callback got an extra argument which is the reason for the callback (`video-src-changed` or `playPaused`)
  Not actually used anywhere for now

### Won't Do
* add a way to preview notifications – it's actually pretty easy to change the style in the menu, and play a song to see
  (this plugin doesn't require a restart to apply those options)
* Make live updates to current time in custom notification – can't do without https://github.com/electron/electron/issues/36825)

----

## Screenshots
<details>
  <summary>Click me</summary>
 
 ### `xml_logo_ascii` ✅
This is how it looked on v1.19.0 (using SnoreToast), I see no reason not to include
image | image
|:-:|:-:|

### `xml_logo_icons_notext` ❔
The gap is unavoidable since there HAS to be a text field inside a button, looks better when [not zoomed in on the toast ](https://user-images.githubusercontent.com/78568641/211195012-54a55fb2-70be-4ed5-9d00-0be11ffbf9c7.png)
image | image
|:-:|:-:|

### `xml_logo_icons` ✅
image | image
|:-:|:-:|
### `xml_hero` ✅
Looks a bit weird because the hero appears on top of the notification (above the close button)
image | image
|:-:|:-:|

### `xml_banner_bottom` ✅
similar to xml_hero but less weird, I see no reason not to include
image | image
|:-:|:-:|

### `xml_banner_top_custom` ❓
 **this one is actually interesting**, notice the special info at the right (Album name, Release year)
Cool concept but the top gap is pretty big…
image | image
|:-:|:-:|

### `xml_banner_centered_bottom` ❓
Both of the `centered` toasts make use of special text properties (like font, position)
the text could be made a bit smaller, but the gap will remain
image | image
|:-:|:-:|

### `xml_banner_centered_top` ❔
I like the idea of a bigger font and centered position, but not sure if it's worth the topgap tradeoff
image | image
|:-:|:-:|
 
</details>

#### dev stuff

[[Notifications Visualizer](https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/notifications-visualizer)] --- [[download appx](http://tlu.dl.delivery.mp.microsoft.com/filestreamingservice/files/b49ee55f-a84f-4f29-9b83-fbac1241d5df?P1=1673123404&P2=404&P3=2&P4=bJ7snAx%2fX%2f%2b5EqZ9rNHuIcSHchogYJml%2bQzAJ5K%2fdDHMgLFOiCaxW48kYJspkv50SXuIrZmxFUzmcYL5kd4f7Q%3d%3d)] --- [[CUSTOM PRESETS](https://www90.zippyshare.com/v/buT4qt5H/file.html)]

more on toast XML:
[schema](https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/schema-root)
[general tutorial](https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=xml)
[toasttemplatetype](https://learn.microsoft.com/en-us/uwp/api/windows.ui.notifications.toasttemplatetype)


Liknox (approved):

Proton7123:
I love this app! EPIC SUGGESTION: ability to create keyboard hotkey shortcuts (on desktop releases) example: I love listening to my own music while gaming online, sometimes I want to thumbs up a track, but am unable to switch apps, so the ability to designate whatever key, to whatever commonly used function the user wishes to define in the settings.
I'm virtually certain this is in the wrong place, but am unfamiliar with github and busy with something else... so here you go. Thanks!

Zo-Bro-23:
> I'm virtually certain this is in the wrong place

Moved it to #947!

Araxeus:
@th-ch @Zo-Bro-23 @MiepHD or anyone who might see this

What do you think of the presets?
Should they all be added as options or just some of them?

`xml_banner_top_custom` + `xml_banner_centered_bottom` + `xml_banner_centered_top` have a gap on the top part of the toast since they use an invisible character to make the top text invisible and render other text instead (not sure if those are event worth it)

MiepHD:
I'd let them choose between `xml_hero`, `xml_logo_icons`.
`xml_logo_icons_notext` would be an option for me if there were no gap under the buttons.
`xml_logo_ascii` could be used instead if the gap can't get removed. There I'd recommend to add some space between the | to get a square icon.

Araxeus:
I've updated the post with my thoughts on each of them, but overall is there a reason not to include any of them?

more customization might be better if there isn't a tradeoff

I think I might just have a submenu called `style` where you can choose any of these styles using radio buttons.
Just have to think of good informative names for each of them

Zo-Bro-23:
This is _**AWESOME**_! Personally, don't really like the look of the ASCII versions, but it doesn't hurt to keep them in. The others are really visually pleasing, and of course, the more options the better (particularly love the hero and big banner ones)! How are you planning to implement these options though? Only if presented with a set of images like these will the user know which one to pick. Simply showing a list of big banner, hero, etc will be hard for the user to know what each one looks like. If you figure out a way to list all the possible options (maybe have a button that brings up a pop-up showing demos for how all these options look like), then you've got yourself a perfect feature! Looking forward to seeing this implemented 🙃

Araxeus:
> don't really like the look of the ASCII versions

Yeah if I made the interactive-notifications plugin from scratch today I probably wouldn't include something like that
But this is how it used to look so #LegacyOption

> maybe have a button that brings up a pop-up showing demos for how all these options look like

very good idea 🌟 -  will require quite a bit of additional code but might be worth it

Thanks for the encouragement :D

Zo-Bro-23:
> Thanks for the encouragement :D

Psst... (got some feature ideas in #909, #831, and #945 - would be great if you could try implementing them!). Just giving me some tips on how to implement them would also be fine, and I can try working on them myself. Keep up the great work 😄

Araxeus:
I've decided for now I'll do it like so:
```js
const pluginOptions = {
	toastStyle: 1,  // 1-7
	hideButtonText: false
}
```
`hideButtonText` effectively doubles the number of styles from 7 to 13 😂

> Psst... (got some feature ideas in #909, #831, and #945 - would be great if you could try implementing them!). Just giving me some tips on how to implement them would also be fine, and I can try working on them myself. Keep up the great work 😄

I'll look into it when I have some downtime. meanwhile you can add me on discord <kbd>Araxeus#0819</kbd>

Zo-Bro-23:
> ```js
> const pluginOptions = {
> 	toastStyle: 1,  // 1-7
> 	hideButtonText: false
> }
> 
> module.exports.ToastStyles = {
> 	logo: 1, // logo_icons 
> 	hero: 2,
> 	banner_top_custom: 3,
> 	banner_bottom: 4,
> 	banner_centered_bottom: 5,
> 	banner_centered_top: 6,
> 	legacy: 7 //  logo_ascii
> }
> ```

So these are the options that will show up on the menu? If I understand this correctly, the user can manually select options like `show_text`, `show_logo`, etc instead of choosing from presets. That's great, then! (No need for having demo images in that case, although it never hurts to do that - I'm personally confused by some app options sometimes, so I have to try them out to see what they do)

> I'll look into it when I have some downtime.

Cool, that's great! Thanks so much 😁

Araxeus:
> So these are the options that will show up on the menu?

image

> If I understand this correctly, the user can manually select options like show_text, show_logo, etc instead of choosing from presets

No, the only customizable option right now is `Hide Button Text`, other than that its one of presets

>  show_logo

`show_logo` would just means it;s preset 1

Zo-Bro-23:
Oh. Having the image previews might be good in that case. Also try if you can have these as options, which will allow for more customization (mix-and-match of different settings).

Araxeus:
> Oh. Having the image previews might be good in that case. Also try if you can have these as options, which will allow for more customization (mix-and-match of different settings).

I understand your meaning, but think about it - what can you actually mix-and-match here? Each of the presets is unique.
that's why they are preset and not options

the only thing that I can  think of that could be an option is `Banner Position`: `top/bot` but:
* it's niche since it would affect only 2/5 presets
* the way the toast XML works, those are built completely differently and require a different template

 compared to `Hide Button Text` which just changes a string to empty - there really isn't any point

Zo-Bro-23:
> I understand your meaning, but think about it - what can you actually mix-and-match here? Each of the presets is unique.
that's why they are preset and not options

You're right - my bad! The show button text is probably the only one that can be an option, and you've already incorporated that. I thought about having choices for the other properties, but that gets confusing, because you can't have BOTH hero and banner, and you can't have BOTH text above and text below. So I've got nothing more to add - presets are the best way to do this!

Araxeus:
This PR is actually mostly complete now with [`7052217` (#946)](https://github.com/th-ch/youtube-music/pull/946/commits/70522173b73fe537df9822c768e9007d23883d56)

If you are on windows, you can clone the branch and try it out !

> maybe have a button that brings up a pop-up showing demos for how all these options look like
I'm not even sure this is needed anymore, you can just select any of them and it will automatically update without needing to restart the app

which makes it easy to test each of them (tho I agree a dedicated panel for demo would be cool, as it stands I'm not sure there is that much to gain)

Zo-Bro-23:
> If you are on windows, you can clone the branch and try it out! :metal: :rocket:

My laptop's gone in for repair, so I'm on Ubuntu right now 🥲. Looking forward to testing it out when I get my laptop back!

menu.js
```diff
@@ -131,16 +131,14 @@ const mainMenuTemplate = (win) => {
 					],
 				},
 				{
-					label: "Single instance lock",
+					label: "Release single instance lock",
```
Araxeus, line 134: Single instance lock enabled by default
(Required for protocol handler)

preload.js
```diff
@@ -91,6 +95,8 @@ function listenForApiLoad() {
 
 function onApiLoaded() {
 	document.dispatchEvent(new CustomEvent('apiLoaded', { detail: api }));
+	//setImmediate()
```
Araxeus, line 98: Delete this

providers/protocol-handler.js
```diff
@@ -0,0 +1,44 @@
...
+			[path.resolve(process.argv[1])]
+		);
+	} else {
+		app.setAsDefaultProtocolClient(APP_PROTOCOL)
+	}
+	  
+	const songControls = getSongControls(win);
+
+	protocolHandler = (cmd) => {
+		if (Object.keys(songControls).includes(cmd)) {
```
Araxeus, line 23: Is this a duplicate check?

Look into it

providers/song-controls.js
```diff
@@ -50,4 +49,9 @@ module.exports = (win) => {
 		search: () => pressKey(win, "/"),
 		showShortcuts: () => pressKey(win, "/", ["shift"]),
 	};
+	return {
+		...commands,
+		play: commands.playPause,
+		pause: commands.playPause
```
Araxeus, line 55: Add dedicated commands using song-controls-front and ipc?

providers/song-info-front.js
```diff
@@ -17,14 +14,63 @@ ipcRenderer.on("update-song-info", async (_, extractedSongInfo) => {
...
 const srcChangedEvent = new CustomEvent('srcChanged');
 
+const singleton = (fn) => {
+	let called = false;
+	return (...args) => {
+		if (called) return;
+		called = true;
+		return fn(...args);
+	}
+}
```
Araxeus, line 24: Can extract this function so that it is available app wide

providers/song-info.js
```diff
@@ -61,7 +61,8 @@ const handleData = async (responseText, win) => {
 		songInfo.album = data?.videoDetails?.album; // Will be undefined if video exist
 
 		const oldUrl = songInfo.imageSrc;
-		songInfo.imageSrc = videoDetails.thumbnail?.thumbnails?.pop()?.url.split("?")[0];
```
Araxeus, line 64: Shouldn't use `pop` if not intending to change the original array
Wanted to use `.at(-1) but it requires node 16.6+ aka electron 23+

providers/song-info.js
```diff
@@ -95,15 +96,15 @@ const registerProvider = (win) => {
 		await handleData(responseText, win);
 		handlingData = false;
 		callbacks.forEach((c) => {
-			c(songInfo);
+			c(songInfo, "video-src-changed");
```
Araxeus, line 99: Unused right now, but now every plugin has the option to know whats the reason for the songInfo callback

providers/song-controls.js
Araxeus: a bit much for now... maybe another time if someone really wants a command that only play or pause

plugins/notifications/utils.js
```diff
@@ -13,38 +25,37 @@ module.exports.icons = {
...
-		? centerNativeImage(songInfo.image)
-		: icon
+module.exports.notificationImage = (songInfo) => {
+	if (!songInfo.image) return icon;
+	if (!config.get("interactive")) return nativeImageToLogo(songInfo.image);
+
+	switch(config.get("toastStyle")) {
+		case module.exports.ToastStyles.logo:
+		case module.exports.ToastStyles.legacy:
+			return this.saveImage(nativeImageToLogo(songInfo.image), tempIcon);
```
Araxeus: This is inconsistent stylewise, I'm not sure which one should be used here

What do you think @th-ch ?😅
Should we use `module.exports.` or `this.`?

Araxeus:
@th-ch any thoughts?

menu.js
th-ch, line 134: nit: it seems the button will request the lock it it does not have it, name can be misleading
```suggestion
					label: "Single instance lock",
```

plugins/notifications/config.js
```diff
@@ -0,0 +1,23 @@
+const { setOptions, setMenuOptions } = require("../../config/plugins");
+const defaultConfig = require("../../config/defaults");
+
+let config = defaultConfig.plugins["notifications"];
+
+module.exports.init = (options) => {
+	config = { ...config, ...options };
+};
+
+module.exports.setAndMaybeRestart = (option, value) => {
```
th-ch, line 10: nit: should we make the name more explicit? Sthg like
```suggestion
module.exports.setInMenu = (option, value) => {
```

plugins/notifications/utils.js
th-ch: No strong preference, `this` is fine :)

providers/song-info.js
Araxeus: Well it has been released since then
https://www.electronjs.org/blog/electron-23-0

but it drops support for Windows 7/8/8.1, so I'm unsure when should we bump that

Liknox (approved):
Looks great, merge is expected.

plugins/notifications/config.js
Araxeus: I think `setAndMaybeRestart` is more explicit since the function name indicates that the difference is the restart option
`setInMenu` doesn't tell me much tbh

PR #953, opened 2023-01-09
[New plugin] Music visualizers

th-ch:
This PR implements a new plugin that replaces the video by a music visualizer:
- Butterchurn:
butterchurn
- Vudio:
vudio
- Wave:
wave

New visualizers can be implemented by creating a new class with the same interface in the `visualizers` folder. Possible future improvements:
- Wave resizing seems to have issues, particularly in fullscreen
- have a toggle between the video and the visualizer (currently needs to disable the plugin)


Zo-Bro-23:
> I just stumbled upon a way to make youtube actually play only music ,(not even download the video)

That'll be great for saving resources and increasing performance.

> but maybe all of the above can be consolidated into one plugin?

Yes, that might give a better user experience overall. Video toggle plugin with option to force no video, have visualizers instead, or have the original video available with the toggle.

Araxeus:
Pretty sure you want migrations for this one

probably should use `setDefaultPluginOptions`
https://github.com/th-ch/youtube-music/blob/35bd62cc0d8856ed161bbd865255c88e15a08855/config/store.js#L13

th-ch:
All these ideas above make sense (audio only, consolidating video toggle and visualizer, disable rendering when not focusing, etc) - I will merge the PR as is as a start, we can have follow up PRs for the rest!

PR #956, opened 2023-01-09
Fixed video-toggle aligning running before #main-panel exists

MiepHD:
> This can possibly break the plugin (and it did for me, multiple time) `Cannot read properties of undefined.` `mainpanel.style.setProperty("--align", "0px");`
> 
> this whole code block needs to be inside an `apiLoaded` closure, we can't guarantee that `$('#main-panel')` exists when this code is called
> 
> In other words, instead of being inside the default export function, it should be inside the `setup` function which is called only when the page is loaded
> 
> @MiepHD @th-ch plz fix

### DIFF
--- a/plugins/video-toggle/button-switcher.css
+++ b/plugins/video-toggle/button-switcher.css
@@ -22,7 +22,6 @@
     color: #fff;
     padding-right: 120px;
     position: absolute;
-    left: var(--align);
 }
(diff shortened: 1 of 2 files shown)


Araxeus:
thanks

th-ch (approved):

PR #973, opened 2023-01-14
Add CI job for dependency review

th-ch:
Fixes https://github.com/th-ch/youtube-music/issues/965


Araxeus:
Why not also enable CodeQL? or is it not a workflow file anymore?

th-ch:
> Why not also enable CodeQL? or is it not a workflow file anymore?

It seems to be a UI setting, I enabled it, unless I missed something, it [runs as well](https://github.com/th-ch/youtube-music/actions/runs/3919100917/jobs/6699896079)!

Araxeus:
cool
image

PR #974, opened 2023-01-14
Update electron and various dependencies

th-ch:
This PR update electron to v22 and other dependencies - it also forces node >= 16 (already the case in CI).

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
 		"release:win": "yarn run clean && electron-builder --win -p always"
 	},
 	"engines": {
-		"node": ">=14.0.0",
+		"node": ">=16.0.0",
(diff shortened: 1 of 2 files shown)


Araxeus:
To be fair, I noticed the node version in package.json doesn't matter at all. It is completely enforced by the electron version

(maybe it does matter for scripts or something)

Araxeus:
Breaking changes:

Removed: WebContents new-window event
https://www.electronjs.org/docs/latest/breaking-changes#removed-webcontents-new-window-event

https://github.com/th-ch/youtube-music/blob/3a822f611af13f31d74f5c4636cc73534548aa56/index.js#L321-L331

@th-ch

Issue #975, opened 2023-01-14
Menu option to copy current URL


kotx:
The share function doesn't provide the whole URL, so you can't share what playlist you're on, etc.

Araxeus:
Should that actually be a button in the already cluttered menu?

maybe there's some better discreet place for that

kotx:
> maybe there's some better discreet place for that

I don't think there would be a better place for it since it needs to be accessible everywhere in the app. It being in the Navigation menu (like in the above PR) seems fine.

PR #977, opened 2023-01-14
Add menu entry to copy current URL

th-ch:
Fixes https://github.com/th-ch/youtube-music/issues/975


Araxeus:
I don't think there is a need of an external dependency
https://www.electronjs.org/docs/latest/api/clipboard

Issue #940, opened 2023-01-06
Windows / Bypass Age Restriction


lvlts:
The latest build (1.19) does not seem to bundle the bypass restrictions plugin:

```
Uncaught Error: Cannot find module 'simple-youtube-age-restriction-bypass/Simple-YouTube-Age-Restriction-Bypass.user.js'
Require stack:
- C:\Users\user\AppData\Local\Programs\youtube-music\resources\app.asar\plugins\bypass-age-restrictions\front.js
- C:\Users\user\AppData\Local\Programs\youtube-music\resources\app.asar\preload.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:940:15)
    at i._resolveFilename (node:electron/js2c/renderer_init:33:1095)
    at Module._load (node:internal/modules/cjs/loader:785:27)
    at c._load (node:electron/js2c/asar_bundle:5:13343)
    at i._load (node:electron/js2c/renderer_init:33:356)
    at Module.require (node:internal/modules/cjs/loader:1012:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at module.exports (C:\Users\user\AppData\Local\Programs\youtube-music\resources\app.asar\plugins\bypass-age-restrictions\front.js:3:2)
    at C:\Users\user\AppData\Local\Programs\youtube-music\resources\app.asar\preload.js:54:4
    at C:\Users\userAppData\Local\Programs\youtube-music\resources\app.asar\plugins\utils.js:48:3
```

Araxeus:
not sure what is going on here, it is included, and also from my testing it works in both
https://github.com/th-ch/youtube-music/releases/download/v1.19.0/YouTube-Music-Setup-1.19.0.exe
and in dev mode

https://github.com/th-ch/youtube-music/blob/v1.19.0/package.json#L99

https://github.com/th-ch/youtube-music/blob/v1.19.0/plugins/bypass-age-restrictions/front.js#L1-L4

lvlts:
@Araxeus 

I uninstalled. Pulled the same link you provided and reinstalled. Same issue in the developer tools.


MD5 of the installer:
```
008e08b05e92a6604dff00babbf39e9c  /home/user/Downloads/YouTube-Music-Setup-1.19.0.exe
```


[...]

PR #984, opened 2023-01-16
Fix bypass-age-restriction lib import

th-ch:
Fixes https://github.com/th-ch/youtube-music/issues/940 by generating a package.json for the lib to ensure electron-build picks it up (+ fix the case in the import)


Araxeus:
When is `generate:package` called? I'm not quite following what's going on here

Araxeus:
very handy that you still had `generate:package` from a while back when we had YoutubeNonStop 😁

PR #1005, opened 2023-01-28
[Shortcuts] MPRIS fixes, Repeat Language bug fix

Skydeke:
The MPRIS-Interface (Linux) has the range 0.0 - 1.0 for volume. It is also better to use the constants provided by the library. I changed the setupRepeatChangeListener() function because it returned the mode of the repeat of the music in the language that Youtube-Music has active, but because those strings are hard coded in mpris.js, it would break MPRIS if YT-music was set to another language. Thus I used the cookie set by YT-music to determine the repeat mode and mapped that to the english valuse since those were used in code.


plugins/shortcuts/mpris.js
```diff
@@ -94,40 +84,48 @@ function registerMPRIS(win) {
 		player.on('seek', seekBy);
 		player.on('position', seekTo);
 
+		player.on('shuffle', (enableShuffle) => {
+			shuffle();
+		});
+
 		ipcMain.on('volumeChanged', (_, value) => {
-			player.volume = value;
+			if (config.plugins.isEnabled('precise-volume')) {
```
Araxeus: What is this `if` about? `precise-volume` doesn't change the way the volume is transmitted?
`api.getVolume()` always returns a number between 0-100, regardless of if `in-app-menu` is enabled
https://github.com/th-ch/youtube-music/blob/0f09f8a8ed8d6ffe5e95aaf47fce7368800bbdf4/providers/song-info-front.js#L79-L85

Still a good fix, should just probably remove the `if`

providers/song-info-front.js
```diff
@@ -55,25 +55,62 @@ module.exports = () => {
...
+								ipcRenderer.send('repeatChanged', "Repeat off");
+							else if (splitpair[1] === "ONE") //MPRIS Playlist and Track Codes are switched to look the same as yt-music icons
+								ipcRenderer.send('repeatChanged', "Repeat one");
+							else if (splitpair[1] === "ALL")
+								ipcRenderer.send('repeatChanged', "Repeat all");
+						}
+					}
+				});
+			}
+		});
```
Araxeus: Instead of all this cookie stuff, try:
```js
const mp = { NONE: "Repeat off", ONE: "Repeat one", ALL: "Repeat all" } 
ipcRenderer.send('repeatChanged', mp[
	mutations[0].target.__dataHost.getState().queue.repeatMode
])
```
> `mutations[0].target.__dataHost` = `$('ytmusic-player-bar')`

and actually since this format is specific to mpris - this mapping should probably not be done globally
just send the event as `ONE | NONE | ALL` and do the `mp` in mpris.js

Araxeus (changes requested):

plugins/shortcuts/mpris.js
Skydeke: Yes, I must have missed that from when I was debuging. I will remove the if.

providers/song-info-front.js
Skydeke: Your short version works, I'd still choose to send "Repeat off" | "Repeat one" | "Repeat all" since that was what was sent before I touched the function and I dont know if any other plugins also listen for 'repeatChanged'. I'd hate to break those.

Araxeus: >  I dont know if any other plugins also listen for 'repeatChanged'

For now there aren't any, that's why I proposed moving code specific to MPRIS to its own file

>  since that was what was sent before I touched the function

Yeah well doesn't mean that it's right 😅

Anyway you can leave as is, doesn't really matter

Araxeus:
LGTM (except the missing padding spaces inside curly braces, but that's just a style diff 😋)

Araxeus (approved):

Skydeke:
After some more testing I did find a bug: Sometimes when the volume is changed cpu usage will spike to 100% for a few seconds. Don't merge until I have fixed that.

Skydeke:
The reason the cpu sometimes spiked was that sometimes the mpris volume slider and the internal one got desynced and then recursively called the update function of the other. Its fixed now.

th-ch (approved):

Issue #634, opened 2022-03-05
[Feature Request] Fade between Songs (Crossfade)


xdaWolf:
Fade between Songs (like Crossfade in Spotify for example)

ForlornWolf:
Yes

TemDevz:
bump

xdaWolf:
Dude, that's awesome!! Just wanted to let you know that I absolutely love this software! Thanks for always adding more features. Keep it up!! Very appreciated!! :)

PR #1012, opened 2023-02-01
[new plugin] Add first version for crossfade plugin

th-ch:
This PR adds a plugin to crossfade between songs (fixes https://github.com/th-ch/youtube-music/issues/945 and https://github.com/th-ch/youtube-music/issues/634) - it's a first version that plays the sound in the background to be able to crossfade. A simpler version could be shipped that just fades in/out the music without crossfading.


plugins/crossfade/front.js
```diff
@@ -0,0 +1,152 @@
...
+	const video = document.querySelector("video");
+	const videoFader = new VolumeFader(video, {
+		fadeScaling: crossfadeOptions.fadeScaling,
+		fadeDuration: crossfadeOptions.fadeInDuration,
+	});
+
+	await transitionAudio.play();
+	await transitionAudio.seek(video.currentTime);
+
+	video.onseeking = () => {
```
Araxeus, line 78: couldn't those `video.onseeking` be `video.addEventListener('seek' => {})`?

PR #1022, opened 2023-02-08
Migrate to yarn v3

th-ch:
This PR migrates the project to yarn v3 to reduce [errors like `ESOCKETTIMEDOUT `](https://github.com/th-ch/youtube-music/actions/runs/4109181466/jobs/7090721146)


Araxeus:
So this requires having Yarn v3 installed globally right?

Wouldn't that clash with other locally installed projects that use Yarn v1?
I for one have quite a bit of those, not sure how to deal with this potential situation

Araxeus:
Another unrelated question:
in build.yml on each build step the `secrets.GH_TOKEN` is passed, why? isn't this only needed in the release step?
https://github.com/th-ch/youtube-music/blob/70c55ca587f75aa9024c82a3884fe21e8a817a20/.github/workflows/build.yml#L48-L53

this cause the build step to fail on forks since the `GH_TOKEN` isn't defined
https://github.com/Araxeus/youtube-music/actions/runs/4136517862/jobs/7150485857

th-ch:
> So this requires having Yarn v3 installed globally right?

Not necessarily! I typically have yarn `1.22.19` globally, but running `yarn --version` in the project will return `3.4.1` (it uses `.yarn/releases/yarn-3.4.1.cjs`)

th-ch:
> on each build step the secrets.GH_TOKEN is passed, why? isn't this only needed in the release step?

The actually, this step both builds and releases a draft version, and the release stage will make that draft a public release - so it requires the GH token, though the case of forks is a good point, I'll see what can be done!

Issue #571, opened 2022-01-24
Last.fm scrobbler plugin doesn't create the app properly


DeRRudi77:
Hi all,

When trying to connect youtube-music to my last.fm account. I am able to go through the entire flow, but in the end, it doesn't create the app in my settings and scrobbling doesn't work. Any ideas on how to fix this? See screenshots for the flow. It all looks successful.

![image](https://user-images.githubusercontent.com/1484504/150815226-51829907-5853-4181-8cdf-395b84f46096.png)

I had this connection setup before, but removed it. Now re-adding seems to be a problem.

Rutger

Araxeus:
could be the cause of #546 and #540
@semvis123 maybe a problem with the registered api_key/secret?
https://github.com/th-ch/youtube-music/blob/61eb23614a27a50e6901306f039cc7c0aed446db/config/defaults.js#L44-L45

JellyBrick:
Fixed due to #1032

Impasse52:
Is this PR merged in the latest release? Sadly this still doesn't work for me, and it seems like I can't build from source either. I get `npm ERR! Missing script: "build:linux"`.

JellyBrick:
> Is this PR merged in the latest release? Sadly this still doesn't work for me, and it seems like I can't build from source either. I get `npm ERR! Missing script: "build:linux"`.

Use `npm run dist:linux`

Impasse52:
Yep, this did the trick, thank you!

PR #1032, opened 2023-02-17
try to fix songInfo time&album

### LINKED ISSUE #540: Scrobbling Issues
Using version 1.15.0. When playing an album, only the first track will be properly scrobbled to Last.fm. Subsequent tracks will appear in Last.fm as "scrobbling now" but then not commit.

image
image

### LINKED ISSUE #744: Album cover label on Discord off by one
With the Discord rich presence, the name of the album when hovering the cover displays the previous song's album.
ex

### LINKED ISSUE #806: Last.fm scrobbler is shoddy
when using the latest release build only the first song in the queue is scrobbled, and the album title isn't taken into account at all leading to bootleg kanye anime scrobbles. a fix to this was committed back in like april but it's not in the release build for some reason. building it manually fixes this but i don't see why the commits weren't applied to the release build?

image

and on that note, the issue of only properly scrobbling the first track isn't fixed. clicking on the album for this portishead song leads me to the previous song's respective album title.

image
image

### LINKED ISSUE #902: Last.FM scrobbling issue. (and Discord RPC 00:00)
I noticed that the last.fm scrobbling doesn't seem to work at all, except for the very first song of the day.

I dug into the issue, and I've realized that it's due to the fact that the `elapsedSeconds` field that gets taken from `video.currentTime` on the video element returns a huge number instead of 0 when switching to a new track. When that happens, [this](https://github.com/th-ch/youtube-music/blob/master/plugins/last-fm/back.js#L151-L152) calculation fails, as the plugin thinks we are past the halfway part of the song, so it never starts the scrobbler timeout. 

Fixing this should also fix the issue where discord lists 00:00 as the current progress of the song.

I don't have any proposals at the moment how this could be fixed, other than just overriding the field with `0` whenever `video-src-changed` event is fired, which I will be using as a workaround for now. 

It seems like the `video.currentTime` keeps going up as a sum of all song lengths played instead of resetting to 0 when a new song starts playing. I think it's not reliable to depend on this either way, as it seems they've implemented their own stream that the player just listens to.

### LINKED ISSUE #957: Last.FM scrobble doesn't work
Hi!

Thanks for your work on the app, it's great :)

I just noticed last.fm scrobble doesn't seem to work. I get one first scrobble every session, and then nothing. I'd like to know if this is a known issue and if there is any fix or workaround I could try.

I'm using the .deb version on Linux Mint.

### LINKED ISSUE #967: Make last.fm support better
Hey I am using your client for a while now and like it a lot but the last.fm plugin could be improved.

* Add a filter to the songs so that for example (Official Video) gets filtered, or the channel name is not the artist name.
https://github.com/web-scrobbler/metadata-filter
* When I listen to a song it shows up twice on last.fm and when the next one plays it just shows up once

### LINKED ISSUE #1006: LastFM album scrobbling not working correctly.
I use pano scrobbler on my phone and it tells me what from what album the song is.

The problem is that most of the times the album just doesn't appear or uses the album from the last song played.

I've tried looking into the settings but just couldn't find anything.

### LINKED ISSUE #1035: Uploaded songs provide incorrect album info to the songInfo object.
Was trying to get Tuna working with the app, but ran into the following issue: I have a large number of songs that I've _uploaded_ to YouTube Music (well, most of them are from the Google Play Music days, but no matter) and when I try to grab the name of the album for display on stream I instead get the artist name corresponding to the first _played_ track in the playlist.

This does not appear to be a problem with Tuna, and indeed when I had a look at the `songInfo` object in the YTM app console, it turned out the album name was wrong there. So it seems that something is going wrong with pushing album names of uploaded songs to `songInfo`.

Araxeus:
`data.videoDetails.elapsedSeconds = 0` - sendSongInfo is called when the video start, no reason for it be anything but

`setTimeout(sendSongInfo)` - Try to delay grabbing the album until it has updated

also [d122ffe](https://github.com/th-ch/youtube-music/pull/1032/commits/d122ffe72eb0db08a06b4440dde5bb23087ffc03) fix #1035

will probably fix #902 fix #957 fix #967 fix #744 fix #806 fix #1006 fix #540 and more


Liknox (approved):

ionyx0:
I have built and run this locally for about an hour now and have noticed that it seems to fix the Last.FM issues reported.  Previously, no albums were scrobbled but they are now being picked up and sent to the `songInfo` object.

Seems safe to merge from my perspective.

Araxeus:
Thank you for testing !

WillPresley:
Hey @Araxeus, is there any reason that the `data.videoDetails.elapsedSeconds = 0` bit was not included in the final push to master? See https://github.com/th-ch/youtube-music/blob/master/providers/song-info-front.js#L103
Or was it somehow overwritten before the final branch push here?

I noticed after a fresh git pull and self-build today that this exact same issue was happening again, and when I checked out the song-info-front.js file, it was still using the currentTime floor method and not `data.videoDetails.elapsedSeconds = 0`. After changing that and rebuilding, scrobbling works as it should again.

shiiii-bal:
i don't code. how to do that as mentioned in the solution?

shiiii-bal:
i'm having the same issue as #571

shiiii-bal:
> @shiiii-bal Already fixed. Use latest version (v3.6.1).

i am using the latest version though.. i even reinstalled it. I connected my last.fm account and entered the api keys but scrobbling isn't working

shiiii-bal:
I installed it through `YouTube-Music-Web-Setup-3.6.1` btw

shiiii-bal:
> @shiiii-bal Did you restart the application?

i did a few times and it's the same

shiiii-bal:
> @shiiii-bal Did you restart the application?

i reinstalled the app and it's working fine now.

PR #1062, opened 2023-03-10
add rome (formatter & linter)

Araxeus:
* Add [rome](https://rome.tools/) (https://github.com/rome/tools) as both a formatter and linter – https://github.com/th-ch/youtube-music/pull/1062/commits/da57d59191bbd0d809748d92a048a9a727db4228

* Apply format suggestions using `rome format . --write` – https://github.com/th-ch/youtube-music/commit/c37ca542c71e0c22018d864f56a9330e70877655

* Manually fix issues arising from `yarn rome ci .` – https://github.com/th-ch/youtube-music/commit/fbc0fbdbcefe1297946448cd79e37eaa715533e8

* Modify `yarn lint` to use rome instead of xo – https://github.com/th-ch/youtube-music/pull/1062/commits/881a7ce93c08a8479035957bb5aa61660b772ff1 *-2380 lines in yarn.lock*

* Add a `yarn check` to check for syntax/lint errors using rome 

* Add action that runs `yarn check` on pull requests and uses https://github.com/Araxeus/setup-yarn-pnp-action to reduce boilerplate – https://github.com/th-ch/youtube-music/pull/1062/commits/5c6e9265fe88d94b5dddd29588bfca005bf6e6d1

* add rome to vscode recommended workspace extensions – https://github.com/th-ch/youtube-music/pull/1062/commits/eb646319d76bf40d3f31760ff312dfef46d466b3

> TODO: merge upstream, bump rome, fix issues


Araxeus:
@th-ch overall thoughts? rome is in heavy development and in early stages compared to eslint

do you have any preferences for specific [format options](https://docs.rome.tools/configuration/#formatter) / [lint rules](https://docs.rome.tools/lint/rules/) ?

For example I personally prefer the format rules `indent: 4 spaces` + `quoteStyle: single`

```json
  "javascript": {
    "formatter": {
      "quoteStyle": "single"
    }
  },
  "formatter": {
    "enabled": true,
    "indentStyle": "space",
    "indentSize": 4
  },
```
> I didn't add those because the current style doesn't seem to abide to these rules at all

.github/workflows/checks.yml
```diff
@@ -0,0 +1,23 @@
...
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Yarn PnP Setup
```
th-ch: Is it needed? We did not run it previously

.github/workflows/build.yml
```diff
@@ -16,47 +13,8 @@ jobs:
...
-        shell: bash
-        run: |
-          yarn install --immutable --inline-builds
-        env:
-          # CI optimizations. Overrides yarnrc.yml options (or their defaults) in the CI action.
-          YARN_ENABLE_GLOBAL_CACHE: "false" # Use local cache folder to keep downloaded archives
-          YARN_NM_MODE: "hardlinks-local" # Hardlinks-(local|global) reduces io / node_modules size
-          YARN_INSTALL_STATE_PATH: .yarn/ci-cache/install-state.gz # Very small speedup when lock does not change
+      - name: Yarn PnP Setup
+        uses: Araxeus/setup-yarn-pnp-action@v1
```
th-ch, line 19: We don't specify the node version anymore?

package.json
```diff
@@ -92,7 +92,7 @@
 		"build:mac": "yarn run clean && electron-builder --mac dmg:x64",
 		"build:mac:arm64": "yarn run clean && electron-builder --mac dmg:arm64",
 		"build:win": "yarn run clean && electron-builder --win",
-		"lint": "xo",
+		"lint": "rome ci .",
```
th-ch: Should we keep lint for fixing and have a separate command for checking?
```suggestion
		"lint": "rome format",
		"lint:check": "rome ci",
```
and then have CI call `yarn lint:check .`

th-ch (commented):
Good initiative, left some comments/questions but looks good overall!

.github/workflows/build.yml
Araxeus: The node version is locked in Electron apps
[It is bundled with it](https://www.electronjs.org/docs/latest/tutorial/electron-timelines)

Unless I'm missing something?

.github/workflows/checks.yml
Araxeus: it's just a boilerplate action that I use to set up yarn and cache, you can see the action here: https://github.com/Araxeus/setup-yarn-pnp-action/blob/main/action.yml

It isn't really needed, I could revert this change if you want, it just removes some duplicate boilerplate lines (notice it is used twice in this file)

package.json
Araxeus: fixed by https://github.com/th-ch/youtube-music/pull/1062/commits/fa069d11b8faf5758a71f8a3d94216726c0fbe11

Araxeus:
@th-ch I think you might have missed my [comment above,](https://github.com/th-ch/youtube-music/pull/1062#issuecomment-1464451804)

th-ch:
Sorry, missed the comment! 4 spaces and single quotes are fine by me, we can definitely have them in rome.json to set a standard 💯

.github/workflows/checks.yml
th-ch: Sorry, I'm fine with the setup action, I just meant: do we need to do the setup for the dependency review? We only did the checkout previously

Araxeus: Ooh wait, my bad… you're right!

Araxeus:
Changing the formatting to 4 spaces and single quotes made the total lines changed jump from 2k to 6k 🦄

rome.json
```diff
@@ -0,0 +1,26 @@
...
+  "$schema": "./node_modules/rome/configuration_schema.json",
+  "javascript": {
+    "formatter": {
+      "quoteStyle": "single"
+    }
+  },
+  "formatter": {
+    "enabled": true,
+    "indentStyle": "space",
+    "indentSize": 2
```
th-ch, line 16: So I guess we could switch to 4, but in a later PR to minimize the diff here?

package.json
```diff
@@ -92,7 +92,8 @@
 		"build:mac": "yarn run clean && electron-builder --mac dmg:x64 -p never",
 		"build:mac:arm64": "yarn run clean && electron-builder --mac dmg:arm64 -p never",
 		"build:win": "yarn run clean && electron-builder --win -p never",
-		"lint": "xo",
+		"check": "rome ci .",
+		"lint": "rome format . --write",
```
th-ch, line 96: Formatting the whole source folder can create issues on Mac with the dist folder - how about making it run only on specific files?
```suggestion
		"lint": "rome format --write",
```
(then using `yarn lint FILE`)

.vscode/extensions.json
th-ch, line 1: In addition to the recommended extension, how about configuring rome for the workspace by adding `.vscode/settings.json`:
```
{
  "editor.defaultFormatter": "rome.rome",
  "prettier.enable": false,
  "editor.formatOnSave": true,
  "[javascript]": {
    "editor.defaultFormatter": "rome.rome"
  }
}
```

th-ch (approved):

rome.json
Araxeus: Honestly, the diff doesn't matter now, it's just a matter of deciding on 2 or 4

I just second guessed myself about 4 spaces - it looks horrible if there are callbacks in callbacks and stuff

we already mostly use 2 spaces, so maybe we just continue with that?

th-ch: Looks good to me with 2 spaces then! ✅

package.json
Araxeus: I can just add the dist folder to rome.json .files.ignore

is that good enough?

.vscode/extensions.json
Araxeus: thank you this is really great!

package.json
th-ch: I tried but it does not seem to work 😅 
```
Versions/Current/Resources internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Infinite symlink expansion
  
  ✖ Rome encountered a file system entry that leads to an infinite symbolic link expansion, causing an infinite cycle: Versions/Current/Resources
  

Versions/Current/Resources internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Infinite symlink expansion
  
  ✖ Rome encountered a file system entry that leads to an infinite symbolic link expansion, causing an infinite cycle: Versions/Current/Resources
  

A internalError/io ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ No such file or directory (os error 2)
  

Versions/Current/Resources internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Infinite symlink expansion
  
  ✖ Rome encountered a file system entry that leads to an infinite symbolic link expansion, causing an infinite cycle: Versions/Current/Resources
  

Versions/Current/Libraries internalError/io ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ No such file or directory (os error 2)
  

Versions/Current/Resources internalError/io ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ No such file or directory (os error 2)
  

Versions/Current/Helpers internalError/io ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ No such file or directory (os error 2)
  

A internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Infinite symlink expansion
  
  ✖ Rome encountered a file system entry that leads to an infinite symbolic link expansion, causing an infinite cycle: A
  

A internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Infinite symlink expansion
  
  ✖ Rome encountered a file system entry that leads to an infinite symbolic link expansion, causing an infinite cycle: A
  

A internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Infinite symlink expansion
  
  ✖ Rome encountered a file system entry that leads to an infinite symbolic link expansion, causing an infinite cycle: A
  

Formatted 0 file(s) in 2ms
Error: no files were processed in the specified paths.
```

I think it fails when listing the files, even if the whole folder is ignored

.vscode/extensions.json
Araxeus: Oh right I remembered why I didn't commit it, the vscode folder is in the gitignore

I guess I'll remove it from there

package.json
Araxeus: very weird... If that really is the case, then maybe you should submit a bug report to https://github.com/rome/tools

Araxeus: Wait so `yarn check` throws the same error for you?

Araxeus: Oh so actually it's a known bug with `rome format` and should be fixed in the next version

https://github.com/rome/tools/issues/4221

th-ch: Same with `yarn check` but it does not prevent both commands from running correctly (it's just noisy), so I guess we can live with it!

th-ch (approved):
The merge of recent PRs created conflicts, but otherwise looks good! ✅

Araxeus:
I'm really conflicted about 2 vs 4 spaces

I keep changing my mind, but I'm currently leaning towards 4 because it's much more accessible – it's actually easier to quickly understand complicated code

th-ch:
Rome has been deprecated and with the migration to TS, this PR can now probably be closed!

PR #1065, opened 2023-03-12
[crossfade] add menu options

### LINKED ISSUE #1046: [Feedback]: Crossfade too abrupt
The crossfade plugin is nice, but it sometimes sounds to abrupt. The volume quickly drops, which makes it seem as if the song has suddenly stopped. Making it more gradual would be nice. Sounds great otherwise!

Araxeus:
fix #1046

### DIFF
--- a/plugins/crossfade/front.js
+++ b/plugins/crossfade/front.js
@@ -8,13 +8,7 @@ let transitionAudio; // Howler audio used to fade out the current music
 let firstVideo = true;
 let waitForTransition;
 
-// Crossfade options that can be overridden in plugin options
-let crossfadeOptions = {
-	fadeInDuration: 1500, // ms
-	fadeOutDuration: 5000, // ms
-	exitMusicBeforeEnd: 10, // s
-	fadeScaling: "linear",
-};
+const defaultOptions = require('../../config/defaults').plugins.crossfade;
 
 const getStreamURL = async (videoID) => {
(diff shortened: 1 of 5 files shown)


plugins/crossfade/front.js
```diff
@@ -8,13 +8,7 @@ let transitionAudio; // Howler audio used to fade out the current music
...
 let waitForTransition;
 
-// Crossfade options that can be overridden in plugin options
-let crossfadeOptions = {
-	fadeInDuration: 1500, // ms
-	fadeOutDuration: 5000, // ms
-	exitMusicBeforeEnd: 10, // s
-	fadeScaling: "linear",
-};
+const defaultOptions = require('../../config/defaults').plugins.crossfade;
```
th-ch: nit: considering we override it with the plugin options, I would name it differently (to avoid the confusion about the defaults), eg
```suggestion
const pluginOptions = require('../../config/defaults').plugins.crossfade;
```

th-ch (approved):

plugins/crossfade/front.js
Araxeus: sure but won't matter after https://github.com/th-ch/youtube-music/pull/1064

th-ch (approved):

Araxeus:
btw @th-ch maybe the default `exitMusicBeforeEnd` is a bit too high?

th-ch (approved):

PR #1073, opened 2023-03-15
add starting page option

### LINKED ISSUE #1071: Start app pointed to a specific page
### Feature

Every time I open the app, it starts up on the "Home" page (which is the same as https://music.youtube.com ). I get that Google would prefer that I use their Radio stations (which generates ad revenue), but I always want to listen to my own uploaded songs. To do so, I have to click Library (relative URL `/library`), then Uploads (`/library/uploads`), then Songs (`/library/uploaded_songs`), then Shuffle All.

It would be nice to specify a default start page (in my case `/library/uploaded_songs`) that the app would always go to when I open it up.

### Solution

Perhaps add a setting where the user could choose one of the main sections or subsections of the YT Music app to serve as their personalized start location.

- Home (`/`)
- Explore (`/explore`)
  - New releases (`/new_releases`)
  - Charts (`/charts`)
  - Moods & Genres (`/moods_and_genres`)
- Library
  - Main Library tab (`/library`
    - Playlists (`/library/playlists`)
    - Songs (`/library/songs`)
    - Albums (`/library/albums`)
    - Artists (`/library/artists`)
  - Uploads tab (`/library/uploads`)
    - Playlists (`/library/playlists`) (interestingly, this is the same as above)
    - Songs (`/library/uploaded_songs`)
    - Albums (`/library/uploaded_albums`)
    - Artists (`/library/uploaded_artists`)

### Alternatives

- A simple text entry box for the user to put in their preferred start URL (`https://music.youtube.com/library/uploaded_artists`) or relative URL (`library/uploaded_artists`)
- The existing "Continue where you left off" setting starts you off with the same song you were playing before (that's good!). However, it's in a Now Playing queue of all of your songs, not a "Shuffle All" queue.

Araxeus:
This allows choosing a `startingPage` option which is independent of `resumeOnStart`

fix #1071

image

> I considered using a `custom-electron-prompt` select instead, but this just felt more natural looking. i'm open to feedback


menu.js
```diff
@@ -81,6 +82,17 @@ const mainMenuTemplate = (win) => {
 						config.setMenuOption("options.resumeOnStart", item.checked);
 					},
 				},
+				{
+					label: 'Starting page',
+					submenu: Object.keys(startingPages).map((name) => ({
```
th-ch, line 87: Is there to have no starting page once you have selected any? typically staying on the song page

Araxeus: I'm not totally sure that I understand your question, but if you select the first option (`Home`) then it doesn't do anything since Home is the default starting page

th-ch: Unless I missed something, currently if "resume on last song" is checked, it will load the song URL (`https://music.youtube.com/watch?v=XXX`) - if Home is picked as starting page as well, it will redirect to `https://music.youtube.com/` (and so minimize the player), which is not the same behavior - having the option to unselect a previously starting page would resolve that!

Araxeus: If home is selected, it won't do anything

https://github.com/th-ch/youtube-music/blob/f5758bfe939d60f7d9d562af17279a245e87a977/preload.js#L136-L138

And if something else is selected then yes it will minimize the player page, but I couldn't think of a better solution

th-ch: Isn't it `Home` with a [capital letter](https://github.com/th-ch/youtube-music/pull/1073/files#diff-c7db0c4eb7f7457e25664ef3b731ac3738f118ee769f698b938bd8bd5a409c4eR2)? 

But redirect to the home page feels like a good option, we should just allow unselecting the currently selected starting page for noop :)

Araxeus: You're right, both the typo and not having the option to start there. My bad 

I'm currently writing from my phone lying in bed, so i'll get on that tomorrow :)

Araxeus: fixed by https://github.com/th-ch/youtube-music/pull/1073/commits/45fa963818f9209450d078b4ff75b7dfd9624210 (3 lines added)

th-ch (approved):

PR #1077, opened 2023-03-16
add option to hide the like buttons

### LINKED ISSUE #1075: Remove thumbs up/down buttons
Previously I was able to remove them using uBlock Origin when its integrated through the browser, removing them frees up more room for the song information, since many times you can't see the year it was released or the entire album name, I also literally never use the thumbs up/down buttons so it would be nice if this could be added

Araxeus:
Now you can select if the like buttons style should be `Default` / `Force Show` / `Hidden`
fix #1075 

image

also added migrations

### DIFF
--- a/config/store.js
+++ b/config/store.js
@@ -19,6 +19,11 @@ const migrations = {
 				...pluginOptions,
 			});
 		}
+
+		if (store.get("options.ForceShowLikeButtons")) {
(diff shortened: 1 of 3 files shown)


th-ch (approved):

Gollum69:
ok coolio love u

PR #1080, opened 2023-03-17
[lyrics-genius] Improved reliability of east asian language detection

DereC4:
- Sometimes Korean characters were not being picked up by the expression so I have updated it and it now matches very reliably

temp2


DereC4:
oops

PR #1082, opened 2023-03-17
[lyrics-genius] Improved reliability of east asian language detection #1080

DereC4:
- Sometimes Korean characters were not being picked up by the expression so I have updated it and it now matches very reliably

temp2

### DIFF
--- a/plugins/lyrics-genius/back.js
+++ b/plugins/lyrics-genius/back.js
@@ -7,7 +7,7 @@ const fetch = require("node-fetch");
 
 const { cleanupName } = require("../../providers/song-info");
 const { injectCSS } = require("../utils");
-let eastAsianChars = new RegExp("[\uac00-\ud7af]|[\u1100-\u11ff]|[\u3130-\u318f]|[\ua960-\ua97f]|[\ud7b0-\ud7ff]|[\u3040-\u309F]|[\u3000-\u303F\u3400-\u4DBF\u4E00-\u9FFF]");
+let eastAsianChars = /\p{Script=Han}|\p{Script=Katakana}|\p{Script=Hiragana}|\p{Script=Hangul}|\p{Script=Han}/u;
 let revRomanized = false; 
 
 module.exports = async (win, options) => {
--- a/plugins/lyrics-genius/back.js
+++ b/plugins/lyrics-genius/back.js
@@ -7,7 +7,7 @@ const fetch = require("node-fetch");
 
 const { cleanupName } = require("../../providers/song-info");
 const { injectCSS } = require("../utils");
-let eastAsianChars = new RegExp("[\u{3040}-\u{30ff}\u{3400}-\u{4dbf}\u{4e00}-\u{9fff}\u{f900}-\u{faff}\u{ff66}-\u{ff9f}]");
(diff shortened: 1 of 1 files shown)


plugins/lyrics-genius/back.js
```diff
@@ -7,7 +7,7 @@ const fetch = require("node-fetch");
 
 const { cleanupName } = require("../../providers/song-info");
 const { injectCSS } = require("../utils");
-let eastAsianChars = new RegExp("[\u{3040}-\u{30ff}\u{3400}-\u{4dbf}\u{4e00}-\u{9fff}\u{f900}-\u{faff}\u{ff66}-\u{ff9f}]");
+let eastAsianChars = new RegExp("[\uac00-\ud7af]|[\u1100-\u11ff]|[\u3130-\u318f]|[\ua960-\ua97f]|[\ud7b0-\ud7ff]|[\u3040-\u309F]|[\u3000-\u303F\u3400-\u4DBF\u4E00-\u9FFF]");
```
Araxeus:
```suggestion
const eastAsianChars = new RegExp("\\p{Script=Han}|\\p{Script=Katakana}|\\p{Script=Hiragana}", "u");
```
Sources:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes#scripts_and_script_extensions

* https://www.regular-expressions.info/unicode.html  #Unicode Scripts

Araxeus (changes requested):

plugins/lyrics-genius/back.js
DereC4: oh that's cool; lemme take a look to see if it will work

DereC4: also do you know where one might find the scripts for Hangul? Korean characters are part of that regex as well

DereC4: Oh nevermind found it; I guess we can omit Kanji since it overlaps with Chinese most cases and chances are extremely high Japanese lyrics will have majority Hiragana

DereC4: Hmm seems to be an invalid regex string, maybe I'm doing it wrong. 

On a side note I was checking the second link you gave me and it said that Js doesn't have Unicode support through its RegExp class but does support \uFFFF for matching; does that have anything to do with this?

Araxeus: > On a side note I was checking the second link you gave me and it said that Js doesn't have Unicode support through its RegExp class but does support \uFFFF for matching; does that have anything to do with this?

The source of knowledge on javascript is developer.mozilla.com (first link explains it great)
use the second link only to find Unicode Scripts Codes which are omitted from mozilla from some reason

> Hmm seems to be an invalid regex string, maybe I'm doing it wrong.

Sorry, because you are using `new RegExp` with a string instead of writing a regex literal - we need to escape the <kbd>\\</kbd> (because we want to pass it "as is" to the regex constructor)
I've updated my suggestion above

you could write either

```js
const eastAsianChars = new RegExp("\\p{Script=Han}|\\p{Script=Katakana}|\\p{Script=Hiragana}", "u");
```
or
```js
const eastAsianChars = /\p{Script=Han}|\p{Script=Katakana}|\p{Script=Hiragana}/u;
```

DereC4: ohh that explains it; thanks! 

What's the difference between doing \p{Script = script name} and the \p{script name} ?

Araxeus: If you look at the mozilla docs I sent, it says that in javascript you can only do `\p{Script=<NAME>}`

it has to be exactly `Script=` and not `script=`

Araxeus (approved):
LGTM

DereC4:
nicee

th-ch (approved):

PR #1094, opened 2023-03-31
[DEPRECATED] API Plugin

Zo-Bro-23:
Adds an API plugin. Plugin provides a REST API to do the following:
- View all song-info
- Control the player

API uses keys that are arbitrarily assigned - custom keys can be chosen if desired. The same key can be used on multiple devices and installations. This works very similarly to the SponsorBlock key system.

I have also built an Android remote app that works using this plugin. I will add a link once Google Play approves it.

#### `GET /status`

**Query Params:**
- `key`: Your API key

#### `POST /control`

**Request Body:**
- `key`: Your API key
- `controls`: A list (array) of controls to execute; controls can be:
    - `play`
    - `pause`
    - `playPause`
    - `previous`
    - `next`
    - `like`
    - `dislike`
    - `go10sBack`
    - `go10sForward`
    - `go1sBack`
    - `go1sBack`
    - `shuffle`
    - `switchRepeat`
    - `volumeMinus10`
    - `volumePlus10`
    - `fullscreen`
    - `muteUnmute`
    - `maximizeMinimisePlayer`
    - `goToHome`
    - `goToLibrary`
    - `goToSettings`
    - `goToExplore`
    - `search`
    - `showShortcuts`

## Android App
Download from [Google Play]() or use the latest APK from the `./android/` folder.

> **Warning**
> The API key set using the Android App is semi-permenant. To change the key go to Settings > Apps > Remote > Clear Data > Clear All Data.

### Screenshots
<p align="left">

</p>


Araxeus:
I don't understand why is the server not local?

why would all my data go through your server? this doesn't make any sense to me...

where's the security here? I could literally send commands to your pc if I have your api key?

The server should be on a local network only, there's no reason to expose it to the internet.
see how YTM does this:
https://github.com/ytmdesktop/ytmdesktop/blob/master/src/providers/companionServer.js

in addition, you are missing this line here:

```js
    ipcMain.on('apiLoaded', () => win.webContents.send('setupTimeChangedListener'));
```

plugins/api/front.js
```diff
@@ -0,0 +1,26 @@
...
+        return playerBar?.__data?.likeButtonRenderer_?.likeStatus;
+    }
+
+    const loop = () => {
+        if (currentStatus !== getLikeStatus()) ipcRenderer.send('api-like-button-status', getLikeStatus())
+        currentStatus = getLikeStatus()
+        setTimeout(loop, 100)
+    }
+
+    loop()
```
Araxeus, line 25:
```suggestion
const likeObserver = new MutationObserver((mutations) => {
	mutations.forEach((mutation) => {
		ipcRenderer.send(
			"api-like-button-status",
			mutation.target.getAttribute("like-status"),
		);
	});
});

likeObserver.observe(document.querySelector("ytmusic-like-button-renderer"), {
	attributes: true,
	childList: false,
	characterData: false,
	subtree: false,
	attributeFilter: ["like-status"],
});
```

I didn't know you intended to poll it when you asked about it… this should be much better than a loop that runs 10 times per second

Araxeus: I didn't know you intended to poll it when you asked about it… this should be much better than a loop that runs 10 times per second

Zo-Bro-23: Alright thanks.

Zo-Bro-23:
> Good job on the initiative, but I don't understand why is the server not local?
> 
> why would all my data go through your server? this doesn't make any sense to me...
> 
> where's the security here? I could literally send commands to your pc if I have your api key?
> 
> The server should be on a local network only, there's no reason to expose it to the internet. see how YTM does this: https://github.com/ytmdesktop/ytmdesktop/blob/master/src/providers/companionServer.js
> 
> in addition:
> 
> you are missing the line below, without it you are not guaranteed to get events from `ipcMain.on('timeChanged')`
> 
> ```js
>     ipcMain.on('apiLoaded', () => win.webContents.send('setupTimeChangedListener'));
> ```
> 
> and you should use the new synced config provider, examples in #1064

How about having a separate plugin that uses local endpoints? I personally prefer having a server-based approach so that I can control my computer even when I'm on different WiFi networks. We can have a local API plugin for those worried about security concerns.

plugins/api/back.js
```diff
@@ -0,0 +1,74 @@
+const axios = require("axios");
```
Araxeus, line 1: We already have a `node-fetch` dependency, is this really necessary?

readme.md
```diff
@@ -190,13 +192,13 @@ module.exports = () => {
...
-3. Run `yarn build:OS`
-   - `yarn build:win` - Windows
-   - `yarn build:linux` - Linux
-   - `yarn build:mac` - MacOS
+2. Run ```yarn```
+3. Run ```npm i rimraf```
+4. Run ```yarn build:OS```
+	- ```yarn build:win``` - Windows
+	- ```yarn build:linux``` - Linux
+	- ```yarn build:mac``` - OS X
```
Araxeus, line 201: What's that about?

Araxeus:
> How about having a separate plugin that uses local endpoints? I personally prefer having a server-based approach so that I can control my computer even when I'm on different WiFi networks. We can have a local API plugin for those worried about security concerns.

I still think it's a privacy and security concern to connect to some random external server.
Most users won't even know (or expect) that all their data is going through a 3rd party server.
You should ask @th-ch what he thinks, he has the final say here.

But anyways if this goes through, how about using [websocket](https://www.npmjs.com/package/ws) instead of REST?

And btw, where is the source for the android app?

QuditWolf:
One thing that comes to my mind to do both local and wan apis is just make the api endpoint customizable through an input field.(I don't know if you do that already)

And if the api-server code is open, we can run our own servers local or global.

Or better yet, the server could be run on the local machine running youtube-music inside the app itself.

Then if we can access the device running youtube-music locally, we can key in the local address(on android/controller interface). and if it has a global address we can key that in. 

for global control with devices with non-static or non global ips, we could add an option(for the user to choose) to tunnel the api server thru a ssh tunnel(we don't have to worry about snooping) using something like serveo.net

btw I made an api and a webinterface which fetches data through that api( local api and through an ssh tunnel) to show what I'm listening to on my personal website, and the latency's quite less.
image

readme.md
ArjixWasTaken: I'd assume a horrible auto formatter.

What kind of formatter uses multiline blocks for an inline block tho?
;-;

JellyBrick:
#1909

PR #1098, opened 2023-04-02
Protocol handler arguments are not correctly parsed

pathmann:
The last character from a protocol handler argument is removed, so the argument won't be parsed correctly to call the requested song control function.

So executing youtubemusic://playPause will be parsed to playPaus, on the other hand executing youtubemusic://playPauseX will trigger the playPause song control.

### DIFF
--- a/index.js
+++ b/index.js
@@ -381,7 +381,7 @@ app.on("ready", () => {
 		const uri = `${APP_PROTOCOL}://`;
 		const protocolArgv = commandLine.find(arg => arg.startsWith(uri));
 		if (protocolArgv) {
-			const command = protocolArgv.slice(uri.length, -1);
+			const command = protocolArgv.slice(uri.length);
 			if (is.dev()) console.debug(`Received command over protocol: "${command}"`);
 			handleProtocol(command);
 			return;


Araxeus:
Hmm this is weird, it definitely works on my pc (Windows 10)

what OS are you on?

pathmann:
(Gentoo) Linux 64 bit

What is the purpose (maybe only on Windows) of removing that last character?

Araxeus:
Oh I see the problem now. on windows there is always an appended character - `/`

So we just need to check if it ends with `/`

image

Araxeus:
The code just have to check if it ends with `/` like I said

If you don't want / can't do it yourself, just tell me and ill gladly opena PR

pathmann:
If you don't mind, sure. I'm not very experienced in Javascript :grin:

Araxeus:
Ok FYI next time you fork a repo to open a Pull Request – create a new branch first, don't open PR's from your default branch

Araxeus:
I've opened a PR https://github.com/th-ch/youtube-music/pull/1099 can you try it out?

just check this branch out https://github.com/Araxeus/youtube-music/tree/fix-protocol-handler-on-unix

pathmann:
Yep, works.

Thanks a lot

PR #1116, opened 2023-04-15
fix security issues in dependencies

Araxeus:
aims to fix all the issues in https://github.com/th-ch/youtube-music/security/dependabot

image

* [x] minimatch 3.0.4 is in @electron/universal:
electron-builder@npm:23.6.0 (via npm:^23.6.0)
├─ app-builder-lib@npm:23.6.0 (via npm:23.6.0)
├─ @electron/universal@npm:1.2.1 (via npm:1.2.1)
├─ asar@npm:3.2.0 (via npm:^3.1.0)
│  ├─ glob@npm:7.2.3 (via npm:^7.1.6)
│  └─ minimatch@npm:3.1.2 (via npm:^3.0.4)
├─ dir-compare@npm:2.4.0 (via npm:^2.4.0)
│  └─ minimatch@npm:3.0.4 (via npm:3.0.4)
└─ minimatch@npm:3.1.2 (via npm:^3.0.4)

* [x] yargs-parser 4.2.1 is in electron-icon-builder
electron-icon-builder@npm:2.0.1 (via npm:^2.0.1)
└─ icon-gen@npm:2.1.0 (via npm:^2.0.0)
   └─ svg2png@npm:4.1.1 (via npm:4.1.1)
      └─ yargs@npm:6.6.0 (via npm:^6.5.0)
         └─ yargs-parser@npm:4.2.1 (via npm:^4.2.0)

* [x]  request 2.88.2 is in electron-icon-builder
     electron-icon-builder@npm:2.0.1 (via npm:^2.0.1)
      └─ icon-gen@npm:2.1.0 (via npm:^2.0.0)
         └─ svg2png@npm:4.1.1 (via npm:4.1.1)
            └─ phantomjs-prebuilt@npm:2.1.16 (via npm:^2.1.14)
               └─ request@npm:2.88.2 (via npm:^2.81.0)
* [x] svg2png 4.1.1 is in electron-icon-builder
electron-icon-builder@npm:2.0.1 (via npm:^2.0.1)
      └─ icon-gen@npm:2.1.0 (via npm:^2.0.0)
         └─ svg2png@npm:4.1.1 (via npm:4.1.1)

* [x] xml2js is in mpris-service and electron-icon-builder

> **Note**
> If you look closely you'll notice that most of those are because of `electron-icon-builder`

PR #1117, opened 2023-04-15
Bump version to 1.20.0

th-ch:


Araxeus:
Would be nice to merge  #1116 before releasing

PR #1120, opened 2023-04-19
[DEPRECATED] Feature: Adds integration of topik/youtube-music-obs-widget as plugin

topik:
Hello,

I got a [request](https://github.com/topik/youtube-music-obs-widget/issues/4) to add a plugin for my [obs widget](https://github.com/topik/youtube-music-obs-widget) so here it is. 

Plugin makes a simple http server and give a info about current song in JSON. The JSON structure is the same as in other YT music software that is being used so the obs widget is compatible.

Feel free to rename the plugin.

Thanks

### DIFF
--- a/plugins/obs-widget/back.js
+++ b/plugins/obs-widget/back.js
@@ -10,7 +10,7 @@ const data = {
 		seekbarCurrentPosition: 0,
 	},
 	track: {
-		author: "",
+		author: [],
 		title: "",
 		album: "",
 		cover: "",
@@ -31,14 +31,20 @@ module.exports = async (win) => {
 			res.end(JSON.stringify({}));
 			return;
 		}
-		data.player.hasSong = true;
-		data.player.isPaused = currentSongInfo.isPaused;
-		data.player.seekbarCurrentPosition = currentSongInfo.elapsedSeconds;
-		data.track.author = [currentSongInfo.artist];
-		data.track.title = currentSongInfo.title;
-		data.track.album = currentSongInfo.album;
-		data.track.cover = currentSongInfo.imageSrc;
-		data.track.duration = currentSongInfo.songDuration;
+                Object.assign(data, {
+                    player: {
+                        hasSong: true,
+                        isPaused: currentSongInfo.isPaused,
+                        seekbarCurrentPosition: currentSongInfo.elapsedSeconds
+                    },
+                    track: {
+                        author: [currentSongInfo.artist],
+                        title: currentSongInfo.title,
+                        album: currentSongInfo.album,
+                        cover: currentSongInfo.cover,
+                        duration: currentSongInfo.duration
+                    }
+                });
 		res.end(JSON.stringify(data));
 	};
 	const server = http.createServer(requestListener);
--- a/plugins/obs-widget/back.js
+++ b/plugins/obs-widget/back.js
@@ -1,0 +1,56 @@
+const {ipcMain} = require("electron");
+const http = require("http");
+const registerCallback = require("../../providers/song-info");
+
+let currentSongInfo;
+const data = {
+	player: {
+		hasSong: false,
+		isPaused: true,
+		seekbarCurrentPosition: 0,
+	},
+	track: {
+		author: "",
+		title: "",
+		album: "",
+		cover: "",
+		duration: 0,
+	}
+};
+
+/** @param {Electron.BrowserWindow} win */
+module.exports = async (win) => {
+	const requestListener = function (req, res) {
+		if(req.url !== '/query'){
+			res.writeHead(404);
+			res.end("404 Not found!");
+		}
+		res.setHeader("Access-Control-Allow-Origin", "*");
+		res.writeHead(200);
+		if (currentSongInfo !== undefined && !currentSongInfo.title && !currentSongInfo.artist) {
+			res.end(JSON.stringify({}));
+			return;
+		}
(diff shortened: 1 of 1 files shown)


Araxeus:
This looks extremely similar to https://github.com/th-ch/youtube-music/blob/master/plugins/tuna-obs/back.js

Could you give a quick explanation why a new plugin is needed?

plugins/obs-widget/back.js
```diff
@@ -0,0 +1,56 @@
...
+			return;
+		}
+		data.player.hasSong = true;
+		data.player.isPaused = currentSongInfo.isPaused;
+		data.player.seekbarCurrentPosition = currentSongInfo.elapsedSeconds;
+		data.track.author = [currentSongInfo.artist];
+		data.track.title = currentSongInfo.title;
+		data.track.album = currentSongInfo.album;
+		data.track.cover = currentSongInfo.imageSrc;
+		data.track.duration = currentSongInfo.songDuration;
```
ArjixWasTaken: Just a small lifehack :^)

```suggestion
                Object.assign(data, {
                    player: {
                        hasSong: true,
                        isPaused: currentSongInfo.isPaused,
                        seekbarCurrentPosition: currentSongInfo.elapsedSeconds
                    },
                    track: {
                        author: [currentSongInfo.artist],
                        title: currentSongInfo.title,
                        album: currentSongInfo.album,
                        cover: currentSongInfo.cover,
                        duration: currentSongInfo.duration
                    }
                });
```

plugins/obs-widget/back.js
```diff
@@ -0,0 +1,56 @@
...
+
+let currentSongInfo;
+const data = {
+	player: {
+		hasSong: false,
+		isPaused: true,
+		seekbarCurrentPosition: 0,
+	},
+	track: {
+		author: "",
```
ArjixWasTaken: Since you assign a list to it, shouldn't the initial value be an empty list?

```suggestion
		author: [],
```

ArjixWasTaken (commented):
PS: You should use a formatter on your code :^)

topik:
> PS: You should use a formatter on your code :^)

You are absolutely right, no idea why i didn't do that. I would do it in any other language :| ... Also thanks for pointing out wrong variable type.

JellyBrick:
@topik 
Could you rework this PR? The entire plugin system has been reworked.

JellyBrick:
If you are still interested, this may be helpful: https://github.com/th-ch/youtube-music/pull/2723/files

ArjixWasTaken:
Nah, a rewrite to work with the new codebase wouldn't be enough.
I did just that

```typescript
import { createPlugin } from '@/utils';
import { t } from '@/i18n';

import { Hono } from 'hono';
import { serve, ServerType } from '@hono/node-server';
import registerCallback, { SongInfo } from '@/providers/song-info';

let server: ServerType | null = null;
const port = 9863;

export default createPlugin({
  name: () => t('plugins.obs-widget.name'),
  description: () => t('plugins.obs-widget.description'),

  authors: ['ArjixWasTaken'],
  restartNeeded: false,

  backend: {
    start() {
      let songInfo: SongInfo | null = null;
      registerCallback((info) => {
        songInfo = info;
      });

      const app = new Hono();

      app.get('/query', async (ctx) => {
        ctx.res.headers.set('Access-Control-Allow-Origin', '*');
        return ctx.json({
          player: {
            hasSong: !!songInfo,
            isPaused: songInfo?.isPaused ?? false,
            seekbarCurrentPosition: songInfo?.elapsedSeconds ?? 0,
          },
          track: {
            author: songInfo ? [songInfo?.artist] : [],
            title: songInfo?.title ?? '',
            album: songInfo?.album ?? '',
            cover: songInfo?.imageSrc ?? '',
            duration: songInfo?.songDuration ?? 0,
          },
        });
      });

      server = serve({ fetch: app.fetch, port });
    },
    stop() {
      server?.close();
      server = null;
    },
  },
});
```

but when I saw the client code, it doesn't even use the `/query` endpoint, it uses `/api/v1/*` stuff specific to ytmd

ArjixWasTaken:
One solution would be to fake compatibility with the ytmd API, and only implement the endpoints used, but do we want that?

ArjixWasTaken:
Faking ytmd compatibility requires socket.io as a dependency, and I am not a fan of that.
But hey, it can be done

image

topik:
If there is an accessible endpoint, I can change the widget so it works with both programs, so there will be no authentication for https://github.com/th-ch/youtube-music/tree/master. The question is, who's gonna make the endpoint and how. 

PS: good job 😂

ArjixWasTaken:
> If there is an accessible endpoint, I can change the widget so it works with both programs

we do not currently have websockets, but the plugin `api-server` does have an endpoint to get the player state as well as the current song, so it would be similar to the `/query` you used to call in the past

I do see that having websockets would be useful, so enhancing the `api-server` plugin with it would be an option

topik:
Then I can just use the plugin that already exists  🙂 I think that would be the easiest solution.

ArjixWasTaken:
you'll have to deal with authentication there as well 😔, although it is optional iirc

ArjixWasTaken:
since I'm invested in this topic already, leave it to me, I'll make a PR to your widget

ArjixWasTaken:
man I completely forgot about this

PR #1124, opened 2023-04-25
revert adblocker bump

### LINKED ISSUE #1105: Adblock plugin makes application laggy and unresponsive
I built the app myself a few days ago, since the releases are pretty old, but now, if I play music from my playlist after a while it gets very slow and laggy, as if my internet or CPU wasn't good enough. This is a new issue though and my internet and CPU were able to handle everything fine before.
I'm going to attach a video how it looks, note that I had the "Crossfade" plugin on. This is not the cause, but does make it worse. Also, I had to compress the video down a lot, the quality is better normally.

https://user-images.githubusercontent.com/96952506/230140583-a3840e6e-6a5c-4fd1-b703-d76a774f8b16.mp4

Also, I'm running on Feren OS x86_64 (Ubuntu)
CPU: AMD Ryzen 3 3200G (4) @ 3.600GHz
GPU: NVIDIA GeForce GTX 660
Speedtest at time of filming: 35mbps Down;  10mbps Up; 14 Ping

Installed via self-compiled .deb package

Araxeus:
This is supposed to fix the performance regression introduced in #1100 by reverting to https://github.com/ghostery/adblocker/releases/tag/v1.26.3

fix #1105

Will need to re-open the issues that were closed in #1100

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
 		"npm": "Please use yarn instead"
 	},
 	"dependencies": {
-		"@cliqz/adblocker-electron": "^1.26.5",
+		"@cliqz/adblocker-electron": "1.26.3",
 		"@ffmpeg/core": "^0.11.0",
 		"@ffmpeg/ffmpeg": "^0.11.6",
 		"@foobar404/wave": "^2.0.4",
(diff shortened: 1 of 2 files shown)


Lucasamiel0406:
Confirmed to fix performance of [mentioned issue]( https://github.com/th-ch/youtube-music/issues/1105).

th-ch (approved):

Issue #1129, opened 2023-05-02
Ads ads ads...


colineberly:
I noticed on master ads are coming thru constantly, before first song load, after 1 or 2 songs. Seems to behave like normal if there wasn't an adblocker loaded. Any way to manually remedy this?

Lucasamiel0406:
Due to a new update on how the Adblock dependency/package works, it fixed ads in the beginning and started making the program unresponsive. In the end, a bump revert was needed.

Take a look on #1040, test it and provide feedback so we can get it soon.

PR #1040, opened 2023-02-24
experimental faster adblocker

Araxeus:
This requires user testing.

Anyone that can test and report how it works for them would be greatly appreciated !

Either clone and build this branch, or download from https://github.com/Zo-Bro-23/youtube-music/releases/tag/v1.19.0

PR #1134, opened 2023-05-08
Multiple implementations for the Adblocker plugin

th-ch:
This PR bumps cliqz/adblocker-electron (seeing ads otherwise - the bump should fix https://github.com/th-ch/youtube-music/issues/1129) and also implements [in-player blocking](https://github.com/th-ch/youtube-music/pull/1040) as an alternative (users can choose between the 2) in case of performance issues like https://github.com/th-ch/youtube-music/issues/1105.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
 		"npm": "Please use yarn instead"
 	},
 	"dependencies": {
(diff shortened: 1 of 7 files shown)


th-ch:
Merging this one and closing https://github.com/th-ch/youtube-music/pull/1040

PR #1138, opened 2023-05-11
add xesam:url mpris from songInfo.url

QuditWolf:
there's an unused mpris property, xesam:url...
which might be useful for people making their own integrations to show what's currently playing on different platforms and integrations. (I use it to show what's playing on youtube-music on my website)

anyways, just a property add, fetching the url from songInfo.url
seems minor

thx


th-ch (approved):

th-ch:
Thanks for the improvement!

Issue #1127, opened 2023-04-29
ctrl with + not working


viperflame:
ctrl with - works but their is an issue with ctrl with + , not working .. plzz resolve this . see what happened , its so small i cant see this
![Screenshot (72)](https://github.com/th-ch/youtube-music/assets/132129346/422a9a10-0822-476b-af8a-94283306e927)
![Uploading Screenshot (72).png…]()

JellyBrick:
Fixed at #1163

viperflame:
nope its not fixed still facing the issue with latest update
![Screenshot (183)](https://github.com/th-ch/youtube-music/assets/132129346/098598ac-b49e-4139-aeb0-b84013174139)

JellyBrick:
> nope its not fixed still facing the issue with latest update ![Screenshot (183)](https://user-images.githubusercontent.com/132129346/273390057-098598ac-b49e-4139-aeb0-b84013174139.png)

It's not released yet.

JellyBrick:
[v2.0.0](https://github.com/th-ch/youtube-music/releases/tag/v2.0.0) released!

If you using in-app-menu, please you Ctrl+I and Ctrl+O

PR #1154, opened 2023-05-21
Zoom-in Keyboard Shortcut Fix

DanielTheMiller:
Tackles #1127
Added a global keyboard shortcut to allow the window's webContents to zoom in my an arbitrary zoom factor of 0.1

(Previously only the zoom-out keyboard shortcut would work. I'm not even sure where that one is implemented)


DanielTheMiller:
Hey, I hope this is useful. This is my first open source contribution on GitHub. 
I hope this is a useful adjustment, and I hope I've implemented it in a good place for you.
Looking forward to feedback 😅

DanielTheMiller:
> I noticed the same thing a few days ago. Apparently ctrl+_shift_++ is the keybind for zoom. a little unintuitive I think.

Ah, so it is. Maybe the keyboard shortcut label in the View menu should be changed to reflect this?

PR #1155, opened 2023-05-22
add player method "open" for mpris

QuditWolf:
small addition (just one line)
add functionality for the `playerctl open url` command.
opens the given uri through win.loadURL(url), if there is a better method, please edit/suggest.

perhaps we could also add regex checking for url to see if it's a valid url and it's music.youtub.com one, youtube also sometimes shortens urls so that might be a little complex. 
these are first set of commits/programs for js, so I'm not that proficient :grimacing:


th-ch:
Thanks, ported in https://github.com/th-ch/youtube-music/commit/8b65f1d6e408bc110f5c412b0b1a151194062329 in TS!

PR #1159, opened 2023-05-24
Fix for windows zoom (ScaleFactor)

Sickjuicy:
Fixes the ScaleFactor with windows zoom.

Adds the scalefactor, for one monitor or multiple, testet on 2 monitors and one, on my laptop and pc also sandbox.

Laptop zoom: 150% and than 100%
PC two monitor:

zoom: 100%
zoom: 150%

References to  #868


Sickjuicy (commented):
Adds the scalefactor, for one monitor or multiple, testet on 2 monitors and one, on my laptop and pc  also sandbox.

Laptop zoom: 150% and than 100%
PC two monitor: 
1. zoom: 100%
2. zoom: 150%

plugins/startup-FN/back.js
JellyBrick, line 1: Why did you add this plugin?

plugins/startup-FN/front.js
JellyBrick, line 1: Same as the comment above.

JellyBrick (changes requested):

plugins/startup-FN/back.js
Sickjuicy: I have not seen the function in the shortcuts plugin, the other plugin is there for when I start the PC that you can use the function keys directly without the need to move the program into focus. I have noticed in general in the installer version the function keys often do not work but in the standalone version. I have the plugins in the after no also deleted again, unfortunately, this is very confusing in github just, should I reopen the post? 
PS: english is not my main language

Sickjuicy: I deleted the plugin.

index.js
```diff
@@ -177,7 +177,28 @@ function createMainWindow() {
...
+	const size = config.get("window-size");
+	const position = config.get("window-position");
+	
+	if (size && size.width && size.height) {
+	  const scaledSize = {
+		width: size.width / scaleFactor,
+		height: size.height / scaleFactor,
+	  };
+	  win.setSize(scaledSize.width, scaledSize.height);
+	}
```
JellyBrick:
```suggestion
	if (size && size.width && size.height) {
		const scaledSize = {
			width: size.width / scaleFactor,
			height: size.height / scaleFactor,
		};
		win.setSize(scaledSize.width, scaledSize.height);
	}
```

index.js
```diff
@@ -177,7 +177,28 @@ function createMainWindow() {
...
+	
+	if (position && position.x && position.y) {
+	  const scaledPosition = {
+		x: position.x / scaleFactor,
+		y: position.y / scaleFactor,
+	  };
+	  win.setPosition(scaledPosition.x, scaledPosition.y);
+	}
```
JellyBrick:
```suggestion
	if (position && position.x && position.y) {
		const scaledPosition = {
			x: position.x / scaleFactor,
			y: position.y / scaleFactor,
		};
		win.setPosition(scaledPosition.x, scaledPosition.y);
	}
```

JellyBrick (changes requested):

JellyBrick (approved):

th-ch:
Thanks, ported in https://github.com/th-ch/youtube-music/commit/d36fb592d0ad12152e2d46429bbb8bb13519e63e in TS!

PR #1163, opened 2023-05-28
Changed ZoomIn ZoomOut Shortcuts to Ctrl+I/Ctrl+O

DanielTheMiller:
Changed ZoomIn ZoomOut Shortcuts to Ctrl+I / Ctrl+O (Or cmd+I/cmd+O on linux/mac)

I attempted to change the keyboard shortcut for zoom in to Ctrl+ over the current default Ctrl+Shift++ (which is labelled Ctrl++) but that wasn't binding for some reason.

A simple compromise that seems to work is to just use Ctrl+I Ctrl+O instead. The slightly less conventional zoom keyboard shortcut. There seems to be no conflicting keyboard shortcut.

I hope this is useful 😄


DanielTheMiller:
Tackles #1127

DanielTheMiller:
I see, this is an underlying issue in electron. Seems it might only affect UK or German keyboards? https://github.com/electron/electron/issues/15496 https://github.com/electron/electron/issues/29996

bradisbell:
@DanielTheMiller I can't speak for your app, but the Electron bug is definitely affecting American keyboards as well.

th-ch:
Thank you, I ported the change in https://github.com/th-ch/youtube-music/commit/da9cb8e2f5c82d5bb6c4761db8a6d315b6f83a80 to TS!

Issue #770, opened 2022-06-29
[Feature Request] Extensive and dynamic themes


kerichdev:
Since CSS/HTML injection plugins require rebuilding the whole client, which is highly impractical, it would be really nice having an already existing theme engine, like stylus, added to the client, BUT ideally - [this](https://chrome.google.com/webstore/detail/themesong-for-youtube-mus/bgfiegdbajagebogifobkhambpljbfmk) would be the best one, if there's a way to contact the developer for the source code, or decombile the extension - it would be totally worth it, since it includes dynamic album cover based themes for both light and dark mode.

[Link to the reddit post of the developer](https://www.reddit.com/r/YoutubeMusic/comments/t8x9e1/im_not_a_big_fan_of_the_default_black_theme_so_i/)

[Developer's reddit profile](https://www.reddit.com/user/GraphiteEaterArt/)

Araxeus:
You could maybe ask him if he wants to contribute here?

KristofferTroncoso:
Oh hey guys, it's me the ThemeSong dev (reddit: GraphiteEaterArt).
Thanks for reaching out to me @kerichdev . Glad you like the extension.
I just downloaded this YouTube Music desktop app and am browsing the code now. This is awesome!
My ThemeSong (beta) extension is currently spaghetti and is wildly changing every month so for now I'll look into contributing the core features of ThemeSong so we can get the basic dynamic theme features into this desktop app asap. 
Thanks for putting this on my radar. Great work from all the contributors!

- KT

PR #1200, opened 2023-07-15
Album Color Theme plugin

EdiBOI25:
Created a plugin which changes the color of certain elements based on the cover album to make a theme like the Android/iOS app.


MiepHD:
Fixes #1209, fixes #770

th-ch:
Thanks, the plugin has been added in https://github.com/th-ch/youtube-music/commit/063ba1b6c7c7f8e3454e3b4536705f6399c217ab in TS!

Issue #1199, opened 2023-07-14
"Remove Upgrade Button" doesn't remove upgrade button anymore


FQQD:
Since today's interface update, the "Remove Upgrade Button" option doesn't remove the upgrade button anymore.

![image](https://github.com/th-ch/youtube-music/assets/96952506/4fe4ba18-e8e2-4338-8426-be4634823e04)

pinkkuh:
this issue is actually related to the bypass-age restriction #1205 .. the remove upgrade button will work if the bypass-age restriction is disabled (in my case). you can reproduce this issue. try it if it's the same.

JellyBrick:
Fixed at #1206

Rairof:
please reopen this post as this issue still persists (at least on v3.2.2.0) due to the feature being inconsistent on hiding the button every time the app restarts
![image](https://github.com/th-ch/youtube-music/assets/107706560/a55c40dd-2d14-4b8e-a8a8-54374f007807)

JellyBrick:
[...]

PR #1206, opened 2023-07-17
Fix Remove upgrade button

GoudronViande24:
Since last visual update from YouTube, the Remove upgrade button visual tweak does not work.

I made this fix, seems to work on my PC, but since it's just some CSS I don't think there is any reason it would not work on some other platform.

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -139,7 +139,7 @@ function onApiLoaded() {
 
 	// Remove upgrade button
 	if (config.get("options.removeUpgradeButton")) {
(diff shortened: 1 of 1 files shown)


GoudronViande24:
Closes #1199

MiepHD:
It's easier like this. You don't have to create a new style tag for one single attribute.
``
const upgradeButton = $('ytmusic-guide-section-renderer #items ytmusic-guide-entry-renderer:last-child');
if (upgradeButton) upgradeButton.style.display = "none";
``

MiepHD (approved):

th-ch (approved):
Thanks for fixing!

PR #1234, opened 2023-09-02
Fixed canvas injected at the wrong position

### LINKED ISSUE #1233: Using Visualizer Results to black video screen
OS: Windows 11.
I'm not using any other plugins.
In the screenshot it's paused but looks exactly the same either way.

Not sure if I'm doing something wrong or if it's broken
Screenshot 2023-09-01 224159

MiepHD:
The wrong position caused that the canvas only showed up when `Plugins > video-toggle > Mode > Custom toggle` was enabled
It's now a child of #main-panel like in video-toggle.
Fixes #1233

### DIFF
--- a/plugins/visualizer/empty-player.css
+++ b/plugins/visualizer/empty-player.css
@@ -1,9 +1,5 @@
-#player {
-	margin: 0 !important;
-	background: black;
-}
-
-#song-image,
-#song-video {
-	display: none !important;
+#visualizer {
+	position: absolute;
+	z-index: 1;
+	background-color: black;
 }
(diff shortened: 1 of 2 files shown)


th-ch (commented):
With the change, the visualiser does not seem to work in full screen, is it expected?

MiepHD:
Works the right way now 😅

JellyBrick:
Could you incorporate this fix into #1235?

th-ch:
Thanks for the fix, it has been applied in https://github.com/th-ch/youtube-music/commit/0229ccaa1ea5f69b37db415fac3ae033439a639e !

PR #1279, opened 2023-10-08
feat: fallback to DOM window controls on platforms without native support

### LINKED ISSUE #1274: [Feature Request]: Fallback to DOM window controls on platforms without native support
### Preflight Checklist

### Problem Description

Ironically, after filing #825 and it being implemented, I now use a linux distro as my primary/host OS. And currently, Electron's API does not support gnome/kde window controls, so currently, with in-app-menu enabled, there are no window controls.
image

### Proposed Solution

When native window controls are unavailable and in-app-menu is enabled, the app should fallback to the previous method of DOM based controls.

### Alternatives Considered

If Electron does not provide information on whether native controls are supported, it should be hardcoded specifically for the linux platform.

### Additional Information

_No response_

JellyBrick:
- resolves #1274 

image


nnnlog (approved):
LGTM

PR #1296, opened 2023-10-09
Fix Library removed for Premium users

Lucasamiel0406:
As by now, the code removes the last child of the YT's buttons sidebar. It's good for non-premium users but affects premium users, as it removes the "Library" button.

This small fix targets the 4th child (usually the Upgrade button location) instead of last child.

A bad move/practice, but does its job and remove the Upgrade button while not removing the Library one.

### DIFF
--- a/preload.ts
+++ b/preload.ts
@@ -187,7 +187,7 @@ function onApiLoaded() {
   // Remove upgrade button
   if (config.get('options.removeUpgradeButton')) {
     const styles = document.createElement('style');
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
If you are ok with the selector breaking when they change their icon, then you can use

```css
ytmusic-guide-entry-renderer:has(path[d*="0 016-6zm0-3a9 9 0 109 9 9 9 0 00-9-9m0-1A10"]) {
    display: none;
}
```

JellyBrick (approved):

PR #1297, opened 2023-10-10
Fix downloader Korean filename on Windows, etc.

nnnlog:
## Abstract
There is various unicode normalize in UTF-8. [#See here in Korean example](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#%ED%95%9C%EA%B8%80%EC%97%90_normalize_%EC%82%AC%EC%9A%A9%ED%95%98%EA%B8%B0) Windows uses `NFC`, but MacOS uses `NFD`. If download music which includes Korean in title, file name in Windows is seemed as below.
image
This PR resolves ubove problem. See below.
image

## Changes
`filenamify` returns always file name normalized `NFD`, not depending on OS. There is not standard for seperating OS in Javascript Runtime Environment, therefore I think it is not good idea for modify `filenamify`. This PR check current OS, then if not MacOS, normalize file name to `NFC`.

### DIFF
--- a/plugins/downloader/back.ts
+++ b/plugins/downloader/back.ts
@@ -199,10 +199,13 @@ async function downloadSongUnsafe(
     presetSetting = presets[preset];
   }
 
-  const filename = filenamify(`${name}.${presetSetting?.extension ?? 'mp3'}`, {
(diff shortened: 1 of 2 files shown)


JellyBrick (approved):
LGTM

PR #1316, opened 2023-10-14
feat: migrate from `npm` to `pnpm`

JellyBrick:
- `pnpm` prevents ghost dependencies (e.g. https://github.com/th-ch/youtube-music/blob/b316aa230144107b06e99be58a67f01d61a9d7b5/package-lock.json#L2982 and https://github.com/th-ch/youtube-music/blob/b316aa230144107b06e99be58a67f01d61a9d7b5/package-lock.json#L323 is ghost dependencies)


package.json
```diff
@@ -87,33 +87,47 @@
...
     "lint": "eslint .",
     "changelog": "auto-changelog",
-    "release:linux": "npm run clean && npm run build && electron-builder --linux -p always -c.snap.publish=github",
-    "release:mac": "npm run clean && npm run build && electron-builder --mac -p always",
-    "release:win": "npm run clean && npm run build && electron-builder --win -p always",
+    "release:linux": "pnpm clean && pnpm build && electron-builder --linux -p always -c.snap.publish=github",
+    "release:mac": "pnpm clean && pnpm build && electron-builder --mac -p always",
+    "release:win": "pnpm clean && pnpm build && electron-builder --win -p always",
     "typecheck": "tsc -p tsconfig.json --noEmit"
   },
```
ArjixWasTaken: Is this really a needed change?

Personally I like it, but this kinda forces the user that wants to build from source w/o intentions of contributing to use pnpm.

One example would be the [AUR package for youtube-music](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=youtube-music#n27), this is a breaking change.

`npm run build` works fine for us pnpm users, but the reverse does not hold true.

ArjixWasTaken (changes requested):

package.json
ArjixWasTaken: `pnpm build` still works w/o the script changing is what I mean

JellyBrick: This package uses the `pnpm patch` (formerly `patch-package` in npm) feature. so I think it should depend on pnpm.

ArjixWasTaken: oh, yeah then its ok

JellyBrick: https://github.com/NixOS/nixpkgs/pull/260995

I also found another package for `nixOS` that depends on `npm`, so I will hold off on changing it

PR #1366, opened 2023-11-07
codegen: Make vite plugin

ArjixWasTaken:
A vite plugin that automatically imports plugins in each context.

Warning: the `backend` context is not using this plugin, because as of now it is not fully compatible with it.


ArjixWasTaken:
Currently the preload script fails to load, the reason is unknown, will investigate this later.

PR #1385, opened 2023-11-08
feat: plugin auto-importer with `vite-plugin-resolve`

JellyBrick:
cc @ArjixWasTaken @Su-Yong

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
     'import/newline-after-import': 'error',
     'import/no-default-export': 'off',
     'import/no-duplicates': 'error',
+    'import/no-unresolved': ['error', { ignore: ['^virtual:'] }],
     'import/order': [
       'error',
       {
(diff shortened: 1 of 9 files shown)


Su-Yong (approved):

ArjixWasTaken:
Instead of having a named export, how about having a default export?

That way, you don't need to alias the imported constant every time.

JellyBrick:
> Instead of having a named export, how about having a default export?
> 
> That way, you don't need to alias the imported constant every time.

https://basarat.gitbook.io/typescript/main-1/defaultisbad

For the above reasons, I wanted to avoid using the `default export`. What do you think?

ArjixWasTaken:
Hmm, then what about having a different named export for each context?
Just like `feat/codegen`.

JellyBrick:
> Hmm, then what about having a different named export for each context? Just like `feat/codegen`.

Before I do this, I want to rename `front.ts` to `renderer.ts` and `back.ts` to `main.ts`.
(`frontPluginList` -> `rendererPluginList`, `backPluginList` -> `mainPluginList`)

So, is it okay if I merge this pull request before doing a rename task?

Issue #1426, opened 2023-11-17
Explain "Start at login" better


inson1:
I didnt know if it starts on login page for youtube account or after you login to pc. Yes, now that I have read the PR, I know, but I think it needs better name.

PR #1459, opened 2023-11-28
"Start at login" to "Start at PC login"

inson1:
Fixes:  #1426 
I think this could be enough. But if somebody has better idea, please share it.

### DIFF
--- a/src/config/defaults.ts
+++ b/src/config/defaults.ts
@@ -25,7 +25,7 @@ export interface DefaultConfig {
     alwaysOnTop: boolean;
     hideMenu: boolean;
     hideMenuWarned: boolean;
-    startAtLogin: boolean;
+    startAtPCLogin: boolean;
     disableHardwareAcceleration: boolean;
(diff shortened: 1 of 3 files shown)


JellyBrick:
How about "Auto-launch at startup"?

inson1:
@JellyBrick 
What about "Auto-launch at PC startup"  (I like to be specific, so its clear and there is no phone support anyway)
Should I Change something, because of breaking change? Is there some migration tool?

ArjixWasTaken:
I assume PC Login means that it runs whenever the user logins, not whenever the pc is powered on.
Logging out and logging back in would still count.

It is misleading to say pc startup

inson1:
@ArjixWasTaken @JellyBrick So in that case "Start at PC login"? Or better ideas?

inson1:
@JellyBrick ?

JellyBrick:
> @JellyBrick ?

we're now using Weblate

PR #1496, opened 2023-12-04
fix: Homebrew latest release url parsing

gmcinalli:
Fixes this error when installing via Homebrew.

```
Error: No such file or directory @ rb_sysopen - https://api.github.com/repos/th-ch/youtube-music/releases/latest
```

### DIFF
--- a/youtube-music.rb
+++ b/youtube-music.rb
@@ -6,7 +6,7 @@ cask "youtube-music" do
   homepage "https://github.com/th-ch/youtube-music"
 
   # Fetch the latest release version from GitHub API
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Technically homebrew is a thing for Linux as well, but from what I see that is not handled here.

Does anyone even use homebrew on Linux?

gmcinalli:
I've only fixed the current Cask file.

JellyBrick (approved):

shredthaGNAR:
still getting same error when trying to install/upgrade via Homebrew on MacOS Sonoma Beta

gmcinalli:
@shredthaGNAR https://github.com/th-ch/homebrew-youtube-music

PR #1505, opened 2023-12-08
Skip Disliked Songs

### LINKED ISSUE #638: [Feature request] Skip disliked songs
Like the title says

MiepHD:
Fixes #638

### DIFF
--- a/src/plugins/skip-disliked-songs/index.ts
+++ b/src/plugins/skip-disliked-songs/index.ts
@@ -1,0 +1,16 @@
+import { t } from '@/i18n';
+import { createPlugin } from '@/utils';
+
+export default createPlugin({
+    name: () => t('plugins.skip-disliked-songs.name'),
+    description: () => t('plugins.skip-disliked-songs.description'),
+    restartNeeded: true,
+    config: {
+      enabled: true,
+    },
+    renderer() {
+        new MutationObserver(() => {
+            if (document.querySelector("#like-button-renderer")?.getAttribute("like-status") == "DISLIKE") document.querySelector("tp-yt-paper-icon-button.next-button")?.click();
+        }).observe(document.querySelector("#like-button-renderer"), { attributes: true, childList: false, subtree: false });
+    },
+  });  
--- a/src/plugins/skip-disliked-songs/index.ts
+++ b/src/plugins/skip-disliked-songs/index.ts
@@ -1,34 +1,41 @@
-import { t } from '@/i18n';
-import { createPlugin } from '@/utils';
+import { t } from "@/i18n";
+import { createPlugin } from "@/utils";
 
 export default createPlugin({
-    name: () => t('plugins.skip-disliked-songs.name'),
-    description: () => t('plugins.skip-disliked-songs.description'),
-    restartNeeded: false,
-    renderer: {
-        observer: null as MutationObserver | null,
-        async start() {
-            const likeBtn = await this.waitForElem('#like-button-renderer');
-            this.observer = new MutationObserver(() => {
-                if (likeBtn?.getAttribute("like-status") == "DISLIKE")
-                    document.querySelector("tp-yt-paper-icon-button.next-button")?.click();
-            });
-            this.observer.observe(likeBtn,
-                { attributes: true, childList: false, subtree: false });
-        },
-        stop() {
-            this.observer?.disconnect();
-        },
-        waitForElem(selector) {
-            return new Promise(resolve => {
-                const interval = setInterval(() => {
-                    const elem = document.querySelector(selector);
-                    if (!elem) return;
-        
-                    clearInterval(interval);
-                    resolve(elem);
-                });
-            });
-        }
-    }
+  name: () => t("plugins.skip-disliked-songs.name"),
+  description: () => t("plugins.skip-disliked-songs.description"),
+  restartNeeded: false,
+  renderer: {
+    observer: null as MutationObserver | null,
+    start() {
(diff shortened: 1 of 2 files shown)


src/plugins/skip-disliked-songs/index.ts
```diff
@@ -0,0 +1,16 @@
+import { t } from '@/i18n';
+import { createPlugin } from '@/utils';
+
+export default createPlugin({
+    name: () => t('plugins.skip-disliked-songs.name'),
+    description: () => t('plugins.skip-disliked-songs.description'),
+    restartNeeded: true,
+    config: {
+      enabled: true,
```
JellyBrick:
```suggestion
```

If the plugin changes the default behavior of `YouTube Music`, `enabled` should be `false` by default.

ArjixWasTaken:
I highly suggest that you run that code through a code formatter, those one-liners are too long.
Also, there is no guarantee that the elements you are querying actually exist, please put some logic to handle that.

I can help you with that if you are not sure how to do it.
Last but not least, please add a `stop()` callback, this plugin should not require a restart in order to be fully disabled.

JellyBrick (changes requested):

src/plugins/skip-disliked-songs/index.ts
ArjixWasTaken: Or entirely remove the config object, since no other config properties aside from `enabled` exist.

MiepHD:
> Also, there is no guarantee that the elements you are querying actually exist, please put some logic to handle that.
I'm not sure how I could handle that

ArjixWasTaken:
> I'm not sure how I could handle that

In some of my userscript's I've used a helper method to ensure my code runs at the correct time.

```js
const waitForElem = (selector) => {
    return new Promise(resolve => {
        const interval = setInterval(() => {
            const elem = document.querySelector(selector);
            if (!elem) return;

            clearInterval(interval);
            resolve(elem);
        });
    });
}
```

and then

```js
const likeBtn = await waitForElem('#like-button-renderer');
.observe(likeBtn, ...)
```

src/plugins/skip-disliked-songs/index.ts
```diff
@@ -0,0 +1,22 @@
+import { t } from '@/i18n';
+import { createPlugin } from '@/utils';
+
+export default createPlugin({
+    name: () => t('plugins.skip-disliked-songs.name'),
+    description: () => t('plugins.skip-disliked-songs.description'),
+    restartNeeded: false,
+    renderer: {
+        observer: MutationObserver,
```
JellyBrick:
```suggestion
        observer: null as MutationObserver | null,
```

src/plugins/skip-disliked-songs/index.ts
```diff
@@ -0,0 +1,22 @@
...
+        async start() {
+            this.observer = new MutationObserver(() => {
+                if (document.querySelector("#like-button-renderer")?.getAttribute("like-status") == "DISLIKE")
+                    document.querySelector("tp-yt-paper-icon-button.next-button")?.click();
+            });
+            this.observer.observe(document.querySelector("#like-button-renderer"),
+                { attributes: true, childList: false, subtree: false });
+        },
+        stop() {
+            this.observer.disconnect();
```
JellyBrick:
```suggestion
            this.observer?.disconnect();
```

JellyBrick (approved):

src/plugins/skip-disliked-songs/index.ts
```diff
@@ -0,0 +1,34 @@
...
+                    const elem = document.querySelector(selector);
+                    if (!elem) return;
+        
+                    clearInterval(interval);
+                    resolve(elem);
+                });
+            });
+        }
+    }
+});
```
JellyBrick:
```suggestion
import { t } from '@/i18n';
import { createPlugin } from '@/utils';

export default createPlugin({
  name: () => t('plugins.skip-disliked-songs.name'),
  description: () => t('plugins.skip-disliked-songs.description'),
  restartNeeded: false,
  renderer: {
    observer: null as MutationObserver | null,
    start() {
      this.waitForElem('#like-button-renderer').then((likeBtn) => {
        this.observer = new MutationObserver(() => {
          if (likeBtn?.getAttribute('like-status') === 'DISLIKE') {
            document
              .querySelector('tp-yt-paper-icon-button.next-button')
              ?.click();
          }
        });
        this.observer.observe(likeBtn, {
          attributes: true,
          childList: false,
          subtree: false,
        });
      });
    },
    stop() {
      this.observer?.disconnect();
    },
    waitForElem(selector) {
      return new Promise((resolve) => {
        const interval = setInterval(() => {
          const elem = document.querySelector(selector);
          if (!elem) return;

          clearInterval(interval);
          resolve(elem);
        });
      });
    },
  },
});
```

- reformatting (2-space indent)
- use `Promise.then` to prevent blocking plugin loader
- use `===` instead of `==`

JellyBrick (changes requested):

src/plugins/skip-disliked-songs/index.ts
```diff
@@ -0,0 +1,41 @@
+import { t } from "@/i18n";
```
JellyBrick: use `single quotes` instead of `double quotes`

JellyBrick:
Other than that, LGTM

ArjixWasTaken:
LGTM as well

ArjixWasTaken (approved):

JellyBrick (approved):

PR #1507, opened 2023-12-09
Enable/Disable Navigation without restart

MiepHD:


src/plugins/navigation/index.ts
```diff
@@ -10,18 +10,33 @@ import backHTML from './templates/back.html?raw';
...
+      const backButton = ElementFromHtml(backHTML);
+      this.waitForElem('#right-content').then((menu: HTMLElement) => {
+        menu.prepend(backButton, forwardButton);
+      });
+    },
+    stop() {
+      document.querySelector('[tab-id=FEmusic_back]')?.remove();
+      document.querySelector('[tab-id=FEmusic_next]')?.remove();
+    },
+    waitForElem(selector: string): Promise<Element> {
```
JellyBrick: `waitForElem` is not needed - the plugin for the renderer is loaded after a DOM is renderered, so `#right-content` is always present.

JellyBrick (changes requested):

JellyBrick (approved):

PR #1515, opened 2023-12-10
feat: Add new plugin `Album actions`

### LINKED ISSUE #831: [Feature Request] Like all songs in an Album

MiepHD:
Adds Undislike, Dislike, Like and Unlike buttons to apply this to all songs in a playlist or album.
image
Fixes #831


ArjixWasTaken:
Don't add `enabled: true` in the config....
Your plugin must not be enabled unless specified otherwise by the user's config.

MiepHD:
I enabled it by default because it just adds these buttons to the UI. If you don't click the buttons nothing happens. But I can remove that if that's a problem

ArjixWasTaken:
Since it does not fix a bug, I consider it an additional feature.
I guess I'll leave it up to @JellyBrick

JellyBrick:
image

The shapes of the buttons can be confusing to users.

MiepHD:
Why? Should I add an explanation what they do?

JellyBrick:
What about showing just one of each button? (1 dislike button, 1 like button)
- If any item is "Liked", show a filled "Like" button; if any item is "Disliked", show a filled "Dislike" button
- If both "Liked" and "Disliked" are checked, show both filled buttons

MiepHD:
Now it shows a 
- undislike button if any item is disliked
- dislike button if any item isn't disliked
- like button if any item isn't liked
- unlike button if any item is liked

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  name: () => t('plugins.album-actions.name'),
+  description: () => t('plugins.album-actions.description'),
+  restartNeeded: false,
+  config: {
+    enabled: true,
+  },
+  renderer: {
+    observer: null as MutationObserver | null,
+    loadobserver: null as MutationObserver | null,
+    changeobserver: null as MutationObserver | null,
```
JellyBrick: use camelCase

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,165 @@
...
+      this.waitForElem('#continuations').then(() => {
+        const buttons: Array<HTMLElement> = [
+          ElementFromHtml(undislikeHTML),
+          ElementFromHtml(dislikeHTML),
+          ElementFromHtml(likeHTML),
+          ElementFromHtml(unlikeHTML),
+        ];
+        const playlist = document.querySelector('ytmusic-shelf-renderer')
+          ? document.querySelector('ytmusic-shelf-renderer')
+          : document.querySelector('ytmusic-playlist-shelf-renderer');
```
JellyBrick:
```suggestion
        const playlist = document.querySelector('ytmusic-shelf-renderer') ?? document.querySelector('ytmusic-playlist-shelf-renderer');
```

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,165 @@
...
+      loader?.style.setProperty('left', '50%');
+      loader?.style.setProperty('position', 'absolute');
+    },
+    applyToList(id: string, loader: HTMLElement) {
+      if (loader.children.length != 0) return;
+      this.loadobserver?.disconnect();
+      let playlistbuttons: NodeListOf<Element> | undefined;
+      const playlist = document.querySelector('ytmusic-shelf-renderer')
+        ? document.querySelector('ytmusic-shelf-renderer')
+        : document.querySelector('ytmusic-playlist-shelf-renderer');
```
JellyBrick, line 133: ditto

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,165 @@
...
+            '#button-shape-like[aria-pressed=true] > button',
+          );
+          break;
+        default:
+      }
+      if (playlistbuttons) {
+        for (const elem of playlistbuttons) {
+          elem.click();
+        }
+      }
```
JellyBrick:
```suggestion
      playlistButtons?.forEach((elem) => elem.click());
```

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,165 @@
...
+            playlist?.querySelectorAll(
+              '#button-shape-like[aria-pressed=false] > button',
+            ).length,
+            playlist?.querySelectorAll(
+              '#button-shape-like[aria-pressed=true] > button',
+            ).length,
+          ];
+          let i = 0;
+          for (const count of counts) {
+            console.log(count);
```
JellyBrick:
```suggestion
```

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,165 @@
...
+            if (count == 0) {
+              buttons.splice(i, 1);
+              i--;
+            }
+            i++;
+          }
+        }
```
JellyBrick: How about `.filter()` instead of this logic?

JellyBrick (changes requested):

inson1:
> Now it shows a
> 
> * undislike button if any item is disliked
> * dislike button if any item isn't disliked
> * like button if any item isn't liked
> * unlike button if any item is liked

Just another opinion. I hope you dont mind it :) That is standard, yes. But it isnt perfect :/. 

What about highlighted (bold lines?) like/dislike button if some are liked/disliked, but not all? (I think that is also used somewhere) Or something like that? Maybe like/dislike button with full white, but black "ball" inside
Just something so normal user would get it and "profi" user would know if all songs are liked or disliked.
Or another mode for this?

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    waitForElem(selector: string) {
+      return new Promise((resolve) => {
+        const interval = setInterval(() => {
+          const elem = document.querySelector(selector);
+          if (!elem) return;
+
+          clearInterval(interval);
+          resolve(elem);
+        });
+      });
```
Su-Yong, line 175:
```suggestion
    waitForElem(selector: string) {
      let limit = 0;
      return new Promise((resolve, reject) => {
        const interval = setInterval(() => {
          limit += 1;
          if (limit > 3000) {
            clearInterval(interval);
            reject();
          }
          const elem = document.querySelector(selector);
          if (!elem) return;

          clearInterval(interval);
          resolve(elem);
        });
      });
```

src/plugins/album-actions/index.ts
```diff
@@ -0,0 +1,171 @@
...
+          }
+        }
+        const menu = document.querySelector('.detail-page-menu');
+        if (menu && !document.querySelector('.like-menu')) {
+          for (const button of buttons) {
+            menu.appendChild(button);
+            button.addEventListener('click', this.loadFullList);
+          }
+        }
+      });
```
Su-Yong, line 108:
```suggestion
      }).catch(() => {
        // Cannot find element
      });
```

Su-Yong (commented):
It seems memory leak bug

Su-Yong (approved):

JellyBrick (approved):

PR #1667, opened 2024-01-22
Improve video title filters

Serial-ATA:
This adds more filters for `cleanupName` that I have encountered in the wild.

Notably: `(video version)`, `[HQ]`, `(live)`, `[4K]`

### DIFF
--- a/src/providers/song-info.ts
+++ b/src/providers/song-info.ts
@@ -150,10 +150,19 @@ const registerProvider = (win: BrowserWindow) => {
 };
 
 const suffixesToRemove = [
-  ' - topic',
-  'vevo',
-  ' (performance video)',
-  ' (clip official)',
+  // Artist names
+  /\s*(- topic)$/i,
+  /\s*vevo$/i,
+
+  // Video titles
+  /\s*[(|\[]official(.*?)[)|\]]/i, // (Official Music Video), [Official Visualizer], etc...
+  /\s*[(|\[]((lyrics?|visualizer|audio)\s*(video)?)[)|\]]/i,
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Instead of matching all the surrounding whitespace (well, at least the whitespace on the left) in every regex pattern, wouldn't it be better to not match it and after the replacements to trim + remove duplicate whitespaces?

Like

```js
let title = "Rap God [live]"
title = title.replace("[live]", "") // on phone, not gonna write regex

title = title.replace(/\s+/g, ' ').trim()
```

?

Serial-ATA:
I don't think it makes that much of a difference (?), but since multiple patterns check the end of the string, you'd have to trim the string after each match anyway.

ArjixWasTaken:
> I don't think it makes that much of a difference (?), but since multiple patterns check the end of the string, you'd have to trim the string after each match anyway.

Sounds like an issue with the patterns to me

Serial-ATA:
In the context of a microbench I can't get better performance out of a whitespace replacement + trim at the end, but that could very well be my mistake.

If you can make a benchmark between the two solutions, please do share it. Regardless though, this is only done when the song changes (likely every few minutes). So, it's not very performance critical in the first place.

ArjixWasTaken:
True

Serial-ATA:
Hey @JellyBrick, any update on this? This would be really helpful with the scrobbler plugin. I have to make manual edits right now with a bunch of junk suffixes falling through.

JellyBrick (approved):

PR #1769, opened 2024-02-20
[MPRIS] Issue Fixes and additional properties support

Totto16:
Depends on #1760

This fixes some issue with MPRIS, some of them were less used features

- add fullscreen support, meaning it can be set, and retrieved correctly, the old fullscreen set method in the code was not working anymore, so I fixed it
- fix a wrong (reversed) boolean condition in pause event of MPRIS 
- make some things more robust in mpris by explicitly checking if we support this (the player.can* values might change at runtime now, e.g. 'CanSetFullscreen' might change on non fullscreenable things)
- type some values stricter, instead of string list all options explicitly
- type MPRIS EventEmitter events explicitly
- added better support for  'player.canGoPrevious' and 'player.canGoNext' by extracting playlist metadata and setting those values appropriately
- fix IPC errors in some places
- change the volume updater, in non-precise plugin activated state, since the player supports now precise setting of the volume

I tested everything locally by sending d-bus requests for every property and checking if it is correct, interacting with the UI and checking if the MPRIS values where as intended. So far everything seems to be correct, but test it locally and give me feedback about it ❤️ 

P.S.: SInce I parse the playlist, I also thought and read about supporting the [org.mpris.MediaPlayer2.TrackList](http://specifications.freedesktop.org/mpris-spec/latest/Track_List_Interface.html) interface, but for that we need some requirements, like unique ID's for tracks, but the playlists get parsed by html only, and there might not be a videoID (we use that for the track id) present. You could parse it from the thumbnail, but they are not correct or even an url in some cases.

Here are some screenshots (sorry for some german parts 😓 ) as how it should work and works now:

image
image

At the beginning of the playlist

image
image

At the end of the playlist, no autoplay or Playlist loop enabled

image
image

At the end of the playlist, autoplay enabled


JellyBrick:
There is a way to get the queue via the API without using an element selector.
See #1562, https://github.com/th-ch/youtube-music/blob/master/src/plugins/music-together/queue/queue.ts

JellyBrick:
Could you check that it's working properly?

JellyBrick:
image

I think it works

Totto16:
@JellyBrick Thank you for continuing this and correcting my mistakes ❤️  I was new to this repo, so I didn't know the functionality, that was already present that well 😓

PR #1964, opened 2024-04-11
feat(downloader): New option to download on finish

Johannes7k75:
This adds an option to the downloader menu. 
When the option is activated it tracks the time that is elapsed and then if the rest time of the song is below 20 seconds it downloads the song.

The range is 20 seconds for when you skip the last bit of the song.


src/plugins/downloader/main/index.ts
```diff
@@ -155,20 +158,54 @@ export async function downloadSongFromId(
...
+  });
+}
+
 async function downloadSongUnsafe(
   isId: boolean,
   idOrUrl: string,
   setName: (name: string) => void,
   playlistFolder: string | undefined = undefined,
   trackId: string | undefined = undefined,
-  increasePlaylistProgress: (value: number) => void = () => {},
```
JellyBrick: What are these changes?

src/plugins/downloader/main/index.ts
```diff
@@ -155,20 +158,54 @@ export async function downloadSongFromId(
...
+function downloadSongOnFinishSetup({
+  ipc,
+  getConfig,
+}: Pick<BackendContext<DownloaderPluginConfig>, 'ipc' | 'getConfig'>) {
+  let currentUrl: string | undefined;
+  let duration: number | undefined;
+  let time = 0;
+
+  registerCallback(async (songInfo: SongInfo) => {
+    const config = await getConfig();
```
JellyBrick: Checking `config` in the callback is not a good idea (IO overhead)
How about calling `registerCallback` if `downloadOnFinish` is true?

JellyBrick (changes requested):

ArjixWasTaken:
What happens on songs with a length of 40 seconds?

20 seconds would be the half, and not the "last bit"

Either make it configurable based on how big a song is, or make it a percentage (like 99%)

Johannes7k75:
> What happens on songs with a length of 40 seconds?
> 
> 20 seconds would be the half, and not the "last bit"
> 
> Either make it configurable based on how big a song is, or make it a percentage (like 99%)

If made it so that you can configure "last x seconds" and "after x percent".
The you have the choice between these to options
image

And here the window where you can configure the two options (if clicked on "Advanced")
image

ArjixWasTaken:
I will test and review this later today.

.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
+  workflow_dispatch:
```
ArjixWasTaken: why?

ArjixWasTaken:
Why does your PR touch things unrelated to the download plugin?

ArjixWasTaken:
Ok, it seems to work fine on my machine.

The only issue that I noticed is kinda unrelated to this plugin, but the song change callbacks are not called for the auto-play song on start-up. (which is totally not the responsibility of this plugin to fix)

Johannes7k75:
> Also, maybe it would be a good idea to allow a separate folder for the auto-downloaded tracks.

I try to add it in the next days

plugins/scroll-down/front.js
JellyBrick, line 1: What is this?

JellyBrick (changes requested):

JellyBrick:
I'm busy with my job, so I can't review it right now.
I'll check this out in a couple days.

Thanks for your contribution!

JellyBrick (approved):

MulverineX:
Does this actually use the cached audio Blob? Or does it effectively download the song twice?

MulverineX:
@Johannes7k75

Johannes7k75:
It just calls the downloadSong function

Johannes7k75:
So i think it downloads the song twice

MulverineX:
would be really awesome if this somehow didn't use that

Johannes7k75:
Would it really benefit from using the cached audio blob?
And beyond that, I don't know enough about it.

Issue #1817, opened 2024-03-02
[Bug]: Ads not being blocked while using the " In-Player " adblocker


XCalibur5678:
### Preflight Checklist

- [X] I use the latest version of YouTube Music (Application).
- [X] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [X] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.3.2

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official YTM web version](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 11 23H2 22631.3155

### What CPU architecture are you using?

x64

### Reproduction steps

1. Switch to Plugins>Adblocker>Blocker>In Player
2. Play any song

### Expected Behavior

Music to play without any ads 

### Actual Behavior

2 30 seconds ads playing

### Enabled plugins

1. Adblocker
2. Album Actions
3. Blur Navigation Bar
4. Downloader
5. In app Menu
6. Lyrics Genius
7. Navigation
8. SponsorBlock
9. Taskbar media control
10. Video Toggle

### Additional Information

If using the In player ad blocker , ads aren't being blocked . Using the " blocklists " function mitigates this issue . Hereby attaching a screenshot of the ad playing . Please feel free to reach out to me if you need any more information, 

![image](https://github.com/th-ch/youtube-music/assets/98412450/26c95c8d-a84c-4612-b5be-1d16ac327c9a)

Joey451-OG:
I can confirm this is happening on Linux as well:

Linux Version:
Ditro: Fedora 39
Architeture: x86_64

Version: 3.3.2

Enabled Plugins:
1. Adblocker (In-Player)
2. Discord Rich Presence
3. Navigation

Please let me know if I can be of further assistance. I love using this tool :D

![image](https://github.com/th-ch/youtube-music/assets/60891047/13e99d30-f881-4995-b343-f0e71a797318)

BrandTruong:
Also have this issue. Swapping between in-player and blocklist does nothing and ads still appear.

OS: Windows 10 Home 
OS Version: 10.0.19045 Build 19045
System Type: x64-based PC

YT Music Version: 3.3.4

Plugins Enabled:
Ad Blocker
Album Color Theme
Discord Rich Presence
Downloader
[...]

PR #1966, opened 2024-04-11
fix: add AdGuard as blocklist sources

Joey451-OG:
Added another blocklist source provided by Trixarian in #1817. This will block ads that many users have been experiencing in the application without requiring them to add the source to the custom blocklist themselves.

If merged, this will issue #1817 should be ready to close.

Thank you all for creating this amazing tool :D

### DIFF
--- a/src/plugins/adblocker/blocker.ts
+++ b/src/plugins/adblocker/blocker.ts
@@ -15,6 +15,8 @@ const SOURCES = [
   'https://raw.githubusercontent.com/uBlockOrigin/uAssets/master/filters/filters-2023.txt',
   // Fanboy Annoyances
   'https://secure.fanboy.co.nz/fanboy-annoyance_ubo.txt',
+  // AdGuard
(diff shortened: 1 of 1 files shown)


src/plugins/adblocker/blocker.ts
```diff
@@ -15,6 +15,8 @@ const SOURCES = [
   'https://raw.githubusercontent.com/uBlockOrigin/uAssets/master/filters/filters-2023.txt',
   // Fanboy Annoyances
   'https://secure.fanboy.co.nz/fanboy-annoyance_ubo.txt',
+  // AdGuard
+  'https://filters.adtidy.org/extension/ublock/filters/122_optimized.txt',
```
JellyBrick, line 19: ~~Is `filters.adtidy.org` a safe and reliable source?~~

- https://www.reddit.com/r/uBlockOrigin/comments/mys0pu/comment/gvwmsd2/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

PR #2033, opened 2024-05-06
Enable arm64 for deb and rpm

ahoneybun:
This updates the README for the new macOS arm64 option as well as adds the option to build arm64 packages for deb and rpm distros.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ latest version.
 Install the [`youtube-music-bin`](https://aur.archlinux.org/packages/youtube-music-bin) package from the AUR. For AUR installation instructions, take a look at
 this [wiki page](https://wiki.archlinux.org/index.php/Arch_User_Repository#Installing_packages).
 
-### MacOS
+### macOS
 
 You can install the app using Homebrew (see the [cask definition](https://github.com/th-ch/homebrew-youtube-music)):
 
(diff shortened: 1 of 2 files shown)


ahoneybun:
Alright so I need to add `no-sandbox` option when launching the binary on my Thinkpad X13 so maybe I did something wrong or it's a bug in Electron.

ArjixWasTaken:
> [!WARNING]
This is not a review or a suggestion, it is just a conversation.

---

I think the `MacOS` vs `macOS` thing is due to `MAC` being an acronym, and `Mac` is not an acronym.
People often use acronyms in lower-case, which makes it harder to discern.

Of course, it is kinda obvious that `mac` is not an acronym, but I could read `macOS` as `Media Access Control Operating System` and you can't say that I am wrong.

Although, since `OS` is an acronym by itself, it also hints that the `mac` part is not an acronym due to the different letter case.
And then there are people that use `macos` and I do not know how to react to that 😂

What is your opinion on this topic?

ahoneybun:
@ArjixWasTaken well I changed it since Apple now has the branding as `macOS` instead of `MacOS`:

https://support.apple.com/macos
https://www.apple.com/macos/sonoma/
https://en.wikipedia.org/wiki/MacOS

PR #2052, opened 2024-05-13
Set activity to "Listening" instead of "Playing"

dvd-22:
It is a simple change but makes it 100x better when seeing my status on Discord. This is indeed possible as seen in [the code for the fork of discord.js used in this repository](https://github.com/xhayper/discord-rpc/blob/6683ce4abb75867ddde1b8ce31657de564f5ed0a/src/structures/ClientUser.ts#L75)

### DIFF
--- a/src/plugins/discord/main.ts
+++ b/src/plugins/discord/main.ts
@@ -180,12 +180,12 @@ export const backend = createBackend<
     }
 
     const activityInfo: SetActivity = {
-      type: ActivityType.Listening,
       details: songInfo.title,
       state: songInfo.artist,
       largeImageKey: songInfo.imageSrc ?? '',
       largeImageText: songInfo.album ?? '',
       buttons,
+      type: ActivityType.Listening,
     };
 
     if (songInfo.isPaused) {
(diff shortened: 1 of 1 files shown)


Randrage:
Yes please! 👍

WillCarlotto:
Yes! "Listening" should be the default

myurtsev:
just merge it :)

myurtsev (approved):

MulverineX:
@dvd-22 did you build the app locally to test this actually works? Last I checked custom activity types were not allowed by Discord by default

dvd-22:
> @dvd-22 did you build the app locally to test if this actually works? Last I checked custom activity types were not allowed by Discord by default.

No, I didn't. I used discord.js v12 some time ago and it was allowed, so I just assumed it worked in newer versions. I will test it and update the pr accordingly.

MulverineX:
Make sure when you test a friend can see the "Listening to" state as well. It might appear to work client-side but not actually be working.

dvd-22:
I tested this, and sadly it doesn't work. It looks like activites set through IPC (not oauth activity.write scope) cannot set anything besides playing. Listening and Watching are overwritten by Discord to playing and the rest are blocked. So unless Discord changes something in their API, it won't be possible anytime soon.

PR #1768, opened 2024-02-19
plugin: Refined Lyrics

ArjixWasTaken:
⚠️ THIS IS A DRAFT ⚠️ 

Made a draft PR, so I can easily work on this from multiple computers.
This plugin will be a replacement for lyrics-genius, but it will support more sources, synced lyrics, romanization, etc.

Roadmap:
- [x] Intercept the lyrics tab header and content.
- - [x] Always enable the lyrics tab
- - [ ] Read/Write lyrics
- [ ] Reach feature parity with lyrics-genius.

*More goals will be added at a later time.

PR #2207, opened 2024-07-08
plugin: Synced Lyrics

Non0reo:
Hi 👋

I made this plugin to make lyrics synchronous with the music.
I used [LRCLIB](https://lrclib.net/) to get lyrics as it's also another open-source project with an easy-to-use API.

I took code from the lyrics-genius to have a base for my plugin. I modified a part of it so it is compatible with new version of Youtube Music (while adding functionalities do fetch, change, and display lyrics)


ArjixWasTaken:
:heart:

src/plugins/synced-lyrics/index.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+
+export type LineEffect = 'scale' | 'offset' | 'focus';
+
+export default createPlugin({
+  name: () => 'Synced Lyrics',
+  description: () => 'Synced Lyrics Plugin Description',
+  authors: ['Non0reo'],
+  restartNeeded: true,
```
Su-Yong, line 15:
```suggestion
  restartNeeded: true,
  addedVersion: '3.4.X',
```

Su-Yong (approved):
I recommend the plugin works with `restartNeeded: false`. but it's okay, I'm not forcing that. anyway, It seems to be great.

+need to be write `addedVersion` field. please accept my suggestion

Non0reo:
Hi, thanks for your feedbacks

Yeah I have no problems merging your work here 👍
I also planed to add a feature to get plain lyrics from genius when the song does not have any lyrics (and making essentially what lyrics-genius did). Is it pertinent to implement it ? And have you worked on something similar, where in this case I should not bother to implement that ?

JellyBrick:
You can get lyrics from YT using `youtubei.js`

src/plugins/synced-lyrics/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  addedVersion: '3.4.X',
+  config: {
+    enabled: true,
```
JellyBrick:
```suggestion
    enabled: false,
```

src/plugins/synced-lyrics/menu.ts
```diff
@@ -0,0 +1,130 @@
+import { MenuContext } from "@/types/contexts";
+import { MenuItemConstructorOptions } from "electron";
+import { SyncedLyricsPluginConfig } from ".";
+
+export const menuContent = async ({ getConfig, setConfig }: MenuContext<SyncedLyricsPluginConfig>): Promise<MenuItemConstructorOptions[]> => {
+	const config = await getConfig();
+
+	return [
+		{
```
JellyBrick: This violates the code style being used in this project. (check eslint)

JellyBrick (changes requested):

Non0reo:
> You can get lyrics from YT using `youtubei.js`

My plan was actually to have an additional source of lyrics and not only lyrics from YouTube. But it seems to be pretty annoying to scrape genius.com, so I don't think that I will try to implement this feature (unless I find a better way to scrape the html from the site)

src/plugins/synced-lyrics/renderer/lyrics/progress.ts
```diff
@@ -16,20 +17,21 @@ export const resetAllVariables = () => {
...
     if (config.preciseTiming) {
-        currentTime = secToMilisec(t);
-        clearInterval(interval!);
-        interval = setInterval(() => {
-          currentTime += 10;
-          changeActualLyric(currentTime);
-        }, 10);
-    } 
+      // MAGIC: Since the transition takes 300ms, we need to add a delay of 300ms to the current time
+      currentTime = secToMilisec(t) + 300;
```
Non0reo: Also, since this change, the lyrics seems to be displayed too early now. But if the way it worked before would have been a source of error we can keep it like that

chiqors:
is this PR ready to test?

ArjixWasTaken:
Personally, I find it better to have the lines text-aligned to the center

image

Although, that is something that can be achieved by using a css theme (if we still support that)

Non0reo:
It's actually centered by default, but I wanted to mimic this spotify aesthetic that I found really great ! (don't ask me why)

image

I don't know if adding another parameter to let the user style this part of the lyrics is a good idea or not but anyway, I don't see any problem in centering the text if you want 👍

Non0reo:
@chiqors I theory this plugin is functional

ArjixWasTaken:
@Non0reo

I am doing some small refactoring, is it ok if I push my changes once done?
Or should I make a PR to your fork?

ArjixWasTaken:
In case you did not already do this, I'd suggest you use eslint and prettier, that way the code not only follows our formatting standards, it also is easier to review in pull requests.

ArjixWasTaken:
LGTM

Additions like more lyric providers can be done in a separate PR.

So I think we should iron out:

- translatable strings
- maybe make more things configurable?
- when in combination with album theme plugin, the lyrics are really hard to read (when an ad is playing for example)

Are there more stuff left to iron out?

Non0reo:
Great for your contribution (you made me realized that my code was not that well written 😅)
For me the plugin seems great as it is, and in addition in what you already said I think we should investigate the followings:

- Add other lyrics provider (and maybe add plain lyrics provider as I mentioned [here](https://github.com/th-ch/youtube-music/pull/2207#issuecomment-2227331079))
- When clicking on the Discover tab and then clicking on the Lyrics tab, the mutation isn't detected and the lyrics don't update correctly
- Clicking on shuffle playlist button does not trigger the `ytmd:update-song-info` and thus does not fetch the lyrics. I will create an issue for this one

Appart from that, everything is okay

ArjixWasTaken:
> When clicking on the Discover tab and then clicking on the Lyrics tab, the mutation isn't detected and the lyrics don't update correctly

In my now closed PR for `refined-lyrics` I made it always enable the lyrics tab, and it always called my code to replace the html within.

https://github.com/ArjixWasTaken/youtube-music/blob/refined-lyrics/src/plugins/refined-lyrics/contexts/renderer.ts

Maybe you can ~~steal~~ use that?
Although the way you and I implemented these, is way too different ngl.

Non0reo:
Oh cleaver ! I'm not really familiar with SolidJS, and also because you already fixed this issue I think that your PR would be better than my code

ArjixWasTaken:
looks like I pushed to the wrong origin...

ArjixWasTaken:
@Non0reo I finally migrated the UI code to use SolidJS, there are some kinks I need to iron out though, like the text size is too small, etc.

ArjixWasTaken:
Btw we can get synced lyrics from youtube music directly

[Here](https://github.com/th-ch/youtube-music/pull/1768/files#diff-df6b8031bbfcab860036c75ad120dca1acbcf18b1e91da0d532fbbc3508b3f34) is working code that does that.

I plan to include that in another PR, once this is merged.

chiqors:
I've tested the app.. it works great, although i know. There are music which doesn't have sync lyric, the source coming from lyricFind. Do you guys have an idea to put yt video subtitles into lyric tab (which we may have an option to switch language to use auto sub, translated auto sub, user manual inputted language?

MulverineX:
@chiqors autogenerated subtitle based lyrics are on the devs' radar afaik

ArjixWasTaken (approved):

ArjixWasTaken:
Other than fixing the styling, I don't think there is much to change for an initial version.

ArjixWasTaken:
I'll re-review this tomorrow and try to get it merged.

Su-Yong:
I found a critical bug. When `render` function is called more than twice (where `tabStates.true`), singals are not synced. I'll be fixing

JellyBrick (approved):

JellyBrick:
need i18n

ArjixWasTaken:
Added i18n, well, for english at least.

src/plugins/synced-lyrics/renderer/components/LyricsContainer.tsx
```diff
@@ -48,29 +57,43 @@ export const LyricsContainer = () => {
 
       <Switch>
         <Match when={!lineLyrics().length}>
-          <yt-formatted-string
-            class="warning-lyrics description ytmusic-description-shelf-renderer"
-            text={{
-              runs: [
-                {
-                  text: 'No lyrics found for this song.',
+          <Show when={isInstrumental()}>
```
JellyBrick: Use `fallback` instead of duplicate `Show`

JellyBrick (changes requested):

ArjixWasTaken:
@Non0reo you messed up in that last commit of yours

src/plugins/synced-lyrics/renderer/components/LyricsContainer.tsx
ArjixWasTaken: I did that originally, but I decided the code was too ugly, so I didn't keep it.
Is it really necessary?

Non0reo:
I don't know if you're working on it but [the text size seems too small](https://github.com/user-attachments/assets/85a80179-3e09-4f78-9b9c-ddce4f0c7469). Maybe I did something wrong with my version or something like that though

Also, when the app start for the first time the lyrics [arent fetched properly](https://github.com/user-attachments/assets/0a0e7472-3ecf-4fc5-b9e8-be45cc783b80)

ArjixWasTaken:
> I don't know if you're working on it but [the text size seems too small](https://github.com/user-attachments/assets/85a80179-3e09-4f78-9b9c-ddce4f0c7469)

The css may need to be updated

> Also, when the app start for the first time the lyrics [arent fetched properly](https://github.com/user-attachments/assets/0a0e7472-3ecf-4fc5-b9e8-be45cc783b80)

That is a known bug that I am trying to fix, although your screenshot is different from what I was experiencing

ArjixWasTaken:
@JellyBrick any more thoughts?

JellyBrick (approved):

JellyBrick:
I have a question. What happens if I run lyric-genius and synced lyrics at the same time?

ArjixWasTaken:
> I have a question. What happens if I run lyric-genius and synced lyrics at the same time?

is lyrics-genius even working?
either way, I think it would be better if we removed lyrics genius once synced-lyrics has multiple lyric providers

I'll handle that on my upcoming PR

sH1222J:
As a user who enjoys listening to K-pop and J-pop, there are almost no applied lyrics.

JellyBrick:
@sH1222J https://lrclib.net/

heyluft:
The lyrics for the song are on lrclib, but it doesnt work with the app, it just says 'No lyrics found for this song' 'Refetch lyrics'.

ArjixWasTaken:
@heyluft this is not the place to hold such discussion, please either open an issue, or comment on an existing issue.

Rairof:
> > I have a question. What happens if I run lyric-genius and synced lyrics at the same time?
> 
> is lyrics-genius even working? either way, I think it would be better if we removed lyrics genius once synced-lyrics has multiple lyric providers
> 
> I'll handle that on my upcoming PR

Would be great if user could just select the lyric provider on their own through in plugin settings whenever multiple lyrics source is implemented <3

ArjixWasTaken:
> Would be great if users could just select the lyric provider manually as well through in plugin settings but by default it should probably be auto selected to whatever source is best whenever multiple lyrics source is implemented <3

I'd imagine a priority order would be best.

Rairof:
> I'd imagine a priority order would be best.

Affirmative sir.

Issue #1641, opened 2024-01-13
[Bug]: Plugin 'Ambiant Mode'


KLalamex59:
### Preflight Checklist

- [X] I use the latest version of YouTube Music (Application).
- [X] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [X] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.2.2.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official YTM web version](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 10 22H2

### What arch are you using?

x64

### Reproduction steps

1. Enable the plugin '**Ambiant Mode**'.
2. Play a song (_working_).
3. Play a video (_not working, need to disable/re-enable the plugin to make it work_).

### Expected Behavior

Reloading the plugin whenever links change might potentially resolve this issue, although I don't consider it an optimal optimization technique.

### Actual Behavior

When the '**Ambient Mode**' plugin is activated, it doesn't funtion properly _when playing music with a video clip_;
it only works for songs without videos.

### Enabled plugins

1. Adblocker (In Player)
2. Ambient Mode
3. Blur Navigation Bar
4. Captions Selector
5. Compact Sidebar
6. Discord Rich Presence
7. Downloader
8. In-App-Menu
9. Lyrics Genius
10. Music Together [Beta]
11. Navigation
12. Picture-In-Picture
13. Playback Speed
14. Precise Volume
15. Skip Disliked Songs
16. Taskbar Media Control
17. Video Quality Changer
18. Video Toggle

![image](https://github.com/th-ch/youtube-music/assets/125516170/79635f56-ac42-4f55-acf8-0b4e7bf123a5)

JellyBrick:
![YouTube_Music_5ZKgoRU6Cb](https://github.com/th-ch/youtube-music/assets/16558115/d33db76d-9cc3-40fd-a62a-3a0ae1feda3c)

It seems works

KLalamex59:
😴😴😴😴
![image](https://github.com/th-ch/youtube-music/assets/125516170/b9b0d954-c92a-46d1-9fee-f73e6f1a79fd)

inson1:
@KLalamex59 did you restart the app?

Hadeik:
[...]

PR #2255, opened 2024-07-18
Fix Ambient Mode not working for videos after restart

craftgeil:
This should fix https://github.com/th-ch/youtube-music/issues/1641

### DIFF
--- a/src/plugins/ambient-mode/index.ts
+++ b/src/plugins/ambient-mode/index.ts
@@ -53,7 +53,18 @@ export default createPlugin({
       const songImage = document.querySelector<HTMLImageElement>('#song-image');
       const songVideo = document.querySelector<HTMLDivElement>('#song-video');
       const image = songImage?.querySelector<HTMLImageElement>('yt-img-shadow > img');
-      const video = songVideo?.querySelector<HTMLVideoElement>('.html5-video-container > video');
+      const video = await waitForVideo();
+
+      async function waitForVideo() {
+        let video = null;
+
+        while (video == null) {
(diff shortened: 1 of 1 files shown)


src/plugins/ambient-mode/index.ts
```diff
@@ -53,7 +53,18 @@ export default createPlugin({
       const songImage = document.querySelector<HTMLImageElement>('#song-image');
       const songVideo = document.querySelector<HTMLDivElement>('#song-video');
       const image = songImage?.querySelector<HTMLImageElement>('yt-img-shadow > img');
-      const video = songVideo?.querySelector<HTMLVideoElement>('.html5-video-container > video');
+      const video = await waitForVideo();
+
+      async function waitForVideo() {
```
JellyBrick, line 58: see https://github.com/th-ch/youtube-music/blob/ebd167f3f2d39302b08fb9d3477ed173890a3f61/src/plugins/album-actions/index.ts#L186

JellyBrick (changes requested):

src/plugins/ambient-mode/index.ts
craftguyL: Hello,

I tried to comply with your request, however I am not able to edit the code to your wishes, due to my lack of knowledge and experience. My current version involved a lot of experimenting as well as asking ChatGPT for help. 

So if anyone else would be able to take a look and edit it accordingly, that would be great. If not, please close this request and I will continue to build the app myself with my suboptimal fix.

PR #2258, opened 2024-07-18
Fix: Incorrect window size on scaled displays

Alawnely:
This should fix th-ch#1716

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -321,10 +321,9 @@ async function createMainWindow() {
     const { x: windowX, y: windowY } = windowPosition;
     const winSize = win.getSize();
     const display = screen.getDisplayNearestPoint(windowPosition);
-    const scaleFactor = is.windows() ? display.scaleFactor: 1;
 
-    const scaledWidth = Math.floor(windowSize.width / scaleFactor);
-    const scaledHeight = Math.floor(windowSize.height / scaleFactor);
+    const scaledWidth = windowSize.width;
+    const scaledHeight = windowSize.height;
 
     const scaledX = windowX;
     const scaledY = windowY;


src/index.ts
```diff
@@ -321,10 +321,9 @@ async function createMainWindow() {
     const { x: windowX, y: windowY } = windowPosition;
     const winSize = win.getSize();
     const display = screen.getDisplayNearestPoint(windowPosition);
-    const scaleFactor = is.windows() ? display.scaleFactor: 1;
```
JellyBrick, line 324: Please explain why this change was necessary. Previously, this logic was required due to incomplete HiDPi support in Windows OS.

Alawnely: Device: Windows 11 23H2 , 1920x1080 at 125% display scaling

In the current stable version, when the YTM app is opened, the window size is based on the values stored in `config.json`.  This size is actually pixel perfect on the display, that is, if the stored width and height are 640x960, the YTM window is actually 640x960 on the display.

Now, when I resize the YTM window, the window size stored to the config file is scaled (for example, 640x960 becomes 512x768 on 125% display scaling).  This includes maximizing and restoring the window.  But because in the current version, the app compensates for the display scaling, reopening the app now would have a window size of what was stored in the config (the window is actually 512x768 on my display, not the 640x960 that I set when manually resizing the window).  Thus the YTM window is smaller than expected.

After removing that line, what's stored in the config remains scaled (in my example, it's still 512x768).  But because the app no longer compensates for display scaling, when I reopen the app, that stored value is multiplied to the display scaling, thus the window size I set is the same as when I resized it a while ago.

I used this approach of fixing because it is consistent with how Chrome handles their window sizing (I think).

What I noticed is this only happens with the default high DPI settings.  When I changed the high DPI override to "System" or "System (Enhanced)", the bug does not occur.

Not sure for older Windows versions.

**TL;DR**
Before code fix:
1. I resize YTM window to some value (for example, 640x960)
2. I close the app
3. I reopen the app
4. The YTM window is smaller (512x768)

After code fix:
1. I resize YTM window to some value
2. I close the app
3. I reopen the app
4. The YTM window size is the same

See video.

https://github.com/user-attachments/assets/d62f03a4-03a2-4a93-bca8-994f84af44c8

JellyBrick (approved):

Sickjuicy:
I have now problems with my second screen, it has 150% zoom

Alawnely:
The YTM window is big, or the page is zoomed in?

Alawnely:
From my testing, after update install, the window size is bigger, due to how window size was saved.  But reopening the app no longer has the bug

Sickjuicy:
The YTM is big, after every restart, I have 2 Screens one with 100% zoom (Normal) and a second one with 150% with YTM open.

Sickjuicy:
I will look for a better method this weekend so that the windows zoom works properly with every screen configuration, when I have a first version I will let you know here again if someone wants to test it :)

PR #2265, opened 2024-07-21
Add rich presence type

bqcrabtree:
Add rich presence type so that discord reports listening rather than playing

Old behavior: Discord reports "Playing a game"
New behavior: Discord will report "Listening to"

### DIFF
--- a/src/plugins/discord/main.ts
+++ b/src/plugins/discord/main.ts
@@ -180,6 +180,7 @@ export const backend = createBackend<
     }
 
     const activityInfo: SetActivity = {
+      type: ActivityType.Listening,
       details: songInfo.title,
       state: songInfo.artist,
       largeImageKey: songInfo.imageSrc ?? '',


ArjixWasTaken:
We already had a PR that did the exact thing, but it turned out that discord ignored the type and still displayed "Playing" instead of "Listening"

https://github.com/th-ch/youtube-music/pull/2052

Have you tested if this actually works?

bqcrabtree:
Eek! I missed that, actually! Let me give it a test and see if it works. Funny enough, I just read through and saw that they were using discord.js when they realized they could do this too.

ArjixWasTaken:
you are also missing an import

ArjixWasTaken:
Just tested it, and it still shows up as "Playing"

ArjixWasTaken:
One possible solution would be to not use discord's IPC, but instead directly communicate with their gateway socket to send presence updates.

But even then, I am not sure if they allow random applications to have the "Listening" presence

ArjixWasTaken:
A friend of mine has implemented this in his android app.

https://github.com/brahmkshatriya/echo-discord/tree/main/app/src/main/java/dev/brahmkshatriya/echo/extension

The app is still in early development so I can't confirm if the Listening presence works, I am waiting for his reply on that matter.

ArjixWasTaken:
> A friend of mine has implemented this in his android app.
> 
> [brahmkshatriya/echo-discord@`main`/app/src/main/java/dev/brahmkshatriya/echo/extension](https://github.com/brahmkshatriya/echo-discord/tree/main/app/src/main/java/dev/brahmkshatriya/echo/extension?rgh-link-date=2024-07-21T21%3A29%3A51Z)
> 
> ~The app is still in early development so I can't confirm if the Listening presence works, I am waiting for his reply on that matter.~
> 
> He said that it indeed works.

image

I got confirmation that someone may get inspired by his code to port this in typescript.

bqcrabtree:
Was randomly doing some research just now, and found this: https://github.com/discord/discord-api-docs/discussions/4690#discussioncomment-9280595 which seems to show that they've silently implemented the feature in the API. It seems to be _not quite right_ though, and some things aren't working. Maybe later. Sorry for the trouble and thank you for correcting me. :-)

ArjixWasTaken:
The hotline code I sent works just fine.
Only problem is that the user has to login with their discord account

PR #2303, opened 2024-08-01
fix: Window gets stuck offscreen in some instances

Alawnely:
Improved offscreen detection logic, fixes th-ch#1894


JellyBrick (approved):

Araxeus:
Can you explain this change? it seems to disallow saving the window if its more than 1/2 of it is offscreen?

Issue #2363, opened 2024-08-20
[Bug]: Library Not Accessible from Side Menu


OpenRift412:
### Preflight Checklist

- [X] I use the latest version of YouTube Music (Application).
- [X] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [X] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.5.1.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 11 23H2

### What CPU architecture are you using?

x64

### Reproduction steps

1. Launch the program

### Expected Behavior

The "Library" Menu Item should show below "Explore" on the left-side menu.

### Actual Behavior

The "Library" item will show for a couple seconds on launch, then disappear, only showing "Home" and "Explore".

### Enabled plugins

1. Album Actions
2. Album Color Theme
3. Ambient Mode
4. Blur Nav Bar
5. Bypass Age Restrictions
6. Discord Rich Presence
7. Downloader
8. In-App Menu
9. Lyrics Genius
10. Navigation
11. Skip Disliked Songs
12. Taskbar Media Control
13. Video Quality Changer

Chochigum:
Have the same problem :/ the Library Tab just disappears quickly after launching

ArjixWasTaken:
I suspect that you have this enabled

![image](https://github.com/user-attachments/assets/69404c4c-44d1-416f-bf5e-cf4667f9703f)

That tweak hides the last button in that menu, I suspect that youtube premium is not available in your country, so you don't see a subscribe button there?

(or you have youtube premium?)

ArjixWasTaken:
Can you turn off that option and see if the issue persists?

OpenRift412:
> Can you turn off that option and see if the issue persists?

Okay yeah, that fixes it! Thanks!

PR #2364, opened 2024-08-22
fix: detect the upgrade btn using the icon

ArjixWasTaken:
Fixes: #2363

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -172,10 +172,19 @@ async function onApiLoaded() {
 
   // Remove upgrade button
   if (window.mainConfig.get('options.removeUpgradeButton')) {
+    const itemsSelector = 'ytmusic-guide-section-renderer #items';
+    let selector = 'ytmusic-guide-entry-renderer:last-child';
+
+    const upgradeBtnIcon = document.querySelector<SVGGElement>('iron-iconset-svg[name="yt-sys-icons"] #youtube_music_monochrome');
+    if (upgradeBtnIcon) {
+      const path = upgradeBtnIcon.firstChild as SVGPathElement;
+      const data = path.getAttribute('d')!.substring(0, 15);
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Assuming the name of the icon does not change, this should not break.
But even if it does break, it defaults back to the old way of removing the button.

(unless they decide to use an entirely different icon whilst the current one, `youtube_music_monochrome`, is still around)

ArjixWasTaken:
@OpenRift412 would it be too much to ask if you could test this change?

(assuming you know how to run a dev build)

OpenRift412:
> @OpenRift412 would it be too much to ask if you could test this change?
> 
> ~assuming you know how to run a dev build~ Actually, scratch that, you can just run
> 
> ```js
> const itemsSelector = 'ytmusic-guide-section-renderer #items';
> let selector = 'ytmusic-guide-entry-renderer:last-child';
> 
> const upgradeBtnIcon = document.querySelector('iron-iconset-svg[name="yt-sys-icons"] #youtube_music_monochrome');
> if (upgradeBtnIcon) {
>   const path = upgradeBtnIcon.firstChild;
>   const data = path.getAttribute('d').substring(0, 15);
>   selector = `ytmusic-guide-entry-renderer:has(> tp-yt-paper-item > yt-icon path[d^="${data}"])`;
> }
> 
> const styles = document.createElement('style');
> styles.textContent = `${itemsSelector} ${selector} { display: none; }`;
> document.head.appendChild(styles)
> ```
> 
> in your console after visiting https://music.youtube.com on your browser like this
> 
> image
> 
> If it doesn't remove the `Library` button, then all is fine.

I would need you to send me a compiled dev build.

ArjixWasTaken:
Have you heard of the chrome/Firefox dev tools?

Issue #2365, opened 2024-08-24
[Bug]: Minimize, Maximize, Close buttons do not follow Color Theme


thesaikis:
### Preflight Checklist

- [X] I use the latest version of YouTube Music (Application).
- [X] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [X] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.5.1

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Other Linux

### Operating System Version

Linux arch 6.10.6-arch1-1

### What CPU architecture are you using?

x64

### Reproduction steps

1. Install from [youtube-music-bin](https://aur.archlinux.org/packages/youtube-music-bin) AUR.
2. Enable "In-App Menu" from plugins, which removes the title bar.
3. (Optional) Enable "Album Color Theme" which emphasizes the issue.

### Expected Behavior

The minimize, maximize, and close buttons follow the color scheme of the application.

### Actual Behavior

The minimize, maximize, and close buttons always remain the same color.

### Enabled plugins

1. In-App Menu
2. (Optional) Album Color Theme

### Additional Information

I am using the KDE Plasma desktop environment. Changing the theme of KDE (changing from light to dark theme) affects the buttons, but only the three I've mentioned and nothing else.

Besides simply changing from light to dark in the settings of KDE, I've left any other themes or config files untouched.

In fact, rather than just being the wrong color, the buttons appear to be different entirely from what the application actually shows in the screenshots. Using the "Album Color Theme" plugin isn't necessary to cause the issue, but makes it easier to see the buttons are incorrect.

Here are a few images where the color theme changes but the buttons remain the same:
![image](https://github.com/user-attachments/assets/f5889912-38f6-4a2d-9676-d3a159404767)
![image](https://github.com/user-attachments/assets/1a134162-5920-4531-ba48-e1129a3a221a)
![image](https://github.com/user-attachments/assets/c37cc468-75bb-4f34-8b23-9271dcb098a2)

ArjixWasTaken:
> In fact, rather than just being the wrong color, the buttons appear to be different entirely from what the application actually shows in the screenshots.

I believe the screenshots (of the README) were taken on Windows 10/11
[...]

PR #2366, opened 2024-08-25
fix: hide native-controls on linux when in-app-menu is used

ArjixWasTaken:
Closes: #2365

On linux, for some weird reason, having both `frame: false` and `titleBarStyle` as well as `titleBarOverlay` causes a mess.
This fix not only hides the ugly window controls (default GTK controls iirc), it also allows electron to use the system-wide theme for when the in-app-menu is disabled.

I haven't seen any obvious regression due to the changes I make in this PR, but just to be safe, I am asking anyone that reads this to test this PR locally and provide feedback.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,6 +56,7 @@ import { loadI18n, setLanguage, t } from '@/i18n';
 import ErrorHtmlAsset from '@assets/error.html?asset';
 
 import type { PluginConfig } from '@/types/plugins';
+import BrowserWindowConstructorOptions = Electron.BrowserWindowConstructorOptions;
 
 if (!is.macOS()) {
   delete allPlugins['touchbar'];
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Before:
image


After:
image

ArjixWasTaken:
@thesaikis

Since you are an arch linux user, I'll assume you know how to run this project from the source code.
Could you please try out this patch and report if it works or not?

I can give instructions if you don't know how.

thesaikis:
I just followed the instructions from the project and used `pnpm start`. I also went ahead and built it, then ran the youtube-music binary from the pack/linux-unpacked directory.

I tested it under XWayland (which it uses by default) and also forced Wayland. It works on my end.

Would it make sense for other users to test it under different WM/DE?

Otherwise, everything's good for me. I'll followup if something seems wrong.

ArjixWasTaken:
@h-banii nice rice, can I have the dotfiles? :pray:

JellyBrick (approved):

h-banii:
@ArjixWasTaken thanks, sorry for the late reply

I want to keep my dotfiles private atm, but I really appreciate the compliment!

The software I used to do **all** my widgets is aylurs-gtk-shell (ags). It's written in javascript/typescript, so the learning curve is relatively low. I recommend reading the official docs and the gjs-docs if you're interested.

Also, I think you should be aware of the "v2" PR that will bring tons of breaking changes once it gets merged. I'm not saying you shouldn't invest time in "ags v1", but you might need to rewrite everything from scratch after v2 drops.

PR #2426, opened 2024-09-18
ESLint Flat Config (v9 support #2229)

SomeAspy:
I have remade the config in ESLint's new format, as required by ESLint v9.

## Major changes

- ESLint Prettier plugin removed - there is no way to properly integrate prettier config into ESLint, unless you want to put the prettier config IN the `eslint.config.mjs` file - which I assume would break actual prettier formatters built into IDEs
- ESLint v9 deprecated a *ton* of rules, these have moved into [Stylistic](https://eslint.style/packages/default) which I have added.
- Closes #2229 

## Other changes

- Using my best judgement, I fixed 2 warnings only the new config generated
1. `src/plugins/precise-volume/override.ts` Line 11: 
```diff
- // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```

2. `src/plugins/synced-lyrics/renderer/lyrics/fetch.ts` Line 19: 
```diff
- // eslint-disable-next-line prefer-const
```


JellyBrick:
Could you convert `.mjs` to `.mts`?

Thanks for yout contribution!

SomeAspy:
I don't think that is possible
https://github.com/eslint/eslint/discussions/17726
image
Perhaps there is some roundabout convoluted way to do this, but I don't think it would be worth the effort when the `//@ts-check` directive effectively makes it a typescript file. https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html

SomeAspy:
actually, let me double check. https://github.com/eslint/rfcs/pull/117

SomeAspy:
@JellyBrick this is the conclusion I arrived at:
To parse config files, ESLint is hard coded to use Node (The possibility of using tsx was suggested in the ESLint RFC PR I mentioned earlier, but ultimately did not make it in)
However, Node recently added an **experimental** flag to strip types from TS files [in 22.6.0](https://nodejs.org/en/blog/release/v22.6.0) which means linting will not work on systems with one of the last 3 latest versions of Node - none of which are LTS (At time of writing)

Ignoring that, this lead me to
`NODE_OPTIONS='--experimental-strip-types' pnpm run lint -c eslint.config.mts`
However, this will not work on Windows because Windows requires setting environmental variables with a different command... which brings me to the latest iteration I've come up with
`pnpm cross-env NODE_OPTIONS='--experimental-strip-types' eslint . -c eslint.config.mts"`
broken down:
- cross-env: PNPM integrated command that properly sets environmental variables, regardless of OS
- NODE_OPTIONS: The Node environment variables, with the type stripping set.
- eslint -c: Manually tells ESLint to use `eslint.config.mts` as ESLint does not auto-detect the config file with non-JS extensions

While it's ultimately up to you, as I have provided a working method, I don't think the complexity of the command is worth the effort, when compared to `// @ts-check` paired with `.mjs`, which gives you typescript linting and works out of the box, with the only pitfall being you cannot use typescript keywords (the most important being the `as` keyword or specifying types with `:type` There is an implentation for this in JS however, being JSDoc which most IDEs will grab information from properly

JellyBrick:
@SomeAspy see https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files

SomeAspy:
I think tha would involve having eslint.config.mjs file point to eslint.config.mts, which seems to mostly defeat the purpose. Why do you want the eslint config to be TS?

JellyBrick (approved):

PR #2441, opened 2024-09-23
fix(`synced-lyrics`): Fix 2 issues

Reycko:
First, this PR fixes an issue with the default text string never showing, as it verified whether it was defined, rather than whether or not it's empty.

Secondly, this fixes another, more important issue.
[Line 182 of the fetcher](https://github.com/th-ch/youtube-music/blob/f2716e1dc8c296dfb6a6d7dfc586c9557fca9a1e/src/plugins/synced-lyrics/renderer/lyrics/fetch.ts#L182) uses the `index` element of the `reduce` function, however, this would cause every lyric after an unparsed one to be offset by 1, causing miscalculations for `duration` thus multiple lines showing as current. This PR fixes the issue by using an iterator.

Before:
image

After:
image


Dzheremi2:
If you fixing issue with default string value, please make the ability to disable it to make the same effect as it was with this bug.

PLEASE! This effect, without default string value are beautiful!

Reycko:
> If you fixing issue with default string value, please make the ability to disable it to make the same effect as it was with this bug.
> 
> PLEASE! This effect, without default string value is beautiful!

There's already a ` ` default value which should do the same effect

src/plugins/synced-lyrics/renderer/lyrics/fetch.ts
```diff
@@ -178,10 +178,12 @@ export const getLyricsList = async (
   // Add a blank line at the beginning
   raw.unshift('[0:0.0] ');
 
-  const syncedLyricList = raw.reduce<LineLyrics[]>((acc, line, index) => {
-    const syncedLine = extractTimeAndText(line, index);
+  let i = 0;
```
JellyBrick: Why did you change this line?

JellyBrick (changes requested):

src/plugins/synced-lyrics/renderer/lyrics/fetch.ts
Reycko: raw.reduce doesn't account for `acc`'s length. If we don't push a line, `index` would still be increased by 1.

PR #2524, opened 2024-10-17
Added song url to tuna-obs plugin

53fd670830eafb8acdf1284959bc08b9a10d898:
Added URL as a parameter to the tuna-obs plugin so it can access the URL of the song

### DIFF
--- a/src/plugins/tuna-obs/index.ts
+++ b/src/plugins/tuna-obs/index.ts
@@ -16,6 +16,7 @@ interface Data {
   progress: number;
   status: string;
   title: string;
+  url: string;
 }
 
 export default createPlugin({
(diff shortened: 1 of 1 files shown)


JellyBrick:
Is tuna-obs support it?

53fd670830eafb8acdf1284959bc08b9a10d898:
Song sources like Spotify explicitly support the url parameter. When using this app, since you have to select "Web browser" as the song source, web browser doesn't officially support it, so it puts a red error if you edit the "Song info outputs" section: "The selected music source does not support one or more of the used format options!".
Other than the warning, everything works fine though. It might be a case of adding this app to the list of apps in the tuna-obs UI and listing this as a valid parameter so it hides the warning.

53fd670830eafb8acdf1284959bc08b9a10d898:
image

This is the only problem I've found with it so far, which would be fixed by updating the tuna-obs repo to say this app supports song url (if this goes through, and this app supports song urls, obviously)

JellyBrick (approved):

PR #2713, opened 2024-12-14
fix: tab misalignment

daika7ana:
Fixes the tab misalignment caused by the `blur-nav-bar` plugin.
Let YouTube Music handle its own top alignment, since --menu-bar-height isn’t a concern right now. This way, the plugin focuses solely on applying the background color and blur without interfering with alignment (as it should, if you ask me).

Before:
Screenshot 2024-12-14 031109

After:
Screenshot 2024-12-14 030934

### DIFF
--- a/src/plugins/blur-nav-bar/style.css
+++ b/src/plugins/blur-nav-bar/style.css
@@ -5,7 +5,6 @@
 }
 
 ytmusic-tabs {
-  top: calc(var(--ytmusic-nav-bar-height) + var(--menu-bar-height, 36px));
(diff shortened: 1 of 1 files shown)


JellyBrick:
Is it okay if the playback screen or screen width is reduced?

daika7ana:
Small recording while resizing window:

https://github.com/user-attachments/assets/df5b27eb-8e29-48d7-ae3a-410d0dcb9920




As for the playback screen, the component is not visible there, unless it is and I'm not aware of it.

JellyBrick (approved):

Issue #2540, opened 2024-10-22
feat: community style snippets


ArjixWasTaken:
## The problem

Currently, a lot of people have frustrations with the way music.youtube.com is designed, and would like changes to be done.
The issue is that, it doesn't make sense to make a plugin or update a plugin for every single user suggestion.

Especially since a lot of UI issues can be fixed by just using a custom CSS file, which is already supported by `th-ch/youtube-music`.

## Proposal

We make one plugin that "rules them all" (when it comes to CSS).
It would be a plugin that allows users to apply CSS snippets shared among the community,
similar to [BetterDiscord's Themes](https://betterdiscord.app/themes), but focused on small UI tweaks instead of full client themes.

That way, issues like #2539, would not exist in the first place.

## Specifics

As for how we would go about doing this, there are two routes we can take.

1. Host user-themes on a central github repo, so users would need to make a PR in order to add a tweak.
2. We host a backend where users can sign into and submit CSS snippets directly from within the app.

No matter what route we take, we should inspect each and every CSS snippet, to avoid stuff like key loggers or IP loggers.

## Notes

A big benefit to this would be that non-technical users would be able to use a plethora of appearance tweaks that we would not have included otherwise.

ArjixWasTaken:
CC: @th-ch, @JellyBrick, @Su-Yong

ArjixWasTaken:
Example from a cousin project [spicetify](https://spicetify.app/):

![image](https://github.com/user-attachments/assets/67602a94-d525-404b-a9fd-271491d39571)

PR #2790, opened 2024-12-28
feat(synced-lyrics): romanization

ArjixWasTaken:
Started working on romaji support!

TODO:
- Languages:
- - [x] Japanese
- - [x] Mandarin
- - [x] Korean
- UI
- - [x] synced lyrics
- - [x] plain lyrics (e.g. lyrics genius)

## Preview:

synced lyrics:
image
image

plain lyrics (not synced):
image

---

Related discussion https://github.com/th-ch/youtube-music/discussions/2382


ArjixWasTaken:
I think there is a problem with the pinyin library...

image

I've studied chinese for 2 years, and that should have been `ni de`, I don't even know what `teki` is, but don't take my chinese to be trustworthy :pensive:

you can listen to that song yourself, to see that they say "ni de" and not "ni teki"
https://music.youtube.com/watch?v=16M3bHVbtqY&t=58

ArjixWasTaken:
This brings an issue to the surface, should I run only one romanizer and not all of them?
Like, Japanese have many chinese characters, that are often pronounced in different ways, like the example above, where `de` became `teki`

If I run the pinyin romanization first, it will affect japanese lyrics, but if I run the japanese romanization first, it will affect chinese lyrics.....

ArjixWasTaken:
Ok, for now I ended up with the following solution to the problem described earlier:

- if a line contains hiragana or katakana, it is japanese, if it only contains kanji and doesn't use hiragana/katakana at all, it is chinese

That fixed the issues I encountered in chinese and japanese lyrics, unfortunately, if the lyrics are entirely in kanji, I can't do much about it 😔 

More creative solutions are always welcome!

JellyBrick:
> Ok, for now I ended up with the following solution to the problem described earlier:
> 
> * if a line contains hiragana or katakana, it is japanese, if it only contains kanji and doesn't use hiragana/katakana at all, it is chinese
> 
> That fixed the issues I encountered in chinese and japanese lyrics, unfortunately, if the lyrics are entirely in kanji, I can't do much about it 😔
> 
> More creative solutions are always welcome!

You can use lingua for language detection

https://github.com/pemistahl/lingua-js

ArjixWasTaken:
That sounds a little bit overkill, and honestly, I doubt it would work unless it knows the difference between Chinese and Japanese grammar, because a lot of kanji characters are 1:1 the same as Chinese characters

JellyBrick:
If a sentence contains shinjitsu, it can be considered Japanese (although there are cases where it uses kanji, but in those cases I think it's mostly distinguishable by the presence of hiragana and katakana)

- https://github.com/pemistahl/lingua-rs/pull/381
- https://github.com/pemistahl/lingua-go/issues/38

JellyBrick:
If a sentence contains shinjitai (新字体), it can be considered Japanese (although there are cases where it uses kanji (旧字体), but in those cases I think it's mostly distinguishable by the presence of hiragana and katakana)

- https://github.com/pemistahl/lingua-rs/pull/381
- https://github.com/pemistahl/lingua-go/issues/38

JellyBrick:
Also, if a sentence contains Korean kanji, it can be considered Korean.
- https://namu.wiki/w/%EA%B5%AD%EC%9E%90(%ED%95%9C%EC%9E%90)
  - (Korean)

ArjixWasTaken:
Hmm, I think it should be safe to assume that if at least one line contains katakana/hiragana, then we are dealing with japanese for all lines.

If no japanese/korean is detected anywhere, that's when I should use the pinyin library, I think that should work fine for most cases?

I really don't want to add more dependencies

ArjixWasTaken:
It should be much more reliable now, I doubt there are many japanese songs that are exclusively in kanji, if at least one line contains one character that is hiragana/katagana, then all lines are treated as japanese

ArjixWasTaken:
es-hangul doesn't have a method to test if a given text contains hangul, so I found some regex to do the job
the regex doesn't have to be perfect, don't overthink it, we just want to know if at least one line contains a hangul character or not, to determine what romanizer to use, I know korean unicode is complex, but this really should be enough

---

image

kimjammer:
Got an edge case for you, some songs are in 3 languages. Currently it looks like only one romanizer can be active for a song. In this particular song, Japanese and Korean don't appear on lines together so you could just pick the romanizer line by line, but I'm not sure if there are songs out there that do.
image
We love CJK 😃😃😃.
Anyhow, looking forward to this feature.
(I'm also going to assume you've noticed that the fancy font size isn't being applied correctly)

ArjixWasTaken:
> Got an edge case for you, some songs are in 3 languages. Currently it looks like only one romanizer can be active for a song. In this particular song, Japanese and Korean don't appear on lines together so you could just pick the romanizer line by line, but I'm not sure if there are songs out there that do. image We love CJK 😃😃😃. Anyhow, looking forward to this feature. (I'm also going to assume you've noticed that the fancy font size isn't being applied correctly)

fuck those songs (╯°□°）╯︵ ┻━┻

ArjixWasTaken:
@kimjammer could you send a list of 2-3 songs that have lyrics and contain chinese+korean+japanese?

kimjammer:
By 3 languages I meant Japanese + English + Korean, but I think it would be good to have a list of mixed language songs for testing so:
JE, KE = Japanese and English on same line, Korean and English on same line, Japanese and Korean NOT in same line

J, K, JE, KE: LIT - STAYC
J, K, JE, KE, JK, JKE: Bad Girl for You - EXID
K, C, KE: HIGHLIGHT - SEVENTEEN
C, CK, CE: HWAA (Chinese Ver.) - (G)I-DLE
C, J, K, CE, KE: From Home - NCT U

I don't think any song has Chinese and Japanese on the same line
Good luck!

ArjixWasTaken:
@kimjammer can you confirm if this is correct?

https://music.youtube.com/watch?v=WRWg4EA_dwQ&t=161
image

kimjammer:
I pulled the pr to test it and I can confirm that the top 4 songs look right to me.

ArjixWasTaken:
I've implemented everything I planned to, so I am now ready to accept code reviews!

ArjixWasTaken:
> (I'm also going to assume you've noticed that the fancy font size isn't being applied correctly)

@kimjammer can you elaborate on that? does that still happen with the latest commit?

src/plugins/synced-lyrics/style.css
```diff
@@ -8,16 +8,23 @@
...
   --lyrics-padding: 0;
 
   /* Typography */
-  --lyrics-font-family: Satoshi, Avenir, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell,
-  Open Sans, Helvetica Neue, sans-serif;
-  --lyrics-font-size: clamp(1.4rem, 1.1vmax, 3rem);
+  --lyrics-font-family: Satoshi, Avenir, -apple-system, BlinkMacSystemFont,
+    Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell, Open Sans, Helvetica Neue,
+    sans-serif;
+  --lyrics-font-size: clamp(1.4rem, 1.1vmax, 3rem) !important;
```
kimjammer: The "!important" on this line seems to prevent this variable from being set correctly in the code. In the last commit, the fancy mode follows the `clamp(1.4rem, 1.1vmax, 3rem)` font-size even though it should be `3rem`. Removing !important seems to fix the problem.

kimjammer:
One more thing, are you planning to have a menu toggle for romanization?

src/plugins/synced-lyrics/style.css
kimjammer: This also makes the romanization the full 3rem, it might be worth considering making the romanization smaller than 3rem so the lyrics page isn't too cramped.

ValiantZippu:
Nice work but I hope the design would be kind of flawless and modern like the extention Better Lyrics - Chrome
And also translation of lyrics . and can the ui kind of be like this ? the grayish bar and the black side bar looks not as clean as this one . 

image

ArjixWasTaken:
ngl I find that ugly

ValiantZippu:
> ngl I find that ugly

I respect your taste but I think everyone has their own choice it would be cool to have a toggle in the ambient plugin to make the entire now playing page like that . Main highlight is the lyrical aesthetic from it

ArjixWasTaken:
@ValiantZippu 

I don't want or need you to respect my taste, and that should apply in reverse as well,
instead of suiting everyone's taste (impossible to do), focusing on a simple interface and letting the community enhance it would be the best

https://github.com/th-ch/youtube-music/issues/2540

if you'd like, you can make suggestions on that issue, to make it a reality

ArjixWasTaken:
Done with uni exams!

ArjixWasTaken:
too lazy to get back at this, someone give me motivation (insults)

kimjammer:
plz finish I've been waiting patiently for two months 🙏

ArjixWasTaken:
oh darn it, fine

DeltaLaboratory:
When we add some lyric translate feature, what would romanized+translated lyric UI look like?
Just asking because I am willing to add translate feature based on this draft.

ArjixWasTaken:
I am going to work on this tomorrow (on the 2nd of the month), and will make it ready for merge.

---

@DeltaLaboratory I guess it'd be below the romanized line? we'd need to add settings so it's the UI is not cluttered tho

ArjixWasTaken:
tweaked the CSS a bit,
image

ArjixWasTaken:
I am ready for reviews, fire them at me!

kimjammer (approved):
LGTM! I tested it, and it works perfectly for all songs that don't contain both J and C. Thank you for finishing it!

src/plugins/synced-lyrics/backend.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+
+    ipc.handle('synced-lyrics:romanize-korean', async (line: string) => {
+      let result = line;
+
+      result = Hangul.romanize(result);
+
+      // handle left-over chinese characters
+      result = result.replaceAll(/[\u4E00-\u9FFF]+/g, (text) =>
+        pinyin(text, {
```
JellyBrick: Korean hanja are pronounced differently from Chinese.

JellyBrick (changes requested):

JellyBrick (approved):

ArjixWasTaken:
@JellyBrick thanks for improving hangul!

I noticed that there is a regression to the `canonicalize` function, as it no longer fixes `" datte "` to `"datte"` as seen in

before:
image

after:
image

---

other than that, amazing work!

ArjixWasTaken:
Since we are on the topic of improving the accuracy of the romanization, what can we do about japanese?

`獣は砂を一握り撒いた` gets romanized as `shishi wa suna o ichi nigiri mai ta`, but what I can hear is `kemono wa su na wo`

I don't know japanese, but I trust my ears to recognize such a different :p

ArjixWasTaken:
> > Since we are on the topic of improving the accuracy of the romanization, what can we do about japanese?
> > `獣は砂を一握り撒いた` gets romanized as `shishi wa suna o ichi nigiri mai ta`, but what I can hear is `kemono wa su na wo, hitonigiri maita`
> > I don't know japanese, but I trust my ears to recognize such a difference :p
> > EDIT: google translate says "Kemono wa suna o hitonigiri maita", which sounds accurate
> 
> @ArjixWasTaken The reason this issue occurs is that Kanji in CJK languages can be read differently depending on context and situation. For example, in the sentence above, the character "獣" meaning "beast" can be read as "けだもの" (kedamono), "けもの" (kemono), or "しし" (shishi). Another example is "行こう" which means "let's go" In a literary or old-fashioned context, it may be read as "yukou" while in everyday usage, it is typically read as "ikou" There is no way to solve this issue other than using AI.

ig in that case, I'll have to think of gathering user-submitted corrections in a database like lrclib

ArjixWasTaken:
actually, should I add support for google's API?
https://cloud.google.com/translate/docs/advanced/romanize-text

the user would have to provide the API keys

JellyBrick:
I will merge this PR for now and work on improvements later.

AMZMA:
> actually, should I add support for google's API? https://cloud.google.com/translate/docs/advanced/romanize-text
> 
> the user would have to provide the API keys but it doesn't have a free tier as far as I can see
> 
> the pricing calculator says that translating 500.300 characters is free, anything above that costs money, dunno how that works

yes, please. and add support for any other non Roman text (Russian, Greek, Arabic, Hindi, etc).

this one here says its free and unlimited
https://github.com/ssut/py-googletrans

google transliteration is superior than any other.

PR #2792, opened 2024-12-28
feat(api-server): add endpoint to get shuffle state

franz-dc:
Added API endpoint: `/api/v1/shuffle` (GET)

`canSetFullscreen` has been renamed to `canUsePlayerControls` to be used by other player control states other than fullscreen. Can be reverted back if that's not the desired outcome.

### DIFF
--- a/src/plugins/api-server/backend/routes/control.ts
+++ b/src/plugins/api-server/backend/routes/control.ts
@@ -173,7 +173,24 @@ const routes = {
       },
     },
   }),
-
+  getShuffleState: createRoute({
+    method: 'get',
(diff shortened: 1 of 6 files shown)


ArjixWasTaken:
You should use `ipc.handle` and `ipc.invoke` instead of `ipc.send()` and `ipc.once()`

ArjixWasTaken:
Oh I see, the renderer thread is what "handles" the request, that's why you did it this way.
So I assume electron doesn't support for the renderer to "handle" an event, because there can be multiple renderers!

JellyBrick (approved):

PR #2804, opened 2024-12-29
fix(async): add missing await statements

yilmazcabuk:
### Summary
This pull request ensures that all asynchronous operations properly use the `await` keyword, improving reliability and consistency in the codebase.

### Details
- **fix:** Added missing `await` statements to resolve potential timing and unhandled promise issues.

### Testing
- Tested affected code paths to confirm no regressions or unintended side effects.

### DIFF
--- a/src/i18n/resources/fr.json
+++ b/src/i18n/resources/fr.json
@@ -317,7 +317,7 @@
           "title": "Nom d'hôte"
         },
         "port": {
-          "label": "Entrer le port du serveur de l'API :",
+          "label": "Entrez le port du serveur de l'API:",
           "title": "Port"
         }
       }
@@ -726,7 +726,7 @@
     "synced-lyrics": {
       "description": "Ajoute des paroles synchronisées aux chansons, grâce à LRClib par exemple.",
       "errors": {
-        "fetch": "⚠️ - Une erreur s'est produite en allant chercher les paroles. Merci de réessayer plus tard.",
+        "fetch": "⚠️\tUne erreur s'est produite en allant chercher les paroles.\n\tMerci de réessayer plus tard.",
         "not-found": "⚠️ - Aucune paroles trouvées pour cette musique."
       },
       "menu": {


src/index.ts
```diff
@@ -495,7 +495,7 @@ async function createMainWindow() {
     }
   });
 
-  win.webContents.loadURL(urlToLoad);
```
JellyBrick: This is intentional behavior: using the `await` keyword prevents the `did-load-failed` event from being called as expected.

src/index.ts
```diff
@@ -823,7 +823,7 @@ app.whenReady().then(async () => {
   }
 
   if (config.get('options.hideMenu') && !config.get('options.hideMenuWarned')) {
-    dialog.showMessageBox(mainWindow, {
```
JellyBrick: In this code, the result of the call to `showMessageBox` does not matter, so the `await` keyword is not required.

src/loader/renderer.ts
```diff
@@ -19,7 +19,7 @@ export const createContext = <Config extends PluginConfig>(
   id: string,
 ): RendererContext<Config> => ({
   getConfig: async () =>
-    window.ipcRenderer.invoke('ytmd:get-config', id) as Promise<Config>,
```
JellyBrick: `await` keyword is not required: `ipcRenderer.invoke` already return `Promise<T>`

src/loader/renderer.ts
```diff
@@ -19,7 +19,7 @@ export const createContext = <Config extends PluginConfig>(
...
+    (await window.ipcRenderer.invoke('ytmd:get-config', id)) as Promise<Config>,
```
JellyBrick: Invalid type: If you use the await keyword here, the return value of await-invoke is `T`, not `Promise<T>`.

src/plugins/downloader/main/index.ts
```diff
@@ -647,7 +647,7 @@ export async function downloadPlaylist(givenUrl?: string | URL) {
     mkdirSync(playlistFolder, { recursive: true });
   }
 
-  dialog.showMessageBox(win, {
```
JellyBrick: Ditto

src/plugins/music-together/index.ts
```diff
@@ -631,7 +631,11 @@ export default createPlugin<
     start({ ipc }) {
       this.ipc = ipc;
       this.showPrompt = async (title: string, label: string) =>
-        ipc.invoke('music-together:prompt', title, label) as Promise<string>;
+        (await ipc.invoke(
```
JellyBrick: Ditto

src/plugins/music-together/index.ts
```diff
@@ -631,7 +631,11 @@ export default createPlugin<
...
+          'music-together:prompt',
+          title,
+          label,
+        )) as Promise<string>;
```
JellyBrick: Invalid type: ditto

JellyBrick (changes requested):

PR #2808, opened 2024-12-30
feat(api-server): add `insertPosition` for `addSongToQueue`

### LINKED ISSUE #2803: [Feature Request]: API Server - play-next for putting song after current playing
### Preflight Checklist

### Problem Description

Enhancing #2767

This thing didn't get putting song just after current one. It is pretty useful, i.e. for streamers, so people request a song and it is played just after one from streamer's queue is completed.

### Proposed Solution

By researching source code, I found that enum is already prepared for making such a thing
https://github.com/th-ch/youtube-music/blob/85d1dc46d120ace62f2f9d18dce5404ba2a5838f/src/types/datahost-get-state.ts#L431

### Alternatives Considered

I'm not sure for any alternatives, cause even building external API with calling all the methods like /api/v1/queue is way more complicated for me. Because even checking for current playing song's position in queue takes a lot of conditions. 

### Additional Information

_No response_

franz-dc:
Closes #2803, but hooking into POST `/api/v1/queue` instead of creating a new endpoint.

Request body:
```json
{
  "videoId": "...",
  "insertPosition": "INSERT_AFTER_CURRENT_VIDEO"
}
```

`insertPosition` can be omitted and will default to the original behavior (`INSERT_AT_END`).

### DIFF
--- a/src/plugins/api-server/backend/routes/control.ts
+++ b/src/plugins/api-server/backend/routes/control.ts
@@ -660,8 +660,8 @@ export const register = (
   app.openapi(routes.queueInfo, queueInfo);
 
   app.openapi(routes.addSongToQueue, (ctx) => {
-    const { videoId } = ctx.req.valid('json');
-    controller.addSongToQueue(videoId);
+    const { videoId, insertPosition } = ctx.req.valid('json');
+    controller.addSongToQueue(videoId, insertPosition);
 
     ctx.status(204);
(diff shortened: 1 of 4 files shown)


cireCloud:
Hello. Would it be possible to have index as well? So it can be removed later on.
Or at least response to give index it was placed in? Thanks!

PR #2964, opened 2025-02-09
Added Global Shortcut Support for Rewinding and Seeking by 10 Seconds.

poipoi300:
Changes were added to the existing global shortcuts plugin, tested both the original and new shortcuts working. 2 i18n strings added.

### src/plugins/shortcuts/main.ts
- Used an if/else statement on shortcut action type to determine if a parameter is required; Rewinding and Seeking require a parameter while previous shortcuts did not.
### src/plugins/shortcuts/menu.ts
- Adjusted height and specified width for the shortcuts menu; The two added entries pushed the ok/cancel out of frame and one of the shortcut descriptions overflowed width-wise. New screen looks like this 
image

## Limitations
The amount of seconds to rewind and seek by are hard-coded in i18n, a change in the amount in the shortcut would require a change there too. However, since 4 files needed to be touched for this PR, I consider that this approach matches the existing structure of the codebase.


poipoi300:
Will re-open when I add the number of seconds to i18n and pr the merge with shortcuts for liking and disliking as well.

Issue #2888, opened 2025-01-19
[Bug]: Downloader not working


Nakxin:
### Preflight Checklist

- [x] I use the latest version of YouTube Music (Application).
- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.7.2.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [x] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 11

### What CPU architecture are you using?

x64

### Last Known Working YouTube Music (Application) version

3.6.1

### Reproduction steps

Go on any music video press download, and be introduced with an error

![Image](https://github.com/user-attachments/assets/2f0f65da-a04f-4cbd-a3c6-eeb205ffcee8)

### Expected Behavior

Find the song 
press the 3 dots get the menu to pop down
press the download option 
for it to work without giving me an error box

### Actual Behavior

I press download it comes with an error
I tried on many songs and the problem still persists

![Image](https://github.com/user-attachments/assets/7d8f728f-78b2-4118-a3a7-31a11b59d994)

### Enabled plugins

Compact sidebar
in-app menu
skip disliked songs
ad blocker
downloader
precise volume
ambient mode 
album color theme
API Server [Beta]

### Additional Information

I tried a custom option on the downloader but was introduced with a download button that did nothing no matter how many times i clicked
I tried Override User-Agent didn't work even after system restarts 
I tried uninstalling and all the files associated to the program and did a full clean install still didn't work with the web.exe
I tried the portable version as well same issue persisted
I tried different type of songs (Short and long songs)
I tried disabling API Server [Beta] and the download button turned into a button that did nothing.
I tried re-enabling API Server [Beta] and the download button worked again but still caused an error when I tried downloading music
This issue is annoying and i've seen a bunch of people have this problem i tried following their steps in fixing the issue and yet it still persists.
I don't have enough knowledge about code to know what to look for when it comes to problems so if ill be given any feedback please dumb it down for me because im not familiar with most terms y'all would use.

jazzybeat:
Same here

[...]

PR #2984, opened 2025-02-16
fix(downloader): correct logic for YouTube Premium detection

franz-dc:
Another fix to address 304 errors when downloading (#2888).

### DIFF
--- a/src/plugins/downloader/main/index.ts
+++ b/src/plugins/downloader/main/index.ts
@@ -57,7 +57,7 @@ let playingUrl: string;
 
 const isYouTubePremium = () =>
   win.webContents.executeJavaScript(
-    '!document.querySelector(\'#endpoint[href="/music_premium"]\')',
+    '!!document.querySelector(\'#endpoint[href="/music_premium"]\')',
   ) as Promise<boolean>;
 
 const sendError = (error: Error, source?: string) => {


src/plugins/downloader/main/index.ts
```diff
@@ -57,7 +57,7 @@ let playingUrl: string;
 
 const isYouTubePremium = () =>
   win.webContents.executeJavaScript(
-    '!document.querySelector(\'#endpoint[href="/music_premium"]\')',
+    '!!document.querySelector(\'#endpoint[href="/music_premium"]\')',
```
JellyBrick, line 60: Invalid fix. If the user is using YouTube Premium, the result of `document.querySelector('#endpoint[href="/music_premium"]')` is null.
so `!document.querySelector('#endpoint[href="/music_premium"]')` is true if using YT Premium.

JellyBrick (changes requested):

src/plugins/downloader/main/index.ts
franz-dc: I do not have YouTube Premium, but it returns null.

franz-dc:
My bad, it seems this is actually caused by another bug, I'll open an issue instead.

Issue #3007, opened 2025-02-21
[Bug]: Downloader wont download


DerpSlie:
### Preflight Checklist

- [x] I use the latest version of YouTube Music (Application).
- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.7.4

### Checklists

- [x] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 10

### What CPU architecture are you using?

x64

### Last Known Working YouTube Music (Application) version

It worked yesterday on 3.0 but i tried keeping it still at the version but it didn't work so I updated with no luck.

### Reproduction steps

Download a song just from the app

### Expected Behavior

Song gets downloaded to my files. 

### Actual Behavior

Doesnt download song and puts up an error code saying "The server responded with a non 2xx status code in..." then it just puts random letters and numbers.

### Enabled plugins

1. Ad blocker 
2. Ambient mode
3. Downloader
4. In app menu
5. Navigation
6. Precise volume
7. Shortcuts 
8.  

### Additional Information

![Image](https://github.com/user-attachments/assets/38826fd9-7a46-4b0f-b191-5dd2b8645382)

Nezgit41:
Similar problem, has anyone found a way to solve it?

Prat-247:
Same issue here, I am facing this issue from past month

senolfndk:
same problem

OGxBEAST:
Same Issue

adit300701:
![Image](https://github.com/user-attachments/assets/5c98d1ca-c222-4db6-a211-c69536281025)

adit300701:
![Image](https://github.com/user-attachments/assets/1dc30570-9e8f-41e4-a83b-7ba6c5b78f89)

senolfndk:
Same problem, can't download songs

JellyBrick:
> Same problem, can't download songs ![Image](https://github.com/user-attachments/assets/240c2eb2-744f-4b6f-93a3-585883f2ae8f)

Can you sign in to your Google account and try again?

senolfndk:
[...]

Issue #2978, opened 2025-02-14
[Bug]: Plugin Downloader doesnt fetch entire playlist, but exactly 100 tracks, if Playlist has more than 100 tracks


Shadowphoenix:
### Preflight Checklist

- [x] I use the latest version of YouTube Music (Application).
- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.7.2.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Win 11 24H2 IoT LTSC - EDIT: Also Tested to happen on a standard Win 10 24H2 Pro Installation

### What CPU architecture are you using?

x64

### Reproduction steps

Use Plugin "Downloader" to try and Download playlist with more than 100 Tracks.

### Expected Behavior

Fetch entire playlist. 

If this isnt feasible because the Array or whatever is storing the Information about each track overflows, allow me to select what ID-Range inside the Playlist i want to fetch (i.e. track 1 to 100 or track 75 to 174 etc.) up to a maximum of 100 individual tracks. This way it would still be able only keep 100 tracks in the array while allowing me to easily split the playlist and simply running several download-sequences starting from different IDs.

### Actual Behavior

The Dowload-Plugin as is sets a cutoff where it doesnt fetch more than 100 songs of a playlist. It seems to always fetch the first one 100 songs and none beyond that.

This is annoying because everything that isnt in the range of tracks that get downloaded the first time seemingly has to be put in a separate playlist to be dowloaded from there and will never be fetched by the "Download Playlist"-Function from within the first playlist.

This seems to be the case with all Playlists i tested, one only barely exceeding the 100 tracks (102), others being over 500 Tracks long.

### Enabled plugins

Downloader
SponsorBlock
Adblock

### Additional Information

For convenience of testing and repeatability, here is the link to a playlist by me that is  longer (about 240ish tracks). in this example, inbetween "DV & LM vs W&W - Arcade" and  "KOKJ - Jungle Vibes" seems to be the cutoffpoint, Jungle Vibes seems never to get fetched.

[...]

PR #3015, opened 2025-02-22
fix(deps): update dependency youtubei.js to v13.1.0

franz-dc:
Fixes #3007
Also fixes #2978

PR #3022, opened 2025-02-25
fix: Filter for only `MusicResponsiveListItem` in playlist items

DeathCamel58:
This is to filter out the `ContinuationItem` that YouTube.js now has in the items in a playlist that cause an error.

Fixes https://github.com/th-ch/youtube-music/issues/3007
Also fixes https://github.com/th-ch/youtube-music/issues/2978

This needs to be applied with [PR #3105](https://github.com/th-ch/youtube-music/pull/3015)


JellyBrick (approved):

JellyBrick:
Thanks!

PR #3025, opened 2025-02-26
fix: Updated tray pause icon for consistency

dvd-22:
I don't know if any of you were bothered by this, but I was.

Now the icon looks the same as the YT Music logo, but with a pause icon. I tested it for reductions and it should look just fine.


dvd-22:
This is how it looks side by side with the YT Music one

image

JellyBrick (approved):

PR #3066, opened 2025-03-07
feat(plugin): settings-ui

ArjixWasTaken:
## About

This is a plugin that introduces a new settings interface!
Configuring everything through menu options is not only a pain, but it is also a pain! (double pain!)
I propose that this plugin is enabled by default, but I won't insist on that if you think otherwise.

## TODO
- [x] Add settings button
- [x] Add settings modal
- - [x] General
- - [ ] Appearance
- - [ ] Plugins
- - [ ] Advanced
- - [ ] About

## Screenshots

Full sidebar             |  Mini sidebar
:-------------------------:|:-------------------------:
image  |  image

general


ArjixWasTaken:
I drew some inspiration from vencord :pensive:

ArjixWasTaken:
Before I develop this further, I'd like your opinions, any suggestions as to what/how the settings should look like?

@th-ch, @JellyBrick, @Su-Yong

src/plugins/settings-ui/components/SettingsModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+  { icon: 'yt-icons:emoji', id: 'plugins' },
+  { icon: 'yt-icons:service_toolbox', id: 'advanced' },
+  { icon: 'yt-icons:info', id: 'about' },
+];
+
+interface SettingsModalProps {
+  close: () => void;
+}
+
+export default ({ close }: SettingsModalProps) => {
```
JellyBrick: In Solid.js, you should not use object destructuring on props. If you destructure props, you will lose reactivity.

JellyBrick (changes requested):

src/plugins/settings-ui/components/SettingsModal.tsx
ArjixWasTaken: Yeah I know, I just don't have reactive values in that specific instance

ArjixWasTaken: ig it is good practice to not destructure the props anywhere, just for the sake of being consistent

ArjixWasTaken:
Question:

Should the settings:
- be applied as soon as you change them (like the menu)
- be applied after pressing a confirm button to close the modal

@JellyBrick, @th-ch

Rairof:
https://github.com/th-ch/youtube-music/pull/3066#issuecomment-2781732433
a) Imo as that's the expected behaviour most users would think and so they won't be confused as to why their selected settings don't work immediately. It would be better if a pop up to restart the client is shown with settings that needs a restart to take effect.

ArjixWasTaken:
@Rairof you didn't understand what I meant, so of course you'd think so

I meant we'd not allow the user to close their settings unless there are no unsaved changes, just like how discord does it

Rairof:
https://github.com/th-ch/youtube-music/pull/3066#issuecomment-2837458315
In that case. my proposal was to automatically apply said user changes inside the settings menu (like how it is now with the in-app menu) with the sole exception being the few plugins/functions that require a restart to function.

JellyBrick:
> Question:
> 
> Should the settings: a) be applied as soon as you change them (like the menu) b) be applied after pressing a confirm button to close the modal
> 
> @JellyBrick, @th-ch

a + throttle (e.g. 5sec)

ArjixWasTaken:
new UI using mdui:

image

:confetti_ball: 

looks so much more better now

ArjixWasTaken:
PS: I am reviving this PR, but from scratch.

PR #3074, opened 2025-03-09
Discord Rich Presence Fix

Tx-ID:
Discord doesn't like it when you put more than 128 characters on their rich presence.

### DIFF
--- a/src/plugins/discord/main.ts
+++ b/src/plugins/discord/main.ts
@@ -40,7 +40,7 @@ const refreshCallbacks: (() => void)[] = [];
 
 const truncateString = (str: string, length: number): string => {
   if (str.length > length)
-    return `${str.substring(0, str.length - 3)}...`;
+    return `${str.substring(0, 125)}...`;
   return str;
 }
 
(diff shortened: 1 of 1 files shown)


src/plugins/discord/main.ts
```diff
@@ -38,6 +38,12 @@ const info: Info = {
  */
 const refreshCallbacks: (() => void)[] = [];
 
+const truncateString = (str: string, length: number): string => {
+  if (str.length > length)
+    return `${str.substring(0, str.length - 3)}...`;
```
JellyBrick:
```suggestion
    return `${str.substring(0, 125)}...`;
```

JellyBrick (changes requested):

src/plugins/discord/main.ts
Tx-ID: now that I think about it, I don't think it should be done that way since it accepts variable length, unless you want to remove that variable length then it's valid

Tx-ID: will make change after this

Tx-ID:
fixed

JellyBrick (approved):

JellyBrick (approved):

PR #3093, opened 2025-03-15
feat: Allow scrobbling using alternative song titles

Yumeo0:
This adds a new setting to the scrobbler plugin which uses the `linkAlternates` youtube provides to scrobble with the alternative title (mostly for non-roman song titles).

image


JellyBrick (approved):

JellyBrick:
Awesome! Thanks for your contribution!

PR #3145, opened 2025-03-27
fix(downloader): allow downloads for signed out users

franz-dc:
For signed out users, the upgrade button will not exist, so the sign in button is used instead for premium checking.

If signed in, the sign in button will not exist in the DOM.

### DIFF
--- a/src/plugins/downloader/main/index.ts
+++ b/src/plugins/downloader/main/index.ts
@@ -56,6 +56,14 @@ let win: BrowserWindow;
 let playingUrl: string;
 
 const isYouTubeMusicPremium = async () => {
+  // If signed out, it is understood as non-premium
+  const signInButton = (await win.webContents.executeJavaScript(
+    'document.querySelector("#sign-in-button")',
+  )) as HTMLElement | null;
+
+  if (signInButton) return false;
+
+  // If signed in, check if the upgrade button is present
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
If we are to improve this, might as well go the extra mile.

To check if the user is signed in or not, you can simply do `yt.config_.LOGGED_IN`.
I haven't found an alternative for checking if the user is premium or not.

franz-dc:
> If we are to improve this, might as well go the extra mile.
> 
> To check if the user is signed in or not, you can simply do `yt.config_.LOGGED_IN`. I haven't found an alternative for checking if the user is premium or not.

Thanks!

src/plugins/downloader/main/index.ts
```diff
@@ -56,17 +56,25 @@ let win: BrowserWindow;
 let playingUrl: string;
 
 const isYouTubeMusicPremium = async () => {
+  // If signed out, it is understood as non-premium
+  const isSignedIn = (await win.webContents.executeJavaScript(
+    'yt.config_.LOGGED_IN',
```
JellyBrick:
```suggestion 
    '!!yt.config_.LOGGED_IN',
```

JellyBrick (changes requested):

JellyBrick (approved):

PR #3162, opened 2025-03-30
perf(synced-lyrics): virtual scrolling

ArjixWasTaken:
An attempt to improve the performance of the synced-lyrics renderer.

TODO:
- [x] use [`virtua`](https://www.npmjs.com/package/virtua) for the virtual scrolling
- [x] move calculation of `current`|`previous`|`upcoming` to the parent component
- [x] fix auto-scroll when the current line is not rendered
- [x] lyrics picker
- - [x] keep mounted - https://github.com/inokawa/virtua/pull/681
- - [x] auto show/hide based on scroll-state and/or mouse movement

This should massively improve the performance of the lyrics renderer, and prevent the huge lags that take place after switching songs.


package.json
```diff
@@ -299,6 +299,7 @@
     "solid-styled-components": "0.28.5",
     "solid-transition-group": "0.3.0",
     "ts-morph": "25.0.1",
+    "virtua": "^0.40.3",
```
JellyBrick:
```suggestion
    "virtua": "0.40.3",
```

Please use **-E** option like `pnpm i -E virtua`

tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "resolveJsonModule": true,
-    "moduleResolution": "node",
```
JellyBrick, line 11: If `moduleResolution` is set to `bundler`, certain dependencies cannot be imported. e.g. `youtubei.js`.

JellyBrick (changes requested):

tsconfig.json
ArjixWasTaken: ~~can you ellaborate on that?~~ I see the issue
any other `moduleResolution` introduces issues

well, technically we are using a bundler, so this is more correct
but I'll have to try to resolve that issue as well

ArjixWasTaken: looks like it wasn't a big deal

ArjixWasTaken:
I am actually stuck about the lyrics picker no longer auto-hiding.
I opened a [discussion](https://github.com/inokawa/virtua/discussions/669) on the `virtua` repository to discuss how this should be properly done.

Otherwise, I'll have to resort to weird trickery using absolute positioning, which would look ass.

ArjixWasTaken:
If `keepMounted` is added for the solidjs variant of virtua, I can fix the css picker to be sticky once again.
https://github.com/inokawa/virtua/pull/674

ArjixWasTaken:
Alright, I think by the end of today I'll mark this PR ready for review!

ArjixWasTaken:
@JellyBrick could you put a warning in the release notes, that this will likely break any custom CSS applied to the synced-lyrics?

ArjixWasTaken:
Squashed the commits because they were too many.

ArjixWasTaken:
@JellyBrick I decided to include a better romanization method in this PR

No need to worry about tiny-pinyin anymore, I not only fixed the `w e i r d s p a c i n g` it has for English, but I massively improved the quality of the romanized lyrics by using `tinyld`

and then as fallback I use the old detection methods, and I noticed a huge improvement in japanese/chinese songs

the reason I included these changes in this PR, is because it would be "laggy" to do it w/o virtual scrolling
virtual scrolling allows us to lazily romanize the lyrics


--

before:
image

after:
image


--

before:
image

after:
image

package.json
```diff
@@ -300,7 +300,9 @@
     "solid-styled-components": "0.28.5",
     "solid-transition-group": "0.3.0",
     "tiny-pinyin": "1.3.2",
+    "tinyld": "^1.3.4",
```
JellyBrick:
```suggestion
    "tinyld": "1.3.4",
```

JellyBrick (approved):

PR #3202, opened 2025-04-06
feat(plugin): playlist and album search plugin

donovan-yohan:
Was recently very irritated by this lack of functionality on ytmusic and decided to implement a pretty naive monkey brained solution that I think is about on par with the fact that this feature hasn't been implemented in over 7 years of complaints and requests.

Nothing too crazy going on here, it just leverages the built-in infinite scroll listener for loading playlists to hide elements that don't match the filter query, triggering a load of more elements which are in turn hidden if they don't meet the query either

for giant playlists (like mine of over 3000 imported from spotify) it can take almost a minute to resolve, but assuming you sort by newest, or some method that'll get you to what you want faster, you can simply click on what you want and ignore the rest. I'd say it's better than nothing.

Bonus points, all those tracks are loaded now, so since it just hides the dom elements, you can change your search on the already loaded items and get results much quicker.

It works using by injecting the exact same component as the search bar, so hopefully it won't conflict with anybody's existing themes (except for one override on the border-radius to keep it from showing the suggestions built-in to the regular youtube search)

Honestly, it seems that the performance of YTMusic in general is not great while content is loading, even on the original browser app. It blocks itself from loading or updating any existing images while a page search is happening. You can type but not see the results until the next page finishes loading, and also the ability to click play on items while it's searching is a little finnicky, but I did this work so I'm putting it here to get thoughts from others.

The translations files looked like they were too pretty to be edited manually, but please let me know what process I should follow to at least get the english started for these. There isn't much, just the plugin name and some placeholder strings for the search bar.

image

Since the Title, Artist, and Album are all already loaded by default, it can check against all of them for a match
image

image

## Summary by CodeRabbit

* **New Features**
  * Added playlist and album search for YouTube Music shelf views, letting you quickly filter visible items as you type.
  * Search behavior adapts to the current shelf type with tailored placeholder text and improved matching for accented characters.
  * Search stays up to date as playlist content changes during browsing.

* **Bug Fixes**
  * Improved navigation handling so search initializes correctly when moving between pages.
  * Cleaned up the search UI by removing conflicting built-in suggestion and clear elements.


Laesx:
I tested it and seems like a really good idea, though from a quick test it seems a bit slow.

For big lists as it's loading the next entries it flashes the list a few times until it finds the song.

I think virtual lists were being implemented but until they are if you load the entire playlist it slows the app to a crawl for long lists, this is not really an issue of this plugin though but my list is only 700 entries, I can't even imagine how bad it must be with several thousand big lists hahaha.

Also maybe add the cross to delete the input and the backdrop shadow when text is inputted but the search is out of focus. Both of those are in the native youtube search bar you can just copy them from there.

src/plugins/playlist-search/index.ts
```diff
@@ -0,0 +1,179 @@
...
+        'ytmusic-responsive-list-item-renderer'
+      );
+      items.forEach((item) => {
+        const textElements = item.querySelectorAll('yt-formatted-string');
+        const hasMatch = Array.from(textElements).some((element) =>
+          this.containsSearchTerm(element, searchTerm)
+        );
+
+        (item as HTMLElement).style.display = hasMatch ? '' : 'none';
+      });
```
ArjixWasTaken, line 70: nitpick

```suggestion
        const textElements = item.querySelectorAll<HTMLElement>('yt-formatted-string');
        const hasMatch = Array.from(textElements).some((element) =>
          this.containsSearchTerm(element, searchTerm)
        );

        item.style.display = hasMatch ? '' : 'none';
      });
```

src/plugins/playlist-search/index.ts
```diff
@@ -0,0 +1,179 @@
...
+    },
+    initializeSearch(shelf): void {
+      const existingContainer = document.getElementById(PlaylistSearchBoxID);
+
+      if (existingContainer) {
+        this.searchContainer = existingContainer as HTMLDivElement;
```
ArjixWasTaken, line 76: nitpick

```suggestion
      const existingContainer = document.getElementById<HTMLDivElement>(PlaylistSearchBoxID);
      if (existingContainer) {
        this.searchContainer = existingContainer;
```

src/plugins/playlist-search/index.ts
```diff
@@ -0,0 +1,179 @@
...
+        return;
+      } else {
+        this.waiting = true;
+      }
+      await waitForElement<HTMLElement>('#continuations');
+      this.waiting = false;
+
+      const shelf =
+        document.querySelector(PlaylistShelfComponent) ??
+        document.querySelector(AlbumShelfComponent);
```
ArjixWasTaken, line 152: nitpick

```suggestion
      if (this.waiting) return;

      this.waiting = true;
      await waitForElement<HTMLElement>('#continuations');
      this.waiting = false;

      const shelf = document.querySelector(`${PlaylistShelfComponent}, ${AlbumShelfComponent}`);
```

src/plugins/playlist-search/index.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        if (this.songListObserver) {
+          this.songListObserver.disconnect();
+        }
+
+        this.songListObserver = new MutationObserver(() => {
+          if (this.currentSearchTerm) {
+            debouncedFilter(this.currentSearchTerm, shelf);
+          }
+        });
```
ArjixWasTaken, line 134: nitpick

```suggestion
        this.songListObserver?.disconnect();
        this.songListObserver = new MutationObserver(() => {
          if (this.currentSearchTerm) {
            debouncedFilter(this.currentSearchTerm, shelf);
          }
        });
```

src/plugins/playlist-search/index.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        const debouncedFilter = debounce((search: string, shelfEl: Element) => {
+          this.filterPlaylistItems(search, shelfEl);
+        }, 300);
+
+        this.searchInput.addEventListener('input', (e) => {
+          const newSearchTerm = (e.target as HTMLInputElement).value;
+          this.currentSearchTerm = newSearchTerm;
+          debouncedFilter(newSearchTerm, shelf);
+        });
```
ArjixWasTaken, line 124: nitpick

```suggestion
        this.searchInput.addEventListener('input', (e) => {
          this.currentSearchTerm = this.searchInput.value;
          debouncedFilter(newSearchTerm, shelf);
        });
```

src/plugins/playlist-search/index.ts
```diff
@@ -0,0 +1,179 @@
...
+        if (this.searchInput) {
+          this.searchInput.value = this.currentSearchTerm;
+        }
+        return;
+      }
+
+      this.currentSearchTerm = '';
+      this.searchContainer = document.createElement(
+        'ytmusic-search-box'
+      ) as HTMLDivElement;
```
ArjixWasTaken, line 87: nitpick

```suggestion
      this.searchContainer = document.createElement<'div'>('ytmusic-search-box');
```

ArjixWasTaken (commented):
I am mostly nitpicking, doing essentially what Copilot's review should have done smh.
Feel free to ignore my review, it's just nitpicks of how I'd write my code, not necessarily improvements.

if I had one general suggestion to improve the quality of the code, it would be "please use generic types as they are intended"

not that the code itself had any issues, but code of that style is bound to have weird typescript issues on a larger scale.
try to never have to use `as`, always try to make typescript happy w/o resorting to it

---

PS: If I sound condescending, please know that when I give reviews, I envision it as giving a review to my past self.
I have no intention of making you upset or anything, this is amazing work.

JellyBrick (changes requested):
Can you fix the things we pointed out in the review?

grandosegood:
Hi all, apologies as I'm not very familiar with github. Has this change been committed yet to the app?

PR #3206, opened 2025-04-06
Synced Lyrics: Also search for lyrics with the original title language

Laesx:
I need to test this a bit more, since I have only tested with Japanese songs (Interface in English (US)), I'm not sure if maybe we should keep the original matching as a toggle in the menu.

Added a new array to the song-info type which gets the tags array from microformat.

From what I've seen this array sometimes contains the english artist name, the original artist name, the album name, the album or song producer/composer in both languages, the album name, the song name, etc.

Then we pass alternativeTitle and tags to LRCLib search, here it first tries to the exact match, then searches by the alternative title first, if not found it will fallback to searching the title only again.

On filtering, if the localized artist match is below the threshold, it will compare each tag with each artist from the response.

The songs I've tested are 
Oracle - Amane Kanata.
Chandelier - Murasaki Shion

Both weren't working before and are now.

### DIFF
--- a/src/plugins/synced-lyrics/providers/LRCLib.ts
+++ b/src/plugins/synced-lyrics/providers/LRCLib.ts
@@ -11,10 +11,17 @@ export class LRCLib implements LyricProvider {
 
   async search({
     title,
+    alternativeTitle,
     artist,
     album,
     songDuration,
+    tags,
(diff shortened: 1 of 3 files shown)


JellyBrick (approved):
Thanks!

PR #3226, opened 2025-04-12
fix(mpris): keep MPRIS volume in sync with the actual volume

### LINKED ISSUE #3225: [Bug]: MPRIS: precise-volume: Volume property unsyncs if set through MPRIS
### Preflight Checklist

- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

master

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

Other Linux

### Operating System Version

NixOS 25.05 (Warbler) x86_64 [nixpkgs rev: github:NixOS/nixpkgs/2c8d3f4]

x64

### Last Known Working YouTube Music (Application) version

likely introduced in 9da0e43 (not entirely sure, I didn't bisect, just git blame)

### Reproduction steps

1. open youtube-music with **shortcuts** and **precise-volume** plugins
2. set volume through mpris (`playerctl --player=YoutubeMusic volume 0.5`)
3. get volume through mpris (`playerctl --player=YoutubeMusic volume`)
4. repeat steps 2 and 3 as many times as you want

### Expected Behavior

Each call to set the `volume` property through mpris should update the `volume` property in the mpris object.

So we should see something like:

```console
$ playerctl --player=YoutubeMusic volume 0.7
$ playerctl --player=YoutubeMusic volume
0.700000
$ playerctl --player=YoutubeMusic volume 0.2
$ playerctl --player=YoutubeMusic volume
0.200000
$ playerctl --player=YoutubeMusic volume 0.8
$ playerctl --player=YoutubeMusic volume
0.800000
```

### Actual Behavior

It doesn't update the mpris object, so when we ask the `volume` property, it returns an outdated value.

The volume in the actual player **does** update (in most cases*), it's just the mpris object that doesn't. So we should hear the volume changing in the application, but mpris says otherwise:

```console
$ playerctl --player=YoutubeMusic volume 0.7
$ playerctl --player=YoutubeMusic volume
1.000000
$ playerctl --player=YoutubeMusic volume 0.2
$ playerctl --player=YoutubeMusic volume
1.000000
$ playerctl --player=YoutubeMusic volume 0.8
$ playerctl --player=YoutubeMusic volume
1.000000
```

*: There's a case where it does **not** update the actual player's volume:

Let's suppose MPRIS volume froze on `1.0`, like in the case above, if we try to set the volume to 1.0 through mpris, it'll **not** update the actual player's volume.

### Enabled plugins

- shortcuts (mpris)
- precise-volume

### Additional Information

Another side consequence is that this doesn't work
```sh
playerctl --player=YoutubeMusic volume 0.1+
```
(it gets the current volume and adds 0.1 to it)

h-banii:
Fixes #3225 by reverting some of the modifications from 9da0e43

The code introduced in that commit was meant to prevent a supposed recursion described in https://github.com/th-ch/youtube-music/pull/1005#issuecomment-1415717977, but I don't think it happens anymore (at least not in the current codebase, maybe it did in the past)

<hr>

Now we can call this multiple times and it'll increment the volume properly (+10% volume per call)
`playerctl --player=YoutubeMusic volume 0.1+`

Before, it would always get the same frozen volume and add 0.1 to it, so it would stop incrementing after the first call.

### DIFF
--- a/src/plugins/shortcuts/mpris.ts
+++ b/src/plugins/shortcuts/mpris.ts
@@ -305,31 +305,14 @@ function registerMPRIS(win: BrowserWindow) {
       console.trace(error);
     });
 
-    let mprisVolNewer = false;
-    let autoUpdate = false;
     ipcMain.on('ytmd:volume-changed', (_, newVol) => {
-      if (~~(player.volume * 100) !== newVol) {
-        if (mprisVolNewer) {
-          mprisVolNewer = false;
-          autoUpdate = false;
-        } else {
-          autoUpdate = true;
-          player.volume = Number.parseFloat((newVol / 100).toFixed(2));
-          mprisVolNewer = false;
-          autoUpdate = false;
-        }
-      }
+      player.volume = Number.parseFloat((newVol / 100).toFixed(2));
     });
 
     player.on('volume', (newVolume: number) => {
       if (config.plugins.isEnabled('precise-volume')) {
         // With precise volume we can set the volume to the exact value.
-        const newVol = ~~(newVolume * 100);
-        if (~~(player.volume * 100) !== newVol && !autoUpdate) {
-          mprisVolNewer = true;
-          autoUpdate = false;
-          win.webContents.send('setVolume', newVol);
-        }
+        win.webContents.send('setVolume', ~~(newVolume * 100));
       } else {
         setVolume(newVolume * 100);
       }


JellyBrick:
Thanks!

PR #3242, opened 2025-04-15
Add Slack Now Playing plugin

curtisgibby:
I wanted to be able to push my "Now Playing" status to my company's Slack.


JellyBrick:
It would be better to separate the Slack scrobbler as a standalone plugin, similar to the Discord RPC. The Scrobbler plugin is intended for services like LastFM or Listenbrainz.

ArjixWasTaken:
This is more akin to discord rich presence than scrobbling.
Maybe we could make a more generic plugin that supports multiple social media platforms?

curtisgibby:
OK, I'll rework this as a standalone plugin.

ArjixWasTaken:
I wouldn't recommend adding a dependency like axios, especially in 2025, when the built-in fetch is as good.

ArjixWasTaken:
Why the extra dependency for formdata?

curtisgibby:
I've refactored this into its own plugin, using `fetch`. Ready for review.

src/plugins/slack-now-playing/main.ts
```diff
@@ -368,9 +362,9 @@ async function uploadEmojiToSlack(songInfo: SongInfo, config: SlackNowPlayingCon
       // If we got here, the request was successful
       // Emoji uploaded successfully
       return true;
-    } catch (apiError) {
+    } catch (apiError: any) {
```
ArjixWasTaken: instead of `any` use `unknown`, that's your intended use-case either way, since you are trying to narrow down the type later

curtisgibby: Refactored to `unknown` in a8350ecfaa0d9caff07db2ed062ff50e0de97707.

curtisgibby:
> Why the extra dependency for formdata?

That was a leftover when I was trying formdata instead of `formdata-node`. Removed now.

src/plugins/slack-now-playing/main.ts
```diff
@@ -0,0 +1,807 @@
...
+      return false;
+    }
+
+    // Prepare the form data for the API request using formdata-node
+    const formData = new FormData();
+    formData.append('mode', 'data');
+    formData.append('name', config.emojiName);
+    try {
+      // Use fileFromPath to get a proper File object for formdata-node
+      const imageFile = await fileFromPath(filePath);
```
ArjixWasTaken: here

package.json
```diff
@@ -273,7 +273,10 @@
     "es-hangul": "2.3.3",
     "fast-average-color": "9.5.0",
     "fast-equals": "5.2.2",
+    "fetch-blob": "^4.0.0",
+    "file-type": "^20.5.0",
     "filenamify": "6.0.0",
+    "formdata-node": "^6.0.3",
```
ArjixWasTaken: here

package.json
```diff
@@ -273,7 +273,10 @@
     "es-hangul": "2.3.3",
     "fast-average-color": "9.5.0",
     "fast-equals": "5.2.2",
+    "fetch-blob": "^4.0.0",
```
ArjixWasTaken: here as well?

ArjixWasTaken:
why so many polyfills?
we are using a modern verison of Nodejs that supports all those out of the box

src/plugins/slack-now-playing/main.ts
ArjixWasTaken: ok, this ain't a built-in, but you don't need an entire library for this
you can write 3 lines of code

src/plugins/slack-now-playing/main.ts
```diff
@@ -0,0 +1,814 @@
...
+      const fileBuffer = await fs.promises.readFile(filePath);
+      const filename = path.basename(filePath) || 'emoji.png';
+      let imageFile;
+      if (typeof File !== 'undefined') {
+        imageFile = new File([fileBuffer], filename);
+      } else if (typeof Blob !== 'undefined') {
+        imageFile = new Blob([fileBuffer]);
+      } else {
+        throw new Error('Neither File nor Blob is available in this environment');
+      }
```
ArjixWasTaken: who hurt you so much that you are checking for compatibility like that?
we ship the environment to the user, if the user tweaks the environment to use a different electron/node version, we don't have to support them

that's the benefit and drawback of using electron, we ship an entire browser to the end user
and it's even crazier when you consider that electron is probably using 2 v8 engines, one for chromium and one for node

ArjixWasTaken: all you need to check, when dealing with built-ins, is if it works on your machine

ArjixWasTaken: PS: If you don't want me to comment on your PR, just say it, I won't hold hard feelings against you.

curtisgibby: How would you resolve this? Just remove the last `else` block? I'm not much of a TypeScript dev, I'm mostly just relying on Cascade. :laugh:

ArjixWasTaken: I'd directly use the File built-in, and remove the fallbacks

ArjixWasTaken: > I'm mostly just relying on Cascade. :laugh:

It's honestly fine, your PR feels nothing like AI slop, I can actually see your effort!

curtisgibby: Updated based on suggestion in 5d4c531ed9eb5ae76bb207c2f4e8e322452e3b26

curtisgibby:
I believe that I've addressed all of the feedback that you've given. If there's something I've missed, or a new issue, please let me know.

ArjixWasTaken:
It would be hypocritical of me to say this in another PR and not here as well, so here we go.

https://github.com/th-ch/youtube-music/pull/3376#issuecomment-2888974155
> Why are you providing translations?
Translations are meant to be contributed using [weblate](https://hosted.weblate.org/projects/youtube-music/).

Please do not use AI/Google Translate for translations.

curtisgibby:
> Please do not use Google Translate/AI for translations.

Reverted all languages except EN & ES (since I speak those languages and know that the meaning is correct).

curtisgibby:
I've updated this PR with the latest commits from master. Is there any interest in merging it?

curtisgibby:
I've finally gotten some free time to rework this plugin and I think it's ready for prime time. Thanks for your patience on this one!

ArjixWasTaken:
> Thanks for your patience on this one!

I'd like to thank our users for *their* patience, since the entire project is moving really slowly :pensive:
Don't be in any hurry :^)

PR #3277, opened 2025-04-22
fix cuted "j" and glow in lyrics

AARP41298:
The glow uses 1.5rem
Before:
image
After:
image


src/plugins/synced-lyrics/style.css
```diff
@@ -73,6 +73,7 @@
 
   & .text-lyrics {
     cursor: pointer;
+    padding-left: 1.5rem;
```
JellyBrick, line 77: Could you please add a comment?

JellyBrick (changes requested):

src/plugins/synced-lyrics/style.css
AARP41298: done

JellyBrick (approved):

JellyBrick:
Thanks for your contribution!

PR #2831, opened 2025-01-05
Add custom equalizer plugin with presets (bass booster, rock...) and the ability to add your own presets via a config.json file

ZDAVanO:
This PR adds a new Custom Equalizer plugin for EQ settings with a large selection of default presets, and the ability to make your own flexible presets with detailed settings and any number of BiquadFilter. There is no UI, custom presets are added via a config.json file.

image

Issue #2186, opened 2024-06-28
[Feature Request]: Support Audio Transposer In-App


defaultkavy:
### Preflight Checklist

- [X] I use the latest version of YouTube Music (Application).
- [X] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

For singer, we usually adjust the pitch of the song to match our range. Before using this app, I always using [chrome extension](https://transpose.video) to change the pitch level of the song in real time. This is a really cool music player desktop app that what I want, and I really need this feature!

### Proposed Solution

A Real-time transpose & pitch level changer, in-app, with the playing song.

### Alternatives Considered

Here is the [chrome extension](https://transpose.video) that achieved real-time audio pitch level changer, so I think this feature is achievable.

defaultkavy:
Sorry for my bad English😓

PR #3291, opened 2025-04-26
feat(plugin): Pitch Shift

kimjammer:
## Pitch Shift

Implements shifting the pitch of the current song, like the Transpose extension.

I did what I could with my limited experience in this codebase, and the plugin is technically complete. But...

### Help Wanted:

1. Better UI: Currently the slider is implemented like playback-speed, but it would probably be beneficial to have a bigger slider since it's kinda hard right now to go up/down 1 semitone.
2. ~~Cent-wise control: Currently you can make sub-semitone changes with the scroll wheel that moves in increments of 0.01, but you might want that to move in units of cents instead.~~ Probably not needed.
3. Unified audio pipeline control: As currently implemented, you probably cannot have pitch-shift and the equalizer #2831 active at the same time (cuz I stole code from that pr).
4. Per-song memory of pitch adjustment.

Related: #2005 #2186 #3168 

**Off-topic:**
This took me waaaaay too long to implement, because I just could not find a good library for this until I used chatgpt deep research to find this port of rubberband with 33 weekly downloads :skull:. And then it took me another eternity trying to get the renderer to load the script locally and not from the internet. But it's done and it works!


src/plugins/pitch-shifter/rubberband-processor.js
JellyBrick, line 1: Is there no other way instead of hardcoding the code into the project?

kimjammer: To create an audioworklet, you must provide a URL to the location of the script. (Because this feature was designed for websites). Since we're in an electron app, we can't host the file. Instead, I use vite to import the file as a string, then turn it into a dataURI that the audioworklet API will accept. (using vite's file?url scheme doesn't work for some reason)

I think this will probably be fine since the original C++ rubberband library and this WASM version has been stable and is unlikely to need updates.

Alternatively, you could try importing the file directly from node_modules but that feels wrong.

kimjammer: I am very open to improvements, this is just how I got it to work. There may be better ways to do this with vite magic, but at the end of the day, we still have to ship the user the WASM code, and they still need to load it into memory to create the audioworklet, so I didn't think too hard about it.

Glebsin:
How can i easily download and install it right now? And will it work on the current version v3.9.0 ?

kimjammer:
@Glebsin This PR was last updated just after 3.9.0 released, so if you go to this branch on my fork, download, and build the project you could use it right now. But if you're not already a developer this is not worth the hassle of doing imo. So probably just wait until this is merged... eventually.

JellyBrick:
Since Rubberband uses GPL 2.0, we cannot merge this PR. Please use another library

kimjammer:
For tracking purposes, all libraries that have been tested/attempted in order of quality:

- rubberband-web
- soundtouchjs-audio-worklet
- olvb/phaze
- Tone.js

kimjammer:
The library has been switched to use SoundTouchJS, which has slightly inferior audio quality but is "good enough". SoundTouch is LGPL, but I believe that this usage follows the spirit of LGPL, as youtube-music connects with SoundTouch as a library and no modifications to the SoundTouch code have been made. If this is still a problem, you could theoretically just download the worker file from a web url at first launch or fetch the file from a web url every time to follow the letter of the license.

kz-n:
is this still being worked on?

kimjammer:
> is this still being worked on?

This PR is functionality complete and currently awaiting maintainer review.

kimjammer:
Please lmk if there's any of these that you want me to take a look at specifically.

PR #3296, opened 2025-04-27
[Bug fix] full screen video player cutoff from bottom

vipulkmr02:
Hi, I noticed that there is some video cutoff from the bottom in full screen player.
So, I fixed it

Before:
image
noticed there's some gap at the bottom.

After:
image
here's not that gap.

Thank you!

### DIFF
--- a/src/plugins/music-together/index.ts
+++ b/src/plugins/music-together/index.ts
@@ -68,8 +68,10 @@ export default createPlugin<
     me?: Omit<Profile, 'id'>;
     profiles: Record<string, Profile>;
     permission: Permission;
-    videoChangeListener: (event: CustomEvent<VideoDataChanged>) => void;
-    videoStateChangeListener: () => void;
+    videoChangeListener: (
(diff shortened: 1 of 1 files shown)


src/youtube-music.css
```diff
@@ -79,9 +79,10 @@ tp-yt-paper-item.ytmusic-guide-entry-renderer::before {
   max-width: calc(100% - var(--ytmusic-player-page-vertical-padding) * 2);
 }
 
-ytmusic-player[player-ui-state='FULLSCREEN'] {
-  margin-top: calc(var(--menu-bar-height, 32px) * -1) !important;
-}
+/* this is causing Video cutoff from bottom on fullscreen mode */
```
JellyBrick, line 82: It recreates another bug.

test-results/.last-run.json
```diff
@@ -0,0 +1,6 @@
+{
```
JellyBrick, line 1: What is this file?

JellyBrick (changes requested):

Issue #3232, opened 2025-04-14
[Bug]: Discord Rich Presence doesn't update on autoplay (shows wrong song on Discord)


Hex34568:
### Preflight Checklist

- [x] I use the latest version of YouTube Music (Application).
- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.8.0.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 10 IOT ENTERPRISE LTSC 21H2

### What CPU architecture are you using?

x64

### Reproduction steps

Sometimes when using YouTube with autoplay on, the Discord Rich Presence doesn’t update properly when a new song starts.
Even though a different song is playing, Discord still shows the previous song's title and metadata.

This seems to happen randomly. After another song plays, it usually fixes itself and starts updating correctly again.

Steps to reproduce (sometimes):

1. Play a YouTube video (music)

2. Let it autoplay to the next song

3. Check Discord Rich Presence – it still shows the old song info

4. Wait for the next autoplay – it might fix itself

### Expected Behavior

Discord Rich Presence should update with the correct song info every time a new video plays, including autoplay.

### Actual Behavior

Sometimes it doesn't update after autoplay. Discord shows incorrect song info.

### Enabled plugins

Only the presence plugin everything else is at default.

### Additional Information

This bug is inconsistent and doesn’t happen every time. When it does, it usually resolves after another autoplay.

Issue #3017, opened 2025-02-23
[Bug]: Discord rich prescence shows music is running even if it's paused


Megashyam2K9:
### Preflight Checklist

- [x] I use the latest version of YouTube Music (Application).
- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.7.4

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 11 Pro 24H2

### What CPU architecture are you using?

x64

### Reproduction steps

Turn on discord rich prescence
once you pause the music, discord rich prescence doesnt stop, it just keeps continuing

![Image](https://github.com/user-attachments/assets/68714bdd-09c1-4946-b1e8-c5a7f0f4593f)

### Expected Behavior

discord rich prescence also should pause

### Actual Behavior

once you pause the music, discord rich prescence doesnt stop, it just keeps continuing

### Enabled plugins

![Image](https://github.com/user-attachments/assets/de113c85-6c4b-488c-a925-0ed663007c90)

JellyBrick:
This issue is caused by Discord automatically updating its progress.

wienieclaurowy:
It used to work tho, thats the thing i have the same issue, and it used to work before but after few updates it broke, it still sometimes works after restart but breaks again after minute or two.

Cacapiston:
Idk if it will suit you, but you can try to set inactivity timeout to 10 seconds or something like that.

PR #3335, opened 2025-05-08
feat(discord): Improve activity updates with throttling and smarter state checks

Sato-Isolated:
- Implemented throttling to comply with Discord rate limits and reduce redundant updates.
- Added detailed state management: handles song changes, pause/resume states, and elapsed time tracking.
- Enhanced state verification before pushing updates: checks for video ID, paused state, and change deltas.
- Ensured elapsed time is only updated when needed to avoid noisy presence updates.

Fixes: #3232, #3017


Sato-Isolated:
The paused state is now correctly reflected in the Rich Presence, even though Discord still shows the green progress bar due to its limitations.

Not perfect visually, but probably the best we can do given Discord's current behavior.

image
image

PR #3358, opened 2025-05-11
feat(scrobblers): Only broadcast the main artist to Last.fm

SuchBlue:
Songs on Last.fm only have the main artist set, and including any featured artists hinders Last.fm's song detection

image
image

Despite a song having multiple artists, Last.fm only considers the main artist as the "Artist" of the song.


angelajfisher:
It looks like this would fix #3641. What still needs to be done to get it over the finish line? Happy to lend a hand if needed ☺️

PR #3366, opened 2025-05-13
fix: use gtk 3 switch as workaround

franz-dc:
The latest version of Electron will throw out an error with GNOME:

```bash
Running (process:10985): Gtk-ERROR **: 16:23:58.349: GTK 2/3 symbols detected. Using GTK 2/3 and GTK 4 in the same process is not supported
```

This commit applies the workaround mentioned [here](https://github.com/electron/electron/issues/46538#issuecomment-2808806722) and [here](https://github.com/electron/electron/blob/main/docs/breaking-changes.md#changed-gtk-4-is-default-when-running-gnome).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -114,6 +114,9 @@ protocol.registerSchemesAsPrivileged([
   { scheme: 'mailto', privileges: { standard: true } },
 ]);
 
+// https://github.com/electron/electron/issues/46538#issuecomment-2808806722
+app.commandLine.appendSwitch('gtk-version', '3');
+
 // Ozone platform hint: Required for Wayland support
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -114,6 +114,11 @@ protocol.registerSchemesAsPrivileged([
   { scheme: 'mailto', privileges: { standard: true } },
 ]);
 
+// https://github.com/electron/electron/issues/46538#issuecomment-2808806722
+if (process.platform === 'linux') {
```
JellyBrick:
```suggestion
if (is.linux()) {
```

JellyBrick (changes requested):

JellyBrick (approved):

PR #3381, opened 2025-05-19
chore(ui): Enhance transparency of main window and adjust background …

DFanso:
…styles

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -352,7 +352,8 @@ async function createMainWindow() {
     icon,
     width: windowSize.width,
     height: windowSize.height,
-    backgroundColor: '#000',
+    transparent: true,
+    backgroundColor: '#00000000',
     show: false,
     webPreferences: {
       contextIsolation: true,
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -352,7 +352,8 @@ async function createMainWindow() {
     icon,
     width: windowSize.width,
     height: windowSize.height,
-    backgroundColor: '#000',
+    transparent: true,
```
JellyBrick, line 355: But why? Original YTM isn't transparent.

src/youtube-music.css
```diff
@@ -92,3 +92,16 @@ tp-yt-iron-dropdown,
 tp-yt-paper-dialog {
   app-region: no-drag;
 }
+
+/* ---- App Customizations ---- */
```
JellyBrick, line 96: Are you using an AI like Cursor or Copilot in this PR?

JellyBrick (changes requested):

src/index.ts
DFanso: Just feel good for me 🙃

src/youtube-music.css
DFanso: Yep

ArjixWasTaken:
This should not be a default, you can't force such a GPU heavy change on everyone

JellyBrick:
Keep tracked at #3529

PR #3440, opened 2025-06-03
feat(api-server): add optional params for search

franz-dc:
This PR adds optional params to the `/search` API.
- `params`: More results from each category (songs, videos, artists, etc.)
- `continuation`: Continuation of each category (lazy loading)

### DIFF
--- a/src/providers/song-controls.ts
+++ b/src/providers/song-controls.ts
@@ -134,16 +134,12 @@ export default (win: BrowserWindow) => {
     },
     clearQueue: () => win.webContents.send('ytmd:clear-queue'),
 
-    search: (params: {
-      query: string;
-      params?: string;
-      continuation?: string;
-    }) =>
+    search: (query: string, params?: string, continuation?: string) =>
       new Promise((resolve) => {
         ipcMain.once('ytmd:search-results', (_, result) => {
           resolve(result as string);
         });
-        win.webContents.send('ytmd:search', params);
+        win.webContents.send('ytmd:search', query, params, continuation);
       }),
   };
 };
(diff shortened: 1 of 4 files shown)


src/providers/song-controls.ts
```diff
@@ -134,12 +134,16 @@ export default (win: BrowserWindow) => {
     },
     clearQueue: () => win.webContents.send('ytmd:clear-queue'),
 
-    search: (query: string) =>
+    search: (params: {
```
JellyBrick: Use parameters instead of the object (for the CLI)

src/renderer.ts
```diff
@@ -279,22 +279,23 @@ async function onApiLoaded() {
...
-      },
-    ) => {
+    async (_, query: string, params?: string, continuation?: string) => {
       const app = document.querySelector<YouTubeMusicAppElement>('ytmusic-app');
       const searchBox =
         document.querySelector<SearchBoxElement>('ytmusic-search-box');
 
       if (!app || !searchBox) return;
 
+      console.log({
```
JellyBrick: Please remove this console.log

JellyBrick (changes requested):

src/renderer.ts
franz-dc: My bad I forgot to remove that. Thanks!

JellyBrick (approved):

CustardFlan:
> This PR adds optional params to the `/search` API.
> 
> * `params`: More results from each category (songs, videos, artists, etc.)
> * `continuation`: Continuation of each category (lazy loading)

Thank you for the PR!
But, i want to know, what are such parameters to pass?
When i do a search, i want only songs as the result, sometimes a song it's the name of the album too, and, i want only to fetch the result and play the first one
Thank you!

PR #3483, opened 2025-06-13
fix(style): fix duplicated scrollbar

### LINKED ISSUE #3039: [Bug]: duplicated scrollbar
### Preflight Checklist

- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

So basically we have a scroll bar right ?
and it works right ?
but does it looks visually appealing ? No right ?
That's what I am saying 

Image
Image
Image
Image

### Proposed Solution

I have a suggestion that I hope you will consider. The current scrollbar is not visually appealing, and I find it unnecessary. I believe that the same functionality can be achieved with a more modern, smaller scrollbar.

this is the current one
Image

and this is the simple modern sample [ we all know ]
Image

### Alternatives Considered

.

### Additional Information

_No response_

h-banii:
Closes #3039

This PR just hides the body's overflow (catch-all solution to never show the double scrollbars)

|before|after|
|-|-|
|ytm-before|ytm-after|

<hr>

The bug only seems to happen when:
-  **in-app menu** is enabled
- "notification-action" popup is triggered (`save to playlist`, `add to liked songs`, etc)

I think it has something to do with the top margin added by the **in-app menu**
(plus the 1px added by the notification container)

<hr>

Alternative solution. This targets the notification container specifically.

```css
ytmusic-notification-action-renderer {
  position: fixed;
  z-index: 103; /* magic number from `tp-yt-paper-toast`, but anything >4 is fine */
  /* z-index is needed because otherwise it gets hidden under the left drawer menu */
}
```

Or even

```css
ytmusic-notification-action-renderer {
  position: absolute;
}
yt-notification-action-renderer {
  width: 0 !important; /* the width is set to 1px, and it somehow causes vertical scroll to appear */
  /* height is also set to 1px, but just the width is enough */
}
```

<hr>

> [!NOTE] 
> Interestingly, the double scrollbars don't go away when you close the notification popup... 
> 
> Youtube just creates a node for each notification in the DOM tree, then hides it with css and never removes it.
>
> It's a bit concerning in terms of memory, but that's a separate issue lol.

### DIFF
--- a/src/youtube-music.css
+++ b/src/youtube-music.css
@@ -96,3 +96,8 @@ tp-yt-iron-dropdown,
 tp-yt-paper-dialog {
   app-region: no-drag;
 }
+
+/* Fix duplicated scrollbar */
+body {
+  overflow: hidden !important;
+}


ArjixWasTaken:
> Youtube just creates a node for each notification in the DOM tree, then hides it with css and never removes it. <br>
> It's a bit concerning in terms of memory, but that's a separate issue lol.

Holy shit, we should probably replace their notification system entirely.

src/youtube-music.css
```diff
@@ -96,3 +96,8 @@ tp-yt-iron-dropdown,
 tp-yt-paper-dialog {
   app-region: no-drag;
 }
+
+/* Fix duplicated scrollbar */
+body {
+  overflow: hidden !important;
```
ArjixWasTaken: I'd prefer `overflow-y` since this is for the Y axis, but if someone gets a horizontal scrollbar then ngl that's their problem :rofl:

src/youtube-music.css
```diff
@@ -96,3 +96,8 @@ tp-yt-iron-dropdown,
 tp-yt-paper-dialog {
   app-region: no-drag;
 }
+
+/* Fix duplicated scrollbar */
+body {
```
JellyBrick: I'm worried that this patch might introduce unintended side effects.

h-banii: yeah, fair

youtube seems to be setting `body.style = "overflow: hidden/visible;"` through javascript to dinamically (un)hide the scrollbars when the user goes from the player to home and vice-versa.

Also, I just noticed this could conflict with the `unobtrusive player` plugin too:
https://github.com/th-ch/youtube-music/blob/40429034b6c5f3a2f0a3f480cf3b7a6be5ef86ce/src/plugins/unobtrusive-player/style.css#L1-L3

h-banii:
Since the problem is mainly caused by **in-app-menu's** `margin-top`, I decided to just fix it by using `padding-top` instead.

This seems to fix the issue.

PR #3529, opened 2025-06-25
feat(transparent-player): new plugin for Acrylic, Mica or Tabbed effects

### LINKED ISSUE #3629: [Feature Request]: Acrylic app backgorund
### Preflight Checklist

- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Add Acrylic background option for people wishing to use it (like me). It helps with system consistency if the other apps in the system have transparency. If the custom CSS is not applied, the app will still look normal.

_Homepage_

_Player Bar + Player Page_

### Proposed Solution

In index.ts, `const win = new BrowserWindow({})` starting line 359:
remove line 363: `backgroundColor: '#000',`
add: `backgroundMaterial: 'acrylic', transparent: 'true'

Use custom CSS: (extension changed to txt for upload)
[transparent background.txt](https://github.com/user-attachments/files/21224790/transparent.background.txt)
This a modified version of the CSS style from the extension for Zen Browser, [Zen Internet](https://www.sameerasw.com/zen). It changes a few other aspects of the player to make it look something like this:

### Alternatives Considered

- Leave it alone and build a custom version for myself when a new YTM version is released

### Additional Information

_No response_

frostybiscuit:
# Transparent Player plugin

## Options
- Opacity
    - Allowing to set the opacity of the apps background-color from 10-100%
- Type
    - Allows the user to select effects from "Acrylic", "Mica" and "Tabbed" without the need of reloading the app

## Known limitations
- As of now this will only work for Windows 11 22H2 and above. Support for MacOS is possible, but I don't have a device to test this on.
- The transparency for Acrylic is only visible when the window is currently in focus. This is a limitation by Windows and can only be "fixed" by using a third party application like "Mica for Everyone".

#### Acrylic (10%)
Acrylic_10%
#### Acrylic (50%)
Acrylic_50%
#### Acrylic (100%) - basically fully opaque
Acrylic_100%
#### Mica
Mica
#### Tabbed
Tabbed
#### Disabled
Disabled

## Notes
- The plugin can be fully disabled, but the initial settings need to be set to an active material in order for it to work properly (I chose Mica for least intrusion, but it should never be visible to the user unless they enable this plugin).
- I had to set the initial "backgroundColor" for the electron window instance to a transparent black for the effects to function, because this cannot be changed properly during runtime. The transparent background will never be visible without my plugin, since the website itself has a background-color set.

Closes https://github.com/th-ch/youtube-music/issues/3629


frostybiscuit:
@JellyBrick Not sure what went wrong with the build here, but considering it hit a timeout, I guess it could have been a hiccup during the build? ¯\_(ツ)_/¯

Could I kindly ask you to re-trigger the build? That is, if the whole concept of this plugin even matches with your vision for the app of course!

ArjixWasTaken:
@frostybiscuit

don't sweat it about the CI errors
also, there is no vision for the app, any idea that is shared by many people and doesn't negatively impact others will probably end up getting accepted

ArjixWasTaken:
> The plugin can be fully disabled, but the initial settings need to be set to an active material in order for it to work properly (I chose Mica for least intrusion, but it should never be visible to the user unless they enable this plugin).

I'd suggest marking the plugin as `restartNeeded: true`, and only add an initial active material if the plugin is enabled.

ArjixWasTaken:
Also, according to the docs, dynamically changing the material is exclusive to windows.
image

Dunno if the exclusivity also applies to the initial property of `backgroundMaterial`, but I also can't find it in the docs.

frostybiscuit:
@ArjixWasTaken Thanks for looking into it, very much appreciated!

I initially tried not having a backgroundMaterial set unless the plugin is initialised, but then the whole effect would not work at all, not even with restartNeeded set to true.

I would prefer this approach myself and tried around for a few days, but didn't find any working solution (other than loading a config before the initial window config) so I thought Mica being a native effect on W11 which can be disabled on OS level should have very little if any performance impact on the Users.

That said, I can continue to find a proper and clean solution to have unchanged initial settings until the plugin is enabled!

frostybiscuit:
Ignore this PR for now until I found a solution to not have any backgroundMaterial set initially, because I assume that there could be an issue on macOS running electron with this setting applied, even if only for the headless instance for testing.

ArjixWasTaken:
The config is available on `index.ts`
https://github.com/frostybiscuit/youtube-music/blob/feature/transparent-player-plugin/src/index.ts#L132-L138

you just have to write the json path

frostybiscuit:
@ArjixWasTaken I guess the test error was caused by the fact that the initial window options had a backgroundMaterial on macOS, even if the value was undefined. Now the builds run through, but in general the plugin only works on Windows for now.

I know that macOS also has a way to get this working, using "vibrancy", but I don't have a macOS device to test that on my end, so I wouldn't feel too comfortable testing this, but I'd be fine adapting the plugin itself accordingly to set the correct values.
As for Linux, I can't find a way to guarantee that it will work, especially since it is also distro dependent.

Would it be acceptable to have this as a Windows-only plugin for now, or should we first make sure that at least macOS is supported as well before continuing?

ArjixWasTaken:
On that topic, I'd like to make a suggestion

@JellyBrick currently, for platform-dependent plugins, we have an imho ugly way of disabling them on unsupported platforms.

e.g. https://github.com/th-ch/youtube-music/blob/master/src/index.ts#L64-L69

I suggest we introduce a way for a plugin to declare that it supports specific platforms.
e.g. using an enum

```typescript
export default createPlugin({
  platform: Platform.Windows | Platform.Darwin
});
```

In my opinion, declarative is better than imperative

---

@frostybiscuit it's totally fine to release your plugin for a specific platform, getting the plugin merged means that other people may contribute their own platform in a separate PR

frostybiscuit:
> I suggest we introduce a way for a plugin to declare that it supports specific platforms. e.g. using an enum
> 
> ```ts
> export default createPlugin({
>   platform: Platform.Windows | Platform.Darwin
> });
> ```

Fully agree that initializing plugins only if they support the current platform is supported makes most sense! For now I would maybe add "(Windows)" to the plugin label, unless you say that it isn't necessary.

Also, if you need a helping hand with the platform specific initialization, feel free to reach out, I would gladly help with that!

ArjixWasTaken:
> For now I would maybe add "(Windows)" to the plugin label, unless you say that it isn't necessary.

The plugin should be completely hidden from the user if their platform is not supported, as long as it's properly dealt with.
e.g. https://github.com/th-ch/youtube-music/blob/master/src/index.ts#L64-L69

frostybiscuit:
Gotcha! I will tackle this tomorrow

frostybiscuit:
@ArjixWasTaken The plugin is now fully removed on non-windows devices and the config is only touched after the plugin was enabled. Let me know if you want anything else adapted!

JellyBrick (approved):

Lyosey:
I really dont understand what im supposed to do to make the player transparent, do I download a file, is it an option im missing orr?

frostybiscuit:
@Lyosey It's a new plugin, tho currently only for Windows devices. So if you're on a Mac or on Linux, you won't see the plugin at all.

Lyosey:
@frostybiscuit Im on windows, I just dont get what im supposed to do, where is the plugin?

Decordy:
> @frostybiscuit Im on windows, I just dont get what im supposed to do, where is the plugin?

1st, check your updates. Make sure you're on latest version 3.11.0. Best way is to check [Releases](https://github.com/ytmd-devs/ytmd/releases) or have Auto-Update turned on in the Options tab.
2nd, this only works on Windows \_**_11_**\_ 22H2 and above, which saddens me tbh, what were Microsoft thinking...

Yilmaz4:
We need this in Linux Wayland ASAP

ArjixWasTaken:
> We need this in Linux Wayland ASAP

if you are on KDE, you can already do this by adding custom window rules

Yilmaz4:
> if you are on KDE, you can already do this by adding custom window rules

Wouldn't that make the entire window transparent, including the text and images, as opposed to just the background? I'm on Hyprland and it also has a similar feature, but that's not what I want

jeremaya123:
is there any way to prevent app from stopping transparent while being in other app?

LeviOP:
I don't really understand why this is windows-only. If you set `transparent: true` on the main window, this plugin works perfectly on linux (although it doesn't work when the devtools are open for some reason? idk)

frostybiscuit:
@LeviOP The reason for that is that I don't have a Linux machine to test this on and don't want to push stuff without proper testing. This is an open source project after all, so if you have the device to test this on, then please feel free to create a PR for Linux support.

frostybiscuit:
@jeremaya123 You could use [MicaForEveryone](https://github.com/MicaForEveryone/MicaForEveryone), select the process and enable "Blur behind" so it will always have a Mica or Acrylic effect, whatever you set it to.

Windhawk also has a mod called "Translucent Windows" which can do the same thing, in case you already use Windhawk for something else or prefer an alternative over MicaForEveryone.

jaynemaxwell:
> I don't really understand why this is windows-only. If you set `transparent: true` on the main window, this plugin works perfectly on linux (although it doesn't work when the devtools are open for some reason? idk)

yes this is a limitation on electron's side, and is a common occurrence in electron projects that allow for window transparency.
see https://www.electronjs.org/docs/latest/tutorial/custom-window-styles#limitations

LeviOP:
> yes this is a limitation on electron's side, and is a common occurrence in electron projects that allow for window transparency.

okay then it's not an issue with linux. then there really is no reason this is windows-only right now.

Astear17:
> @ArjixWasTaken I guess the test error was caused by the fact that the initial window options had a backgroundMaterial on macOS, even if the value was undefined. Now the builds run through, but in general the plugin only works on Windows for now.
> 
> I know that macOS also has a way to get this working, using "vibrancy", but I don't have a macOS device to test that on my end, so I wouldn't feel too comfortable testing this, but I'd be fine adapting the plugin itself accordingly to set the correct values. As for Linux, I can't find a way to guarantee that it will work, especially since it is also distro dependent.
> 
> Would it be acceptable to have this as a Windows-only plugin for now, or should we first make sure that at least macOS is supported as well before continuing?

Hey, I can do macOS testing if you need. Currently I found no option to toggle translucent feature.

PR #3580, opened 2025-07-08
Update README.md with link to live diagram

JetsettingJames:
Adding badge with link to https://repomapr.com/th-ch/youtube-music to help devs understand the repo.
image


ArjixWasTaken:
As a contributor to this project, I can confidently say that it is highly inaccurate (e.g. it says we use React when we use Solidjs), and honestly it's way more confusing than just reading the README and browsing the code.

Also, the "node"s themselves are defined in what appears to be random classification/order.
It also claims that "youtube-dl" is used in this project, which is blatanly wrong.

JetsettingJames:
@ArjixWasTaken thanks for your constructive feedback. If improved, would you see utility in this product?

ArjixWasTaken:
Personally? No
I have aphantasia, among other stuff, so I find it easier to just look at the code.

Would be interesting if it generated documentation, akin to [readthedocs](https://about.readthedocs.com/) and had a node-based explorer, where each node is an entry in the documentation.

But the accuracy of the data is important as well.

ArjixWasTaken:
Overall, th-ch/youtube-music is not a library, so AI generated documentation like this seems out of place.
Maybe if it could focus on specific parts (just like you would do on hand-written docs), and not the entire codebase, it would be better.

JetsettingJames:
@ArjixWasTaken Thanks. Have indexed accurately other proper libraries, so maybe that. Will iterate and ping once we feel we have something better.

SomeAspy:
link is dead

PR #3588, opened 2025-07-10
feat: enable the ESM for main

JellyBrick:


src/index.ts
```diff
@@ -74,9 +74,6 @@ unhandled({
   showDialog: false,
 });
 
-// Disable Node options if the env var is set
-process.env.NODE_OPTIONS = '';
```
MulverineX, line 78: Why was this removed?

JellyBrick: Are there any regressions caused by this change?

JellyBrick: @MulverineX Are there any regressions caused by this change?

MulverineX: @JellyBrick what I mean is, why was that code there in the first place, and what was the reasoning for removing it? Doesn't this make it a little easier to inject into the app?

PR #3592, opened 2025-07-10
feat(api-server): Improved api-server volume and like/dislike state

rewhex:
This PR includes a couple of small but useful improvements to the api-server:

_Volume State Endpoint:_
The volume state endpoint now includes a new `isMuted` boolean. It was a single `state` volume level before, now it also has `isMuted` property, to detect whether volume button is pressed (just what the `/v1/toggle-mute` does).

_New Like State Endpoint:_
Added a new endpoint: `GET /v1/like-state`
It returns the current song's like/dislike status, using the same values: `LIKE, DISLIKE, INDIFFERENT` as YouTube Music already has.

### DIFF
--- a/src/providers/song-info-front.ts
+++ b/src/providers/song-info-front.ts
@@ -79,12 +79,39 @@ export const setupRepeatChangedListener = singleton(() => {
   );
 });
 
+export const setupLikeChangedListener = singleton(() => {
+  const likeDislikeObserver = new MutationObserver((mutations) => {
+    window.ipcRenderer.send(
+      'ytmd:like-changed',
+      (mutations[0].target as HTMLElement).getAttribute('like-status'),
+    );
+  });
+  likeDislikeObserver.observe(document.querySelector('#like-button-renderer')!, {
+    attributes: true,
+    attributeFilter: ['like-status'],
+  });
+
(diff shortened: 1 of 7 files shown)


rewhex:
@JellyBrick hello, I've fixed copilot's comments, can you please merge this PR?

src/providers/song-info-front.ts
```diff
@@ -79,12 +79,48 @@ export const setupRepeatChangedListener = singleton(() => {
   );
 });
 
+const mapLikeStatus = (status: string | null): LikeType =>
+  Object.values(LikeType).includes(status as LikeType)
+    ? (status as LikeType)
+    : LikeType.Indifferent;
+
+export const setupLikeChangedListener = singleton(() => {
+  const likeDislikeObserver = new MutationObserver((mutations) => {
```
JellyBrick, line 90: Is the use of MutationObserver strictly necessary in this part?

rewhex: Hmm, I don't think we can trigger state change by something else. Like any other controls (volume for example) we can get the buttons state with this and show it to the user.

<details>
  <summary>Screenshot</summary>
  
</details>

Issue #2200, opened 2024-07-04
[Bug]: No Minimum Screen Size Limit


SubhamSubhasisPatra:
### Preflight Checklist

- [X] I use the latest version of YouTube Music (Application).
- [X] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [X] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.3.12.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [X] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 11

### What CPU architecture are you using?

x64

### Reproduction steps

1. Open the app, and try to resize the window from the corner. 
2. the issue is you will be able to resize till the navigation bar.

### Expected Behavior

While resize the app, the window should not allow to minimise more after a certain height and width.

### Actual Behavior

When u size the screen it is like a tiny screen.

![Screenshot 2024-07-04 113816](https://github.com/th-ch/youtube-music/assets/43811917/f0223bdb-9721-4e77-b6ae-e052f70e502a)

### Enabled plugins

Nothing

ArjixWasTaken:
Why is this an issue?

patYczakus:
@SubhamSubhasisPatra there's no limit in height - you can also make this height on any browser (for example Edge)
However @ArjixWasTaken, there is no limit in width, and as a result, the appearance becomes distorted and the buttons with _In-app menu_ plugin enabled overlap with standard buttons. This problem can be solved I think

> Too small width
> ![image](https://github.com/th-ch/youtube-music/assets/75316297/b418000a-c696-49a6-9016-4b2651d68cd1)

> Good width, but overlapping buttons
> ![image](https://github.com/th-ch/youtube-music/assets/75316297/c2faaa00-b8e7-412d-9e0c-9e3a59649ad8)

patYczakus:
Theory of relativity, somebody may take this as a issue, someone not
Even thought, better (for my opinion) add some limits, just in case

SubhamSubhasisPatra:
@ArjixWasTaken there should be some restriction on both height and width , if we are not doing that then the UI should be responsive, In my opinion overlapping of button is not a good thing. Even the screen is very small the close icon should not be overlapping.

[...]

PR #3602, opened 2025-07-12
Prevent UI breakage on small window sizes

Codewire-github:
**Changes**
- Added minWidth: 325 and minHeight: 425 to **BrowserWindow** in **_index.ts_** to avoid layout issues
- This ensures the Music controls are visible in smallest window possible

**Before:**
Screenshot 2025-07-12 at 1 02 33 PM

**After:**
Screenshot 2025-07-12 at 1 00 51 PM

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -343,8 +343,8 @@ async function createMainWindow() {
     titleBarStyle: useInlineMenu
       ? 'hidden'
       : is.macOS()
-        ? 'hiddenInset'
-        : 'default',
+      ? 'hiddenInset'
+      : 'default',
     autoHideMenuBar: config.get('options.hideMenu'),
   };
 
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
I am not gonna say that this is a bad change, but I am gonna say that it is a non-issue imo.
You aren't making the UI better at a smaller window size, you just prevent the window from being small in the first place.

So like, nothing really changed
(in the sense that nobody should be resizing their window to be that small in the first place)

--

PS: I am in favor of this getting merged, there has been [an issue](https://github.com/th-ch/youtube-music/issues/2200) about this for quite a while.

PS2: Add `Resolves: #2200` to the description of this PR

JellyBrick:
I don't want to fix this, because someone might want to make the window smaller.

(However, thanks for contributing!)

briankenwell:
I agree with @JellyBrick, I've previously made the window very small, like the iTunes/Apple Music MiniPlayer. The play/pause button and next track buttons were visible with the artist/song name. Would only resize the window when switching playlists, which I would typically only do once or twice a day at most.
Maybe this could be a plugin or preference rather than something not easily turned off?

PR #3606, opened 2025-07-12
Add Material UI 3 component library

ArjixWasTaken:


ArjixWasTaken:
Honestly, it might be better if we can somehow transcompile mdui to be in ES5 syntax.
But I don't really know if that would work :smiling_face_with_tear:

ArjixWasTaken:
Opened this for review, but I am unsure if it's actually ready to get merged.

src/renderer.ts
```diff
@@ -1,5 +1,10 @@
 import i18next from 'i18next';
 
+import 'mdui';
+import { setTheme } from 'mdui/functions/setTheme.js';
+
+setTheme('dark')
```
JellyBrick: Should `setTheme` be called here?

ArjixWasTaken: I am calling it there so it switches as fast as possible

PR #3618, opened 2025-07-13
feat(synced-lyrics): thai romanization

hax0r31337:
Support of Thai romanization from `@dehoist/romanize-thai`

image

### DIFF
--- a/src/plugins/synced-lyrics/renderer/utils.tsx
+++ b/src/plugins/synced-lyrics/renderer/utils.tsx
@@ -8,6 +8,8 @@ import hanja from 'hanja';
 
 import pinyin from 'tiny-pinyin';
 
+import { romanize as romanizeThai } from '@dehoist/romanize-thai'
+
 import { lazy } from 'lazy-var';
 
 import { detect } from 'tinyld';
@@ -91,8 +93,8 @@ export const canonicalize = (text: string) => {
 export const simplifyUnicode = (text?: string) =>
   text
     ? text
-        .replaceAll(/\u0020|\u00A0|[\u2000-\u200A]|\u202F|\u205F|\u3000/g, ' ')
-        .trim()
+      .replaceAll(/\u0020|\u00A0|[\u2000-\u200A]|\u202F|\u205F|\u3000/g, ' ')
+      .trim()
     : text;
 
 // Japanese Shinjitai
@@ -155,26 +157,9 @@ const hasKorean = (lines: string[]) =>
 const hasChinese = (lines: string[]) =>
   lines.some((line) => /[\u4E00-\u9FFF]+/.test(line));
 
-export const romanize = async (line: string) => {
-  const lang = detect(line);
-
-  const handlers: Record<string, (line: string) => Promise<string> | string> = {
-    ja: romanizeJapanese,
-    ko: romanizeHangul,
-    zh: romanizeChinese,
-  };
-
-  const NO_OP = (l: string) => l;
-  const handler = handlers[lang] ?? NO_OP;
-
-  line = await handler(line);
-
-  if (hasJapanese([line])) line = await romanizeJapanese(line);
-  if (hasKorean([line])) line = romanizeHangul(line);
-  if (hasChinese([line])) line = romanizeChinese(line);
-
-  return line;
-};
+// https://en.wikipedia.org/wiki/Thai_(Unicode_block)
+const hasThai = (lines: string[]) =>
+  lines.some((line) => /[\u0E00-\u0E7F]+/.test(line));
 
 export const romanizeJapanese = async (line: string) =>
   (await kuroshiro.get()).convert(line, {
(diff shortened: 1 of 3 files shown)


src/plugins/synced-lyrics/renderer/utils.tsx
```diff
@@ -190,3 +175,42 @@ export const romanizeChinese = (line: string) => {
...
+  ? new Intl.Segmenter('th', { granularity: 'word' })
+  : null;
+
+export const romanizeThai = (line: string) => {
+  if (!thaiSegmenter) return romanizeThaiFrag(line);
+
+  const segments = Array.from(thaiSegmenter.segment(line));
+  const latin = segments
+    .map((segment) => segment.isWordLike ? romanizeThaiFrag(segment.segment) + ' ' : segment.segment.trim())
+    .join('')
```
ArjixWasTaken:
```suggestion
    .map((segment) => segment.isWordLike ? romanizeThaiFrag(segment.segment) : segment.segment.trim())
    .join(' ')
```

src/plugins/synced-lyrics/renderer/utils.tsx
```diff
@@ -190,3 +175,42 @@ export const romanizeChinese = (line: string) => {
...
+
+  return latin;
+};
+
+const handlers: Record<string, (line: string) => Promise<string> | string> = {
+  ja: romanizeJapanese,
+  ko: romanizeHangul,
+  zh: romanizeChinese,
+  th: romanizeThai,
+};
```
ArjixWasTaken, line 198: does `detect(line)` not detect Thai?

hax0r31337: it does detecting Thai, i move it out of the function to stop the object getting created each call

hax0r31337: i have also found the romanizing library is quite large in code, what about moving the romanization hard lifting to the main process to reduce the size of the JavaScript injected to the renderer process?

ArjixWasTaken: That's how it was originally done, but Jellybrick decided to bring the romanization to the renderer, to which I do not disagree with.

Is there no lighter library to use?

hax0r31337: For Mandarin language, maybe it's a better idea to lazy-load a lookup json table instead of putting it in code.   
Here's a [lookup table](https://github.com/GitYCC/g2pW/blob/master/g2pw/char_bopomofo_dict.json) but it's character wise lookup so it couldn't handle [polyphonic characters](https://en.wikipedia.org/wiki/Chinese_character_sounds#Polyphonic_characters) as it's determined from context.

Edit: But it's possible to increase accuracy of polyphonic character romanization by selecting most used word (looking up words might not be practical as the dictionary file are too large)    
[a word dictionary with data from Ministry of Education of Republic of China processed to JSON format](https://github.com/g0v/moedict-data) for distilling polyphonic characters

hax0r31337: i also suggest to expand the language detection to entire lyrics instead of a single line   
as the language detection are mistaking japanese sentence with no kana to mandarin
image

ArjixWasTaken: I am talking about Thai

hax0r31337: The romanizing library for Thai is not large after all

ArjixWasTaken: > it does detecting Thai, i move it out of the function to stop the object getting created each call

nevermind, when I viewed the code from github mobile for some reason I did not see the `th: detectThai` line

ArjixWasTaken:
> i also suggest to expand the language detection to entire lyrics instead of a single line
as the language detection are mistaking japanese sentence with no kana to mandarin

that is a really hard problem to solve, since there are songs that mix japanese/korean/chinese

hax0r31337:
> > i also suggest to expand the language detection to entire lyrics instead of a single line
> > as the language detection are mistaking japanese sentence with no kana to mandarin
> 
> that is a really hard problem to solve, since there are songs that mix japanese/korean/chinese

well, I never thought there are cases like that

src/plugins/synced-lyrics/renderer/utils.tsx
```diff
@@ -190,3 +175,43 @@ export const romanizeChinese = (line: string) => {
...
+const thaiSegmenter = Intl.Segmenter.supportedLocalesOf('th').includes('th')
+  ? new Intl.Segmenter('th', { granularity: 'word' })
+  : null;
+
+export const romanizeThai = (line: string) => {
+  if (!thaiSegmenter) return romanizeThaiFrag(line);
+
+  const segments = Array.from(thaiSegmenter.segment(line));
+  const latin = segments
+    .map((segment) => segment.isWordLike ? romanizeThaiFrag(segment.segment) : segment.segment.trim())
```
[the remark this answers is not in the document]
ArjixWasTaken: 'cause word's can't have spaces in them, dumb AI

ArjixWasTaken:
> > > i also suggest to expand the language detection to entire lyrics instead of a single line
> > > as the language detection are mistaking japanese sentence with no kana to mandarin
> > 
> > 
> > that is a really hard problem to solve, since there are songs that mix japanese/korean/chinese
> 
> well, I never thought there are cases like that

One solution might be to detect what percentage of the lyrics is japanese, if it's above a threshold of like 95%, then we assume all chinese characters are actually kanji

PR #3627, opened 2025-07-14
fix(api-server): Fixes the duplicated clientId saved in authorizedClients

rewhex:
Trying to call /auth/{id} multiple times with the same id resulting in error 500 and duplicated ids in authorizedClients, this one fixes the issue.


JellyBrick (approved):

JellyBrick:
Thanks!

PR #3659, opened 2025-07-21
Fixed missing videochange dataupdated event when using shuffle

rewhex:
Fix issue where videochange-dataupdated event doesn't fire after shuffle

After triggering shuffle via the API, the videochange: dataupdated event sometimes fails to fire. To ensure the song info is still updated, a fallback using a timeout has been implemented.

### DIFF
--- a/src/providers/song-info-front.ts
+++ b/src/providers/song-info-front.ts
@@ -206,12 +206,30 @@ export default (api: YoutubePlayer) => {
     );
 
   const waitingEvent = new Set<string>();
+  const waitingTimeouts = new Map<string, NodeJS.Timeout>();
+
+  const clearVideoTimeout = (videoId: string) => {
+    const timeoutId = waitingTimeouts.get(videoId);
+
+    if (timeoutId) {
+      clearTimeout(timeoutId);
+      waitingTimeouts.delete(videoId);
+    }
+  };
+
+  const clearAllTimeouts = () => {
+    waitingTimeouts.forEach((timeoutId) => clearTimeout(timeoutId));
+    waitingTimeouts.clear();
+  };
+
   // Name = "dataloaded" and abit later "dataupdated"
+  // Sometimes "dataupdated" is not fired, so we need to fallback to "dataloaded"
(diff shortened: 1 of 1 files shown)


src/providers/song-info-front.ts
```diff
@@ -222,10 +242,23 @@ export default (api: YoutubePlayer) => {
...
       }
 
+      clearVideoTimeout(videoData.videoId);
       waitingEvent.add(videoData.videoId);
+
+      const timeoutId = setTimeout(() => {
+        if (waitingEvent.has(videoData.videoId)) {
+          waitingEvent.delete(videoData.videoId);
+          waitingTimeouts.delete(videoData.videoId);
+          sendSongInfo(videoData);
```
JellyBrick, line 247: If `dataupdated` is not fired, `videoData` may be incompletely loaded. In this case, the behavior of some plugins may be broken.

rewhex: Yet this is the best option, because without the fallback to first event the data isn't updated at all. Even if I press next song

rewhex: Seems like in the case when `dataupdated` isn't fired, `dataloaded` has all data just like the `dataupdated`, but someone else need to test and confirm it

src/providers/song-info-front.ts
```diff
@@ -222,10 +242,23 @@ export default (api: YoutubePlayer) => {
...
+        }
+      }, DATAUPDATED_FALLBACK_TIMEOUT_MS);
+
+      waitingTimeouts.set(videoData.videoId, timeoutId);
     }
   });
 
+  window.addEventListener('beforeunload', clearAllTimeouts, { once: true });
```
ArjixWasTaken: This seems redundant, doesn't the window automatically destroy all related resources when it unloads?

ArjixWasTaken: Or did I misunderstand this, and the event gets fired even when the window does not unload?

rewhex: Hmm, yeah, this one is redundant

PR #3665, opened 2025-07-22
refactor: PluginDefinition::platform

ArjixWasTaken:
Allow plugins to declare what platforms they support, streamlining the way platform-specific plugins are declared.

### DIFF
--- a/vite-plugins/plugin-importer.mts
+++ b/vite-plugins/plugin-importer.mts
@@ -139,10 +139,12 @@ export const pluginVirtualModuleGenerator = (
 function supportsPlatform({ platform }: { platform: string }) {
   if (typeof platform !== "number") return true;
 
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
@JellyBrick I am thinking of moving the logic directly to the virtual module that exports `allPlugins`, what do you think?

JellyBrick:
> @JellyBrick I am thinking of moving the logic directly to the virtual module that exports `allPlugins`, what do you think?

👍

ArjixWasTaken:
> > @JellyBrick I am thinking of moving the logic directly to the virtual module that exports `allPlugins`, what do you think?
> 
> 👍

done

ArjixWasTaken:
ok, actually done this time :sweat:

Issue #3316, opened 2025-05-01
[Feature Request]: Option to display artist in Discord Rich Presence instead of "YouTube Music"


yuckdevchan:
### Preflight Checklist

- [x] I use the latest version of YouTube Music (Application).
- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Would be cool to have an option to change the Discord RPC from showing "YouTube Music" to showing the artist that's being listened to.

Spotify recently made this change to their RPC and would be great addition to this app.

### Proposed Solution

Option to change the default behaviour to the following:

Example:

Listening to **The Beatles**

Instead of,

Listening to **YouTube Music**

Would be a great option!

### Alternatives Considered

Also having an option for displaying the song or the album would also be handy.

JellyBrick:
We are unable to implement this feature because Discord limits application names to one.

JellyBrick:
See https://github.com/discord/discord-rpc/issues/375

mudrhiod:
Isn't it possible to implement this now?
https://twitter.com/advaithj1/status/1946340808996839885

PR #3692, opened 2025-07-28
feat(discord): option to display artist/title in status

pynappo:
Related to: #3316, #3368 

Discord has [now made the API public for changing which activity field is featured in the status](https://github.com/discord/discord-api-docs/pull/7674) (as well as allowing each field of the activity to embed separate URLs).

The main thing is that we can now set the status to `Listening to {artist name}` and `Listening to {song title}`.

This PR:

- implements a configurable option for this
- sets the new default to be `Listening to {artist name}` (since it aligns w/ Spotify)
- also embeds the song url into the title

image

image

> [!NOTE]
> Feel free to comment on how the option menu labels are worded, i'm not sure if it's clear/intuitive.

## Misc

i saw there was inconsistent indentation in en.json so that's also fixed here, lmk if it should be separated.


ArjixWasTaken:
Looks like it's so new, that it's not added to the discordjs types yet.
https://github.com/discordjs/discord-api-types/issues/1306

foxfirecodes:
hey there, dev who built these features for rich presence here :3 while StatusDisplayType hasnt been added to discord-api-types, it _is_ available as an enum in the `@xhayper/discord-rpc` lib you're using. is there a reason to block this on being added to discord-api-types? it doesnt even make sense to me to use discord-api-types here since you're not using the rest api, you're using the rpc api (i'm aware they include rpc types, but id think you'd want to use the types from the rpc lib you're using, since that more accurately mirrors what the rpc lib is capable of)

ArjixWasTaken:
@foxfirecodes 

I don't think this PR is "blocked" in that sense, it just hasn't been merged yet.

Jellybrick usually merges PRs whenever he wants to make a new release, I assume that's the case here

PR #3701, opened 2025-07-31
feat(api-websocket)

Johannes7k75:
This PR adds an websocket for applications that would need realtime updates from youtubeMusic.

Users can configure the `hostname` and the `port` which is gona be used for the websocket server.

Data that is send:
- volume change 
- repeat change
- position change via seeking
- mute state change
- song state change (song change, pause, resume)

The user can also send actions:
- play
- pause
- next
- previous
- shuffle
- mute
- repeat (To which mode)
- seek (to position)
- getVolume
- setVolume

This PR also adds an new ytm protocol option:
to open an video like `youtubemusic://openVideo OgvsDHJr6G4`


Duplicake-fyi:
commenting to get updated when this is merged

ArjixWasTaken:
> commenting to get updated when this is merged

image

ArjixWasTaken:
This feature was bound to happen someday, and I've always had thoughts about it.
PS: I am just sharing my thoughts, at the end of the day I am just a contributor, not a maintainer.

I feel like having a more well-defined interface is important.

1. we already have an API plugin, and hono supports [websockets](https://www.npmjs.com/package/@hono/node-ws), so maybe tying them together makes sense?
2. we need a way for plugins to declare upfront what "events" they can publish, and what "actions" they provide.
3. the websocket is accessible to anyone on the network, there is no security in-place (back to point 1)

if you want an example for my earlier points, take the `synced-lyrics` plugin for example
it could send the lyrics over the websocket, so they can be used by an external program (or a popup window)
it could provide actions for changing between the lyrics providers, etc

rewhex:
Hello
Do you seen https://github.com/th-ch/youtube-music/pull/3592
I've did some changes to make mute/like state bi-directional and added current muted/like state to endpoints, maybe we need to combine PRs to improve sockets too, imo websockets are better because for now I continuously polling api plugin

Johannes7k75:
> Hello Do you seen #3592 I've did some changes to make mute/like state bi-directional and added current muted/like state to endpoints, maybe we need to combine PRs to improve sockets too, imo websockets are better because for now I continuously polling api plugin

Yeah i think i should just integrate the socket into the api-server, the actions it can could be removed because they are already provided by the api-server and therefore arent really needed.

PR #3707, opened 2025-08-02
feat(api-server): Add websocket as /ws route

Johannes7k75:
This pr adds an websocket

the data that is send includes:
- songInfo
- isPlaying
- muted
- position
- volume
- repeat

The values are send when they change.

Its based on #3529 so that would be merged first. It uses the volume-change from this PR for volume and muted.

I added also an comment for an implementation for an jwt auth based websocket. So that it could be easier hidden behind auth.

@rewhex i hope this is so ok?

### DIFF
--- a/src/plugins/api-server/backend/api-version.ts
+++ b/src/plugins/api-server/backend/api-version.ts
@@ -1,0 +1,1 @@
+export const API_VERSION = 'v1';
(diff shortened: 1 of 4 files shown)


rewhex:
Hey, you mentioned wrong PR :)
Have you added like/dislike event or I'm just blind?

Johannes7k75:
No i just added the volume state stuff (volume and isMuted).

PR #3737, opened 2025-08-10
feat(discord): add song & artist URLs to rich presence

foxfirecodes:
hey there, dev who implemented the [new rich presence fields](https://github.com/discord/discord-api-docs/pull/7674) here :3

this pr adds clickable artist & song links to the rich presence. i looked through the payload to see if there's a good way to get the album URL, but i couldn't find a field that is reliable. i thought maybe playlistId would work, but when listening to a playlist its the id of the playlist the song is part of rather than the album.

[Screencast_20250810_084314.webm](https://github.com/user-attachments/assets/3e086e8c-1050-4a82-b5f0-9932cce24ca2)

### DIFF
--- a/src/plugins/discord/discord-service.ts
+++ b/src/plugins/discord/discord-service.ts
@@ -1,4 +1,7 @@
-import { Client as DiscordClient } from '@xhayper/discord-rpc';
+import {
+  Client as DiscordClient,
+  StatusDisplayType,
+} from '@xhayper/discord-rpc';
 import { dev } from 'electron-is';
 import { ActivityType } from 'discord-api-types/v10';
 
@@ -98,8 +101,11 @@ export class DiscordService {
 
     const activityInfo: SetActivity = {
       type: ActivityType.Listening,
+      statusDisplayType: StatusDisplayType.STATE,
       details: truncateString(songInfo.title, 128), // Song title
+      detailsUrl: songInfo.url,
       state: truncateString(songInfo.artist, 128), // Artist name
+      stateUrl: songInfo.artistUrl,
       largeImageKey: songInfo.imageSrc ?? undefined,
       largeImageText: songInfo.album
         ? truncateString(songInfo.album, 128)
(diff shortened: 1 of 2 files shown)


src/plugins/discord/discord-service.ts
```diff
@@ -109,7 +111,7 @@ export class DiscordService {
 
     // Handle paused state display
     if (songInfo.isPaused) {
-      activityInfo.largeImageText = '⏸︎';
+      activityInfo.largeImageText = '⏸';
```
ArjixWasTaken: This unicode character makes the diff weird lol

to anyone curious, here is the actual diff
```
> const chars = [`⏸︎'`, `⏸'`]
> chars[0].split('')
[ '⏸', '', "'" ]
> chars[1].split('')
[ '⏸', "'" ]
```

looks like there was a no-width character previously, with the char-code `65038`

ArjixWasTaken (approved):
looks fine to me

src/plugins/discord/discord-service.ts
foxfirecodes: im not sure why my editor insisted on changing that when i saved, but just in case there was some significance to the zero width char ive reverted that one line

src/providers/song-info.ts
```diff
@@ -93,6 +95,7 @@ const handleData = async (
     songInfo.url = microformat.urlCanonical?.split('&')[0];
     songInfo.playlistId =
       new URL(microformat.urlCanonical).searchParams.get('list') ?? '';
+    songInfo.artistUrl = `https://music.youtube.com/channel/${microformat.pageOwnerDetails.externalChannelId}`;
```
JellyBrick:
```suggestion
    if (microformat.pageOwnerDetails?.externalChannelId) {
      songInfo.artistUrl = `https://music.youtube.com/channel/${microformat.pageOwnerDetails.externalChannelId}`;
    }
```

foxfirecodes: the nullability of these fields is so confusing to me since the typescript types make it seem like all the sub objects are always present 😅

PR #3766, opened 2025-08-19
fix: move sendSongInfo to dataloaded event

### LINKED ISSUE #3733: [Bug]: (API Server) /song endpoint does not update properly if an album is played
### Preflight Checklist

- [x] I have searched the [issue tracker](https://github.com/th-ch/youtube-music/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **th-ch/youtube-music has NO affiliation with Google or YouTube**

### YouTube Music (Application) Version

3.10.0

### Checklists

- [ ] I use the portable version of the YouTube Music Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

Other Linux

### Operating System Version

Fedora 42 (GNOME, Flatpak)

x64

### Last Known Working YouTube Music (Application) version

_No response_

### Reproduction steps

Control:
1. Play a song (`Song A`)
2. Hit `/api/v1/song`
3. Response with `Song A` as currently playing song

Issue:
1. Play via clicking an album's play button (containing `Song B`)
2. Hit `/api/v1/song`
3. Response with `Song A` as currently playing song

### Expected Behavior

Response should be the currently playing song 

### Actual Behavior

Response is still the previous song, with `elapsedSeconds` stuck to when it stopped playing. 

### Enabled plugins

- API Server
- Exponential Volume
- Precise Volume
- SponsorBlock
- Unobtrusive Player
- Video Toggle

### Additional Information

Once you play a song again or let the next song play within the album, API response is as expected again.

franz-dc:
Fixes #3733

The event listener callback for `videodatachange` assumes that `dataloaded` comes first then `dataupdated` afterwards.

For songs this is the case. For albums though, it is the reverse, so the song info doesn't get updated as a result. This commit moves the updating of song info on `dataloaded` instead.

I haven't encountered issues after moving this so far, but if there are unintended side effects, please let me know.

### DIFF
--- a/src/providers/song-info-front.ts
+++ b/src/providers/song-info-front.ts
@@ -212,7 +212,6 @@ export default (api: YoutubePlayer) => {
 
     if (name === 'dataupdated' && waitingEvent.has(videoData.videoId)) {
       waitingEvent.delete(videoData.videoId);
-      sendSongInfo(videoData);
     } else if (name === 'dataloaded') {
       const video = document.querySelector<HTMLVideoElement>('video');
       video?.dispatchEvent(srcChangedEvent);
@@ -223,6 +222,7 @@ export default (api: YoutubePlayer) => {
       }
 
       waitingEvent.add(videoData.videoId);
+      sendSongInfo(videoData);
     }
   });
 


jocasrb:
I'm not sure if it is the same issue, can not understand these tecnical stuff you named, but there is obvious a problem with getting "View song credits".

I can not reproduce 100% when it happens, but also when that is present, it is not possible to click on song (to open album?) neither on the artists - on the main bar.

Sorry if it is not connected at all to the issue you mentioned.

JellyBrick:
fixed by #3659

PR #3792, opened 2025-08-29
chore(docs): update copyright footer year

RealCyGuy:


ArjixWasTaken:
might as well make it dynamic using javascript

RealCyGuy:
I can add that if you want it client-side

PR #3854, opened 2025-09-09
feat(api-server): add websocket authorization

cxntered:
adds authorization for websocket connections with bearer tokens (retrieved from the `/auth/:id` endpoint) using the `?token` query param.
if a client is unauthorized, it closes the connection with status code `1008` (policy violation) with the message "Unauthorized". we specifically close the connection with a status code (instead of preventing the upgrade to websocket) so that clients know that authorization failed.
not sure if this was the best way to implement authorization, but i am open to suggestions :>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request routing so authentication middleware no longer blocks versioned WebSocket (`/ws`) traffic.
  * WebSocket endpoints now accept an optional `token` query parameter and authenticate during connection setup when authentication is enabled.
  * Unauthorized WebSocket attempts are handled consistently: connections are closed with an authorization error when the token is missing, invalid, or not permitted.


cxntered:
if security is a concern (we *are* passing in the token directly in the url after all), passing in the token through the `Sec-WebSocket-Protocol` header could be an alternative, but i don't know if that's a major concern

JellyBrick:
@cxntered Your system appears to be affected by the `color@5.0.1` vulnerability. We recommend checking your machine. See issue #3855 for details.

JellyBrick:
Sorry for the delay in reviewing this.

How about using `Sec-WebSocket-Protocol`?
Is there a specific reason why you didn't go with this approach?

cxntered:
> How about using `Sec-WebSocket-Protocol`? Is there a specific reason why you didn't go with this approach?

there's no particular reason i went with query params. i do think using a query param is less awkward though, and it could provide a *little* more compatibility than the ws protocol header (e.g. can't set protocol due to constraints).

as i mentioned, if security is a concern with plainly passing around a token directly in the url for logs to see, i don't mind switching to using a header, but as the server isn't secure anyways (for now) i don't believe it's a huge deal.

JellyBrick (approved):

PR #3913, opened 2025-09-19
Improve video toggle plugin: replace dual buttons with single eye icon toggle

Enoughsdv:
Hello! I had an issue with the design of the toggle-video plugin, so I changed it to a simple icon, and the configuration is now applied correctly. I haven't seen the modes because I haven't focused on that, haha. The position of the thumbnail has also been improved because it used to move when hidden.

image
image
image
image


JellyBrick (changes requested):
Instead of replacing existing functionality, please add it as a new functionality.

PR #3973, opened 2025-10-04
fix(downloader): set player id to 0004e42

ArjixWasTaken:
See:
- https://github.com/LuanRT/YouTube.js/issues/1044#issuecomment-3332095783
- https://github.com/ytmd-devs/ytmd/issues/3947#issuecomment-3363738607


flavioislima:
Tested here and this fixes the issue.

PR #3974, opened 2025-10-04
misc: replace eslint + prettier with biomejs

Yumeo0:
**Why?**
Biome is an objectively faster linter + formatter as a single dependency instead of having 10 dependencies for eslint. 

**What does this PR change?**

- The styling is only minimally different from prettier so most files changed are only ordered imports.
- As far as I checked the linting also checks for the same rules as it did with eslint with the settings I put in.
- The reviewdog workflow is updated to also use Biome.
- All eslint ignores were replaced with Biome ignores.
- All files were formatted with Biome.
- I removed async from a single function. I don't think it broke anything as the visualizer still works for me. Somebody please double check though:

https://github.com/Yumeo0/youtube-music/blob/2810d2752825d5a42925ab78c318416bf9dcf793/src/plugins/visualizer/butterchurn.d.ts#L42C5-L42C71


biome.jsonc
```diff
@@ -0,0 +1,227 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.2.5/schema.json",
+  "vcs": {
+    "enabled": false,
+    "clientKind": "git",
+    "useIgnoreFile": false
```
ArjixWasTaken: I don't think it's a good idea to format files ignored by git.

Yumeo0:
> It'd make it easier to merge if there wasn't all this noise. Could you unformat all the files?

Unformatted all the files and turned on the ignore file.

(Just ignore how it took me 3 tries. I am tired rn)

ArjixWasTaken:
why delete?
I was just delaying the merge because we have other priorities rn

Yumeo0:
Because the way I rebased made it unmergeable. Here is a new one. https://github.com/pear-devs/pear-desktop/pull/4003

It's the same as this one but without merge conflicts because of the sudden rebrand and stuff.

I'm sorry. I was being a bit stupid :sob:

PR #4003, opened 2025-10-10
misc: replace eslint + prettier with biomejs

Yumeo0:
Why?
Biome is an objectively faster linter + formatter as a single dependency instead of having 10 dependencies for eslint.

What does this PR change?

- As far as I checked the linting also checks for the same rules as it did with eslint with the settings I put in.
- The reviewdog workflow is updated to also use Biome.
- All eslint ignores were replaced with Biome ignores.

(I recreated the branch/pr as it was getting messy with all the name changes. I'm still learning git please be patient with me)


JellyBrick:
Blocked by https://github.com/biomejs/biome/issues/2438

PR #4007, opened 2025-10-12
chore: update aur repository url

Syrent:
I’m not sure if the url will change again, but i think it's better to update the README for now to avoid any confusion


JellyBrick:
We can't merge this PR. See https://github.com/pear-devs/pear-desktop/discussions/3999#discussioncomment-14635021

PR #4010, opened 2025-10-14
fix: more accurate song metadata for Discord RPC

notscrappie:
makes the app display better metadata in the Discord RPC by providing the song's title instead of the alternate title and the artist's name (all the artists on the track are also now included in the RPC if they're included in the artist field) instead of selecting the first tag of the music video, the alternate title and tag are still used as fallbacks.


Yumeo0:
Ngl this should be user choice. The alternative title is the original title most of the time (youtube is inconsistent). The "title" is a translated version.

PR #4023, opened 2025-10-20
fix: Allow transparent player plugin to be used on all platforms

trnxdev:
There is literally no reason to make it Windows only.
image


dima-dencep:
I recently tested transparent windows on macOS, so it's here too!
image
(even though it's glfw)

v3nntrnx:
Closing in favor of another (probably way better) PR
#4199

PR #4057, opened 2025-11-01
feat(plugin): Taskbar Progress

### LINKED ISSUE #2416: [Feature Request]: Taskbar Progress in Windows
### Preflight Checklist

### Problem Description

Many players in Windows show their playing progress in taskbar. Here is an example with AIMP: 
image

### Proposed Solution

I want YouTube Music to also show playing progress in taskbar

### Alternatives Considered

No

### Additional Information

_No response_

ImMALWARE:
It shows track playing progress in taskbar:
image
On pause:
image
Resolves #2416

## Summary by CodeRabbit

- **New Features**
  - Added an optional Taskbar Progress plugin that displays current track playback progress in supported taskbars, including Linux Unity.
  - Progress updates during playback, pauses when playback is paused, and clears when playback stops.
  - The plugin is disabled by default and can be enabled without restarting the application.

- **Localization**
  - Added English and Russian translations for the Taskbar Progress plugin.


ImMALWARE:
@ArjixWasTaken can you review?

src/plugins/taskbar-progress/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    let progressInterval: ReturnType<typeof setInterval> | null = null;
+
+    const updateProgressBar = (songInfo: SongInfo) => {
+      if (
+        !songInfo?.title ||
+        typeof songInfo.elapsedSeconds !== 'number' ||
+        !songInfo.songDuration
+      ) {
+        return;
+      }
```
ArjixWasTaken: it'd be better to use zod here

src/plugins/taskbar-progress/index.ts
```diff
@@ -0,0 +1,79 @@
...
+            lastSongInfo &&
+            !lastSongInfo.isPaused &&
+            typeof lastSongInfo.elapsedSeconds === 'number'
+          ) {
+            updateProgressBar({
+              ...lastSongInfo,
+              elapsedSeconds: lastSongInfo.elapsedSeconds + 1,
+            });
+          }
+        }, 1000);
```
ArjixWasTaken: an interval is not guaranteed to run on a precise manner, could you maybe refactor this to not depend on the interval, and maybe calculate the elapsedSeconds using an external variable?

e.g.

```typescript
const intervalStart = performance.now();
setInterval(() => {
    const elapsedSeconds = (performance.now() - intervalStart) / 1000;
    // ...
}, 1000)

File: src/plugins/taskbar-progress/index.ts [hunk 3]
```diff
@@ -0,0 +1,79 @@
+import { createPlugin } from '@/utils';
+import { registerCallback, type SongInfo } from '@/providers/song-info';
+import { t } from '@/i18n';
+
+export default createPlugin({
+  name: () => t('plugins.taskbar-progress.name'),
+  description: () => t('plugins.taskbar-progress.description'),
+  restartNeeded: true,
```
ArjixWasTaken: please properly implement the cleanup logic required to turn this into `restartNeeded: false`, you just need to cleanup some intervals

you cannot unregister the song info callback, but you can remove all side-effects when the plugin is disabled

ArjixWasTaken (changes requested):

src/plugins/taskbar-progress/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  config: { enabled: false },
+
+  backend({ window }) {
+    let lastSongInfo: SongInfo | null = null;
+    let progressInterval: ReturnType<typeof setInterval> | null = null;
+
+    const updateProgressBar = (songInfo: SongInfo) => {
```
JellyBrick: This logic isn't needed—the callback fires every second during playback. You should handle the differentiation using the callback events.

JellyBrick (changes requested):

src/plugins/taskbar-progress/index.ts
ImMALWARE: But it doesn't... It triggers only when I manually change track progress or click play/pause.

ArjixWasTaken:
I'd like to remind you that I did the review you asked for :^)

ImMALWARE:
@ArjixWasTaken, yeah, I saw it! I'm making changes.

ImMALWARE:
Unfortunately, callback only recieves event when I manually change progressbar position or click play/pause. It does not "fire every second during playback". So I had to add method for getting song info at any time. Because plugin can be enabled while song is already playing.

ArjixWasTaken:
PS: If you think you've fulfilled the requested changes, kindly mark them as resolved by pressing "Resolve conversation".

ArjixWasTaken:
Electron promises that this would work under linux, but it doesn't.
In the meantime, I've found out how to update the progress directly via dbus.

```bash
ID="com.github.th_ch.youtube_music.desktop"
VAL=0.8

# 1. Send with application:// prefix (Unity/Older Plasma style)
dbus-send --session --type=signal /org/knotifications/UnityLauncherTest \
    com.canonical.Unity.LauncherEntry.Update \
    string:"application://$ID" \
    dict:string:variant:"progress",double:$VAL,"progress-visible",boolean:true

# 2. Send without prefix (Standard Plasma Wayland style)
dbus-send --session --type=signal /org/knotifications/UnityLauncherTest \
    com.canonical.Unity.LauncherEntry.Update \
    string:"$ID" \
    dict:string:variant:"progress",double:$VAL,"progress-visible",boolean:true
```

I'll try that way

ArjixWasTaken:
Can someone with a macos machine test if this works?

src/plugins/taskbar-progress/index.ts
```diff
@@ -0,0 +1,160 @@
...
+            startProgressInterval(songInfo, window);
+          }
+        });
+      }
+    },
+
+    stop({ window }) {
+      isEnabled = false;
+      stopProgressInterval();
+      setProgressBar(window, -1);
```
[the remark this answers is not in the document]
ArjixWasTaken: Even if invalid, it is not visible thanks to `progress > 0`.

ArjixWasTaken:
As I said earlier, I need a macOS tester for this.

ImMALWARE:
@ArjixWasTaken, I have tested this on macOS. Progressbar just went by itself and it was fully filled in ~10 seconds. I fixed it. And also there is dependencies problem, it doesn't build after this commit: https://github.com/pear-devs/pear-desktop/commit/616bb5da3f2007bd5aadfe354e7d3b8fa45e9da4
I get same error as here when building: https://github.com/pear-devs/pear-desktop/actions/runs/23715577687/job/69082005570

So, before this commit and after my fix, it works just fine

https://github.com/user-attachments/assets/dec2014f-6399-4898-a44e-cf3b9f682fef

ImMALWARE:
@ArjixWasTaken, is it okay?

JellyBrick:
@coderabbitai review

ImMALWARE:
I have separated backend logic to another file, main.ts

Previously, the @jellybrick/dbus-next import was located in index.ts. Since the renderer process imports the index.ts of all plugins to build the UI, Vite was inadvertently bundling dbus-next (and its Node.js dependencies like child_process) into the frontend client bundle. This caused the node_child_process is not defined and Uncaught TypeError: Class extends value [object Object] is not a constructor errors in dev tools on app start.

ImMALWARE:
@ArjixWasTaken @JellyBrick please review

ImMALWARE:
@coderabbitai resume

ImMALWARE:
Do I need to change absolutely everything that Rabbit wants? Because it looks like it's infinite in all PRs...

PR #4069, opened 2025-11-06
Fix downloader plugin

trevin-j:
TLDR; the only way to fix the downloader is to upgrade youtubei.js to version 16 and specify a js interpreter.

As [this comment](https://github.com/LuanRT/YouTube.js/issues/1044#issuecomment-3444828324) says, the current hardcoded player id (from #3973) doesn't work anymore. At the moment, the downloader plugin won't work.

I upgraded youtubei.js to version 16, added in the snippet shown [here](https://ytjs.dev/guide/getting-started.html#providing-a-custom-javascript-interpreter) (specifying a js interpreter), and got the downloads to work again.

Changes:

* Remove hard-coded player_id
  * The specific player_id is blocked now
* Update youtubei.js dependency to version 16 (fixes download issues)
* Specify js interpreter for sig function stuff

### DIFF
--- a/src/i18n/resources/az.json
+++ b/src/i18n/resources/az.json
@@ -123,7 +123,7 @@
           },
           "language": {
             "dialog": {
-              "message": "Dil yenidən başladılandan sonra dəyişəcək",
+              "message": "Dil yenidən başlatmadan sonra dəyişiləcək",
               "title": "Dil dəyişdirildi"
             },
             "label": "Dil",
@@ -131,9 +131,11 @@
               "to-help-translate": "Tərcüməyə kömək etmək istəyirsiniz? Buraya basın"
             }
           },
-          "resume-on-start": "Tətbiq başladıqda son mahnını davam etdir",
+          "resume-on-start": "Tətbiq başladıqda son mahnıdan davam et",
+          "single-instance-lock": "Tək proqram kilidi",
+          "start-at-login": "Giriş səhifəsində başlat",
           "starting-page": {
-            "label": "Başlanğıc səhifəsi",
+            "label": "Giriş səhifəsi",
             "unset": "Təyin edilməyib"
           },
           "tray": {
--- a/src/i18n/resources/cs.json
+++ b/src/i18n/resources/cs.json
@@ -337,7 +337,7 @@
       }
     },
     "audio-compressor": {
-      "description": "Apply compression k audiu (snižuje hlasitost nejhlasitěších částí signálu and zvyšuje hlasitost nejjemnějších částí)",
+      "description": "Aplikuje kompresi k audiu (snižuje hlasitost nejhlasitěších částí signálu a zvyšuje hlasitost nejjemnějších částí)",
       "name": "Audio kompresor"
     },
     "auth-proxy-adapter": {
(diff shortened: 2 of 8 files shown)


rabid-sausage:
I can confirm this works, manually applied the changes and built a working version and can now download without issue.
Myself prior to reading this, I  got as far as upgrading youtubei.js to 16 and including meriyah, but could not wok out what to do with the required script, where to insert it, but can see from what you have done it was actually rather simple (still do not know what the script does). The Player ID fix was a quick get-around a bigger problem, found that out on other groups who first offered the Player ID solution.

Thanks for you effort and showing what you have done.

It will be interesting to see if this get pushed in to the next build, as it looks like some parties like youtubei.js, projects that use their code and others which operate similar are now not including this type of function in pre-compiled downloads, requiring individual users to manually insert/adjust the code and build a working copy themselves to gain this and other functions. Maybe a method avoiding getting in to trouble directly.

LyonDR:
> I can confirm this works, manually applied the changes and built a working version and can now download without issue. Myself prior to reading this, I got as far as upgrading youtubei.js to 16 and including meriyah, but could not wok out what to do with the required script, where to insert it, but can see from what you have done it was actually rather simple (still do not know what the script does). The Player ID fix was a quick get-around a bigger problem, found that out on other groups who first offered the Player ID solution.
> 
> Thanks for you effort and showing what you have done.
> 
> It will be interesting to see if this get pushed in to the next build, as it looks like some parties like youtubei.js, projects that use their code and others which operate similar are now not including this type of function in pre-compiled downloads, requiring individual users to manually insert/adjust the code and build a working copy themselves to gain this and other functions. Maybe a method avoiding getting in to trouble directly.

Can you pass me your app, please?

ArjixWasTaken:
LGTM, thanks for your contribution!

ghost:
I know the eval example is from the youtubei.js repo but wouldn't it be better to run the code using `node:vm` with an empty context like [here](https://github.com/its-iris/pear-desktop/blob/366b309484dd4efa3935abd42fe9b0edd86f3ced/src/plugins/downloader/main/index.ts#L76) ? It's still not good enough to run untrusted code but it should be the best option without an extra dependency.

ArjixWasTaken:
@its-iris would you like to open a PR?

LyonDR:
> Can someone explained how to do that like step by step for noobs?

I recommend asking chatgpt 😅 That's what I did.

You send him the link to this PR and ask him to guide you on how to apply the fix.

Neosettler:
I've been caught in a PR loop on this issue. Running current master, still unable to download. Other efforts are being made to solve this issue?

JellyBrick:
> I know the eval example is from the youtubei.js repo but wouldn't it be better to run the code using `node:vm` with an empty context like [here](https://github.com/its-iris/pear-desktop/blob/366b309484dd4efa3935abd42fe9b0edd86f3ced/src/plugins/downloader/main/index.ts#L76) ? It's still not good enough to run untrusted code but it should be the best option without an extra dependency.

I've tried several approaches, but `BGUtils` doesn't work properly in a `node:vm` environment. (It might look like it's working, but the values aren't valid)

PR #4096, opened 2025-11-19
Fix: MacOS install commands in ReadMe

haythemgataa:
- Removed the Homebrew command as it's currently not working.
- Fixed the app name in the command that fixes the MacOS error when opening the app.


JellyBrick:
Due to legal issues, we cannot merge this PR.

PR #4103, opened 2025-11-22
add: show YTM user pfp / name in Discord RPC

noxygalaxy:
<3

## Summary by CodeRabbit

* **New Features**
  * Discord Rich Presence now displays the connected YouTube account’s avatar and name by default.
  * Added a Discord menu toggle to show or hide account information.
  * Account details update automatically when available, and the preference is saved.
* **Localization**
  * Added English, Russian, and Ukrainian translations for the new menu option.
* **Reliability**
  * Improved account information detection and retry handling when details are temporarily unavailable.


JellyBrick:
Please fix these issues.

noxygalaxy:
> Please fix these issues.

done :)

src/plugins/discord/discord-service.ts
```diff
@@ -400,6 +421,109 @@ export class DiscordService {
...
+              this.fetchYouTubeUserInfo();
+            },
+            5000,
+          );
+        }
+      }
+    } catch (err) {
+      console.error(LoggerPrefix, 'Failed to fetch YouTube user info:', err);
+    }
+  }
```
ArjixWasTaken, line 464: This should probably be using electron's IPC.
The main benefit is that the code would be checked by typescript, and it'd also make editing it easier since we'd have autocomplete, etc.

noxygalaxy: I guess i made what you wanted

noxygalaxy:
so?

src/plugins/discord/preload.ts
```diff
@@ -0,0 +1,133 @@
...
+        return false;
+      }
+
+      lookupInFlight = true;
+      try {
+        let avatar: string | null = null;
+
+        // Find avatar first - this is always visible
+        const accountButton =
+          document.querySelector<HTMLImageElement>(
```
JellyBrick: Since the preload context doesn't have access to the browser's DOM, this logic should be handled in the renderer context instead.

JellyBrick (changes requested):

PR #4113, opened 2025-11-26
feat(synced-lyrics): NetEase provider

### LINKED ISSUE #4059: [Feature Request]: Can lyrics sources from NetEase Cloud Music be added?
### Preflight Checklist

- [x] I use the latest version of Pear Desktop (Application).
- [x] I have searched the [issue tracker](https://github.com/pear-devs/pear-desktop/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

In NetEase Cloud Music, there are a large number of Chinese translated lyrics, which are very useful for Chinese users.
I found a Tampermonkey script that can enable this feature for the web version of YouTube Music:
👉 https://greasyfork.org/zh-CN/scripts/548724-youtube-music-spotify-%E7%BD%91%E6%98%93%E4%BA%91%E6%AD%8C%E8%AF%8D%E6%98%BE%E7%A4%BA

So it should definitely be possible — but I don’t know how to develop plugins myself.

### Proposed Solution

Currently, Pear Music (or Pear Desktop) does not display translated lyrics for many international songs, especially in Chinese.
For Chinese-speaking users, translated lyrics are very important for understanding and enjoying foreign songs.

NetEase Cloud Music has a huge library of high-quality, time-synced Chinese translations, and there’s already a Tampermonkey script that integrates them into YouTube Music and Spotify.
This shows that technically, it’s possible to fetch and display NetEase lyric data.

Adding this feature would greatly improve the listening experience for Chinese users.

### Alternatives Considered

Add an optional lyrics source integration for NetEase Cloud Music.
The app could:

Automatically detect songs that exist on NetEase and fetch their lyrics and translations.

Allow users to switch between original lyrics and translated versions.

Cache the lyrics locally for better performance.

If official API access is difficult, a community API (like used in the Tampermonkey script) could be used as an alternative.

### Additional Information

Using external browser extensions like Tampermonkey (works but only for web players, not native apps).

Manually searching for lyrics and importing them (inconvenient and time-consuming).

ArjixWasTaken:
Resolves #4059

Notes: NetEase's search algorithm is a bit rough


ArjixWasTaken:
I rushed the code a bit using gemini, but I did rewrite the search result ranking part to mimic what we did on lrclib.
For that reason, I won't be the one to merge this PR, and will accept your reviews.

src/plugins/synced-lyrics/providers/NetEase.ts
```diff
@@ -0,0 +1,340 @@
+// Code adapted from https://greasyfork.org/en/scripts/548724-youtube-music-spotify-%E7%BD%91%E6%98%93%E4%BA%91%E6%AD%8C%E8%AF%8D%E6%98%BE%E7%A4%BA
+// which is licenced under the MIT licence
+
+import CryptoJS from 'crypto-js';
```
JellyBrick, line 4: I think we can avoid using CryptoJS.

We might be able to achieve this using native APIs (like Web Crypto API or Node's crypto module).

JellyBrick (changes requested):

src/plugins/synced-lyrics/providers/NetEase.ts
ArjixWasTaken: web crypto would be a pain to use, and node's crypto would require context switching via IPC
and I was too lazy to do that

but if you insist, then I am all for it

PR #4119, opened 2025-11-28
Fix: discord activity sanitizer

Tx-ID:
Discord doesn't like it if activity text are less than 2 characters.
So I added extra function just for this and used an invisible non whitespace character to make it 2 characters or just simply return `"undefined"`

### DIFF
--- a/src/plugins/discord/utils.ts
+++ b/src/plugins/discord/utils.ts
@@ -28,7 +28,7 @@ export function sanitizeActivityText(input: string | undefined, fallback: string
   let safeString = truncateString(text, 128);
 
   if (safeString.length === 0) {
-    return 'undefined'; // change if necessary
+    return fallback;
   }
 
   if (safeString.length < 2) {


src/plugins/discord/utils.ts
```diff
@@ -17,6 +17,30 @@ export const truncateString = (str: string, length: number): string => {
...
 
+/**
+ * Sanitizes a string for Discord Activity.
+ * @param input - The string to sanitize.
+ * @returns The sanitized string or undefined.
+ */
+export function sanitizeActivityText(input?: string): string {
+  if (!input) {
+    return 'undefined';
+  }
```
Su-Yong: how about change the signiture of `sanitizeActivityText`?
`sanitizeActivityText: (input: string | undefined, fallback = 'undefined') => string` format seems to be better

Su-Yong (changes requested):

src/plugins/discord/utils.ts
Tx-ID: sure I can work with that

src/plugins/discord/utils.ts
```diff
@@ -17,6 +17,27 @@ export const truncateString = (str: string, length: number): string => {
...
+ * @param input - The string to sanitize.
+ * @param fallback - A fallback string to use if the input is empty or whitespace. Defaults to 'undefined'.
+ * @returns The sanitized string, compliant with Discord's requirements.
+ */
+export function sanitizeActivityText(input: string | undefined, fallback: string = 'undefined'): string {
+  const text = (input && input.trim()) ? input.trim() : fallback.trim();
+  let safeString = truncateString(text, 128);
+
+  if (safeString.length === 0) {
+    return 'undefined'; // change if necessary
```
ArjixWasTaken: Any reason to not use the existing fallback?

```suggestion
    return fallback;
```

src/plugins/discord/utils.ts
```diff
@@ -17,6 +17,27 @@ export const truncateString = (str: string, length: number): string => {
...
+  const text = (input && input.trim()) ? input.trim() : fallback.trim();
+  let safeString = truncateString(text, 128);
+
+  if (safeString.length === 0) {
+    return fallback;
+  }
+
+  if (safeString.length < 2) {
+    safeString = safeString.padEnd(2, '⠀'); // change if necessary
+  }
```
ArjixWasTaken, line 36: Hmm, `.padEnd(2, ' ')` implies that a string of `'  '` is allowed, but that's far from the case.
If the string is empty, it will return the fallback. It's improbable for `.padEnd(2, ' ')` to add 2 characters, it will always add 1.

Maybe it makes sense for this to be?
```suggestion
  if (safeString.length < 2) {
    safeString += '⠀';
  }
```

I am not making a demand, just talking

Tx-ID: When writing, I knew about the redundancy but decided to leave it there since the premise is just to fix the input string

src/plugins/discord/discord-service.ts
```diff
@@ -99,13 +99,17 @@ export class DiscordService {
     const activityInfo: SetActivity = {
       type: ActivityType.Listening,
       statusDisplayType: config.statusDisplayType,
-      details: truncateString(songInfo.alternativeTitle ?? songInfo.title, 128), // Song title
```
JellyBrick, line 102: Please truncate the string.

JellyBrick (changes requested):

src/plugins/discord/discord-service.ts
Tx-ID: Please elaborate, the newly added function already does truncation

PR #4122, opened 2025-11-29
Remove Compact Sidebar plugin

SomeAspy:
This appears to be a native feature now, and is saved between restarts and reloads regardless of whether the plugin is enabled or not


SomeAspy:
didn't mean to include all changes

PR #4123, opened 2025-11-29
Remove compact sidebar plugin

SomeAspy:
This seems to be native functionality now, and is no longer needed.

## Summary by CodeRabbit

* **Revert**
  * Removed the compact sidebar feature from the app, so sidebar behavior will no longer be automatically adjusted.


JellyBrick:
Thanks for the contribution!

PR #4143, opened 2025-12-11
Add devcontainer setup

makisukurisu:
Hi! I've noticed that this project doesn't have devcontainer settings for neither developing the app, or simply building it without hustle.

I've decided to try to add the most basic setup for being able to build pear-desktop without installing pnpm and other binaries.

And while building works just fine, debugging (at least on Fedora Linux with Wayland) doesn't seem to work. I suppose it might work better on other systems (for example with WSL, or directly on Windows), but I can't test this.

If you find this change more harmful then helpful, feel free to ignore it and close the PR.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,7 @@
 {
 	"name": "Pear Desktop - Dev Container",
 	// Keep in sync with `.github/workflows/build.yml`
-	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+	"image": "mcr.microsoft.com/devcontainers/typescript-node:24",
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	"features": {},
 	"postCreateCommand": "pnpm install --frozen-lockfile"


.github/dependabot.yml
JellyBrick, line 1: Please remove this file.
We already use renovate for dependency version management.

makisukurisu: Done :heavy_check_mark: 

As I understand from [this page](https://docs.renovatebot.com/modules/manager/devcontainer/), renovate should automatically check devcontainer files with no extra changes required.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,16 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+	"name": "Pear Desktop - Dev Container",
+	// Keep in sync with `.github/workflows/build.yml`
+	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
```
ArjixWasTaken: I'd rather you don't pin to a debian release, it would also be nice if you used the latest LTS release, in this case `24`.

```suggestion
	"image": "mcr.microsoft.com/devcontainers/typescript-node:24",
```

README.md
```diff
@@ -130,6 +130,10 @@ pnpm install --frozen-lockfile
 pnpm dev
 ```
 
+Instead of installing pnpm on your system, you can also use [devcontainers](https://containers.dev/). You can use devcontainers either as a development environment in VS Code, or as a way to easily build the project without installing dependencies on your host system.
```
ArjixWasTaken, line 133: This is shown as a benefit, but it kinda defeats the purpose of pnpm.
Each project will have its own pnpm store (in a devcontainer), taking more disk space.

Not sure how I'd rephrase it though, I am bad with words.

ArjixWasTaken: Not like I'd use anything other than pnpm

PR #4161, opened 2025-12-19
plugin: clock widget

WyattBreitner:
YTM_Clock

### Description:
Adds a simple clock to the navigation bar. 

### Reason:
Useful when full screen.

### Notes:
Development was done on Linux, windows is untested.

Time-string's locale is currently hardcoded to `en`

### DIFF
--- a/src/plugins/clock/index.tsx
+++ b/src/plugins/clock/index.tsx
@@ -17,7 +17,7 @@ const defaultConfig: ClockPluginConfig = {
 export default createPlugin({
   name: () => t('plugins.clock.name'),
   description: () => t('plugins.clock.description'),
-  restartNeeded: true,
+  restartNeeded: false,
   config: defaultConfig,
   stylesheets: [style],
   menu: async ({ getConfig, setConfig }): Promise<MenuTemplate> => {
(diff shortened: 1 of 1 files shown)


src/plugins/clock/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+const defaultConfig: ClockPluginConfig = {
+  enabled: false,
+  displaySeconds: false,
+  hour12: false,
+};
+
+export default createPlugin({
+  name: () => t('plugins.clock.name'),
+  description: () => t('plugins.clock.description'),
+  restartNeeded: true,
```
ArjixWasTaken: Since you properly dispose of any side-effects in the stop() method, why is this marked as `restartedNeeded; true`?
Is it not safe to enable/disable w/o restarting?

WyattBreitner: I don't fully understand when a restart is needed, so left it for someone who does know, to look at.
This is my first time working with this project, so there may be more I missed as well.

ArjixWasTaken: If the plugin can be started/stopped at arbitrary times, as many times as the user wants, then it doesn't need the entire app to restart.

TL;DR if you cannot cleanup after yourself, you need a restart

PR #4198, opened 2025-12-24
refactor(downloader): Updated to use node:vm for unsafe scripts

ghost:


JellyBrick:
BG does not run on node:vm.

PR #4201, opened 2025-12-24
refactor(music-together): Removed delay dependency

ghost:


src/plugins/music-together/connection.ts
```diff
@@ -104,16 +98,14 @@ export class Connection {
...
-        await delay(10000);
-        try {
-          this.peer.reconnect();
-          return;
-        } catch {
-          //ignored
-        }
+    this.peer.on('error', (err) => {
+      if (err.type === 'network') {
+        setTimeout(() => {
```
JellyBrick, line 103: Have you verified that there are no side effects caused by these changes?

ghost: Yeah. Seems to be fine on my end. It keeps trying to reconnect every 10 seconds until a connection can be made. It can be tested using the throttling setting in dev tools. I don't get why an async delay was used here?

PR #4202, opened 2025-12-24
refactor(taskbar-mediacontrol): Removed jimp dependency

ghost:


JellyBrick:
Is there a specific reason for making these changes?

ghost:
Jimp has a lot of dependencies (for other image formats) and forces most dependencies to older versions. I suppose it would be cleaner to just provide inverted color media icons instead. Besides Jimp is not used elsewhere.

JellyBrick:
> Jimp has a lot of dependencies (for other image formats) and forces most dependencies to older versions. I suppose it would be cleaner to just provide inverted color media icons instead. Besides Jimp is not used elsewhere.

👍 
Could you add the inverted color media icons?

Su-Yong (approved):

PR #4225, opened 2026-01-05
youtube.js yt-dlp switch

SkNightmare:
this is a purpose for an switch button for switch into download method with yt-dlp et youtube.js

i think my code isn't the best code but with my test all working i have tested on linux i can't test it on windows rip but all working fine with me


SkNightmare:
i dont have say'd why i have code that its because for some reason i dont know why but youtube.js method won't work somtime with me

src/plugins/downloader/main/index.ts
```diff
@@ -441,6 +442,135 @@ async function downloadSongUnsafe(
     return;
   }
 
+  // If configured to use yt-dlp, delegate downloading to the external binary
+  if ((config.engine ?? 'youtube.js') === 'yt-dlp') {
+    const findYtdlpExecutable = async (preferred?: string) => {
+      const candidates: (string | undefined)[] = [];
+      if (preferred) candidates.push(preferred);
+      candidates.push('/usr/bin/yt-dlp', '/usr/local/bin/yt-dlp', 'yt-dlp');
```
ArjixWasTaken: Instead of having a hardcoded path of candidates, we should instead be searching based on PATH.
Something like [`which`](https://www.npmjs.com/package/which) could be used.

Also, yt-dlp is a fork, it'd probably be a good idea to also support the original youtube-dl

SkNightmare: i can try but normaly its over my comptence

SkNightmare: but honesly i have place an config on the extention configuration for choose the path with the path finder of the system file explorer avalible on the system personaly i think that probably good

folliehiyuki: > Also, yt-dlp is a fork, it'd probably be a good idea to also support the original youtube-dl

Supporting both is welcomed. But, if only 1 can be supported in this PR, I'll prefer `yt-dlp`. The original `youtube-dl` doesn't even get packaged by any Linux distributions anymore.

ArjixWasTaken: > i can try but normaly its over my comptence

```bash
pnpm add which
pnpm add -D @types/which
```

```typescript
import { which } from 'which';

const ytdlpExecutable = await which('yt-dlp', { nothrow: true })
```

this is over your competence? I beg to differ

SkNightmare: do

PR #4278, opened 2026-01-29
Fix-improve-timestamp-calculation-in-addScrobble-method

Devajuice:
chore: update .gitignore to include Playwright test results chore: update package.json and pnpm-lock.yaml for @types/node and @playwright/test versioning


JellyBrick:
Each PR should focus on one change only. Additionally, did you use an AI tool to create this PR?

PR #4307, opened 2026-02-06
Fix Crossfade randomly muting music until manual unmute

Orionicelion:
Fixes multiple open issues on the crossfade plugin muting after a transition


JellyBrick (changes requested):
Please fix these issues.
(See review comments)

PR #4374, opened 2026-03-16
fix(synced-lyrics) correct milliseconds parse

AARP41298:
Hi
I found this error.
Let me know if its a proper way to fix it.
I was editing some of my karaoke videos when i realize somethings was not perfectly aligned.

Also, i think the `offset` is broken in lines where there's no text after the time stamp, but i will check it later.


ArjixWasTaken:
Good catch!

PR #4381, opened 2026-03-23
fix: ensure WebSocket notifications are sent before application quits

cjuriartec:
## Problem
When closing the application (e.g. via Cmd+Q on macOS), the WebSocket clients wouldn't consistently receive a final "isPlaying: false" state because the application was terminating before the network buffers could be flushed.

## Solution
- Expose a [wsNotifyClose](cci:1://file:///Users/cesarujos/Software/pear-desktop/src/plugins/api-server/backend/routes/websocket.ts:57:2-67:4) function in the api-server plugin.
- Added a `before-quit` event listener in the main backend process that triggers this notification.
- Ensured structural compatibility for WebSocket context types with `as unknown as WSContext<WebSocket>` to prevent TypeScript errors.

## Testing
Verified with Boring Notch connection. The notification is now consistently received before Pear Desktop fully exits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful shutdown of WebSocket connections to properly notify connected clients when the application closes or the backend stops, preventing abrupt disconnections.


src/plugins/api-server/backend/main.ts
```diff
@@ -28,8 +28,14 @@ import type {
 
 export const backend = createBackend<BackendType, APIServerConfig>({
   async start(ctx) {
+    const { app } = await import('electron');
```
JellyBrick, line 34: Is there a reason you used `await import`?

JellyBrick (changes requested):

src/plugins/api-server/backend/main.ts
ArjixWasTaken: Maybe they were not aware that vite does tree-shaking

PR #4481, opened 2026-06-09
feat(chromecast): add Google Cast plugin

Adamkadaban:
## Summary

Adds a **Google Cast** plugin: cast high-quality audio to Cast devices (Chromecast, Google Home, Nest) straight from the player bar. While casting, the local player is muted and acts as a silent "conductor" that drives the queue, so only the speaker is audible.

## Features

- **Device discovery + picker** — mDNS discovery with an in-player cast button and dropdown.
- **High-quality audio** — a local HTTP proxy downloads the best audio via `youtubei.js` and **remuxes it to progressive ADTS/AAC with the bundled `ffmpeg.wasm`**. (Rationale below.) Served with HTTP Range support and gated behind an unguessable per-session token.
- **Unified volume** — the YTM volume slider drives the speaker; the local element stays muted without persisting to YTM's stored volume, and the user's pre-cast mute state is restored on disconnect.
- **Two-way control** — mirrors play/pause and scrubbing to the speaker, and reflects *external* control (e.g. the Google Home phone app) back onto the local player.
- **Tight sync** — a one-shot resync of the muted local player to the speaker after load keeps the conductor from running ahead and clipping the speaker at track boundaries.
- **Ad-safe** — suppresses mirroring while an ad shows locally.

## Why remux?

YouTube's high-quality audio-only formats (itag 140 AAC, 251 Opus, …) are delivered as **fragmented (DASH)** MP4/WebM, which the Cast Default Media Receiver's progressive player rejects (`idleReason=ERROR`). Muxed formats are progressive but low quality and waste video bandwidth. So the proxy downloads the adaptive audio locally and stream-copies/remuxes it into a progressive ADTS/AAC stream the receiver can play.

## Quality

- ESLint: **0 errors**; `tsc` typecheck clean; `pnpm build` clean on Linux/macOS/Windows.
- The backend chain is **side-effect-free**, so the renderer bundle never pulls `electron`/node deps (verified: no `castv2-client`/`bonjour-service`/`@hono/node-server`/`ffmpeg.wasm` in the renderer build).
- **Unit tests** for the pure helpers (LAN-IP detection, HTTP Range parsing, seek/drift math) via the existing Playwright runner.

## Notes

- New dependencies: `castv2-client`, `bonjour-service` (both MIT).
- Plugin is disabled by default and self-registers via the standard `src/plugins/<name>` convention; English strings added to `en.json`.
- The plugin was developed and reviewed with AI assistance, with each change verified against lint/typecheck/build/tests and manually tested against real devices (Nest Audio, Home Mini).

## AI Disclosure

I didn't see an AI policy on this repo, but I thought I'd let ya'll know anyway this code was all LLM-generated. 
But code was all human reviewed and tested.

## Summary by CodeRabbit

* **New Features**
  * Added Chromecast support to cast YouTube Music to Google Cast devices
  * Integrated device discovery to automatically find nearby Chromecast devices
  * Added cast controls in the player interface with device selection menu
  * Included playback synchronization between local player and Chromecast speaker
  * Added volume control and mute options for casting sessions
  * Implemented auto-reconnect functionality for seamless playback continuity


Adamkadaban:
Looking into coderabbit comments

JellyBrick:
YTM supports Cast by default, so implementing the browser Cast API should be enough to support Google Cast without having to build a dedicated stream-downloading feature through youtubei.js

PR #4510, opened 2026-06-28
remove performance-improvement plugin

SomeAspy:
This is a pretty opinionated PR but here goes:
The performance improvement plugin is poorly documented, and the code looks hacky.
If we really want to nitpick performance, I would probably suggest experimenting with flags on electron.

## Summary by CodeRabbit

* **Chores**
  * Removed a performance-related plugin and its associated script assets.
  * Cleaned up locale files so the plugin no longer appears in supported translations.
  * Updated ignore rules to reflect the removed files.


SomeAspy:
oops

SomeAspy:
reopen this i guess

PR #4616, opened 2026-07-29
feat(resume-shuffle): Add plugin to resume the last queue shuffled

### LINKED ISSUE #994: Resume last song in shuffle when app starts
The option to `Resume last song when app starts` is great, but if the playlist was shuffled before the app was closed, it is resumed unshuffled after. Not sure if it's easy to implement, but having it remember the shuffle state as well (or just the existing shuffled "Up Next") would be incredible.

danielchalmers:
Closes #994

"Resume last song when app starts" restores the last played song, but the queue always comes back in its original order. This plugin saves the player bar's shuffle state and re-applies it once the resumed queue has loaded:

- The player bar's `shuffle-on` attribute is mirrored into the plugin's own config, so nothing is added to the core config schema.
- On startup, if resume is enabled and the saved state was shuffled, the queue is shuffled once it has loaded. The web player's own shuffle keeps the playing song at the top, so the resumed song keeps playing.
- Runs on watch-page loads while resume is enabled: the resumed page at startup, and after an in-page reload (such as Memory Saver #4617, where the player would otherwise drop the shuffle). Gives up after 10s so a queue the user starts later can't be shuffled by surprise.

## Summary by CodeRabbit

- **New Features**
  - Added localized i18n text for **Resume Shuffle** (display name and description).
- **Bug Fixes**
  - Improved reliability when enabling/disabling **Resume Shuffle**: shuffle restoration stops if the plugin is turned off mid-resume, and re-enabling resumes cleanly without leftover observers/subscriptions.

### DIFF
--- a/src/i18n/resources/en.json
+++ b/src/i18n/resources/en.json
@@ -769,6 +769,10 @@
         }
       }
     },
+    "resume-shuffle": {
+      "description": "Bring the queue back shuffled when \"Resume last song when app starts\" is enabled",
+      "name": "Resume Shuffle"
+    },
     "scrobbler": {
       "description": "Add scrobbling support (etc. last.fm, Listenbrainz)",
       "dialog": {
--- a/src/plugins/resume-shuffle/index.ts
+++ b/src/plugins/resume-shuffle/index.ts
@@ -29,7 +29,8 @@ const isShuffled = () =>
     .querySelector<HTMLElement>('ytmusic-player-bar')
     ?.attributes.getNamedItem('shuffle-on') ?? null) !== null;
 
-// The queue is filled asynchronously, so wait for it before shuffling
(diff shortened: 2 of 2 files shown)


danielchalmers:
Both CodeRabbit findings were valid and are fixed in 70da26d:

- `stop()` now cancels the queue subscription and the give-up timer, so a pending shuffle cannot fire after the plugin is disabled.
- The fallback `saveShuffleState()` now only runs when no restore was wanted. Previously a missing queue landed in the same branch and overwrote a saved `shuffled: true` with `false`, which would have silently disabled the feature from then on.

Re-verified on a real 50-track playlist afterwards: shuffle on → quit → relaunch still comes back shuffled with the resumed song playing. `pnpm lint`, `pnpm format:check` and `pnpm typecheck` are clean.

PR #4640, opened 2026-08-16
feat(plugins): add mini player

### LINKED ISSUE #214: [Feature Request] Add Floating Mini Player Plugin

dcgancan:
Closes #214.

Adds a `mini-player` plugin: a small, frameless, always-on-top window with the album art, the track title/artist and playback controls, so the player stays reachable while the main window is hidden.

Mini player

On hover it reveals the elapsed/total time and a close button:

Mini player on hover

### What it does

- Album art, title and artist (the title scrolls when it does not fit)
- Like, volume, shuffle, previous, play-pause, next and repeat, all reflecting the current player state (repeat cycles none → all → one)
- Click-to-seek progress line along the bottom edge
- Volume: scroll anywhere on the bar, or hover the speaker icon and the bottom line turns into a volume slider
- Drag the bar to move it; hover the album art for ⤢ to bring the main window back (double-click does the same)
- Window position and size are remembered

### How it works

The mini player is a separate `BrowserWindow` that loads a self-contained asset from `assets/mini-player/`, with a small preload exposing only three channels over `contextBridge`. It does not touch the YouTube Music page: state comes from `registerCallback` in `providers/song-info` plus the existing `peard:volume-changed` / `peard:like-changed` / `peard:shuffle-changed` / `peard:repeat-changed` events, and every action goes back out through `providers/song-controls`. The main window keeps playing while hidden, with background throttling disabled for the duration.

Everything is configurable from the plugin menu: opacity, "fully opaque on hover", always-on-top, whether the main window hides, open-on-startup, the toggle hotkey (`CmdOrCtrl+Shift+M` by default) and a position/size reset. The plugin is disabled by default and does not need a restart to toggle.

### Platform notes

- `BrowserWindow.setOpacity` is a no-op on Linux, so the opacity setting is applied as CSS opacity on that platform instead
- `transparent: true` on Linux needs a running compositor, same caveat as `transparent-player`
- Developed and manually verified on macOS (arm64); Windows and Linux are covered by the build only

### Testing

`tests/mini-player.test.js` launches the app with the plugin enabled and asserts that the window opens with the right flags, that a pushed state renders (title, artist, like state, progress ratio) and that clicking a control reaches the main process through the preload bridge. `pnpm check` and `pnpm build` are clean.

### Notes

- Only `en.json` is touched; the other locales are left to Weblate
- This is separate from the Windows-taskbar widget in #4349 and the tray hover player in #4428 — this one is a standalone floating window, so they should not conflict, but happy to adjust if you would rather see these consolidated
- Happy to change the layout, defaults or the control set if you have a different shape in mind

## Summary by CodeRabbit

- **New Features**
  - Added a floating mini-player with album artwork, track details, playback controls, progress and volume sliders, and draggable positioning.
  - Added synchronized playback, volume, mute, like, shuffle, repeat, and track information.
  - Added quick access from compatible player menus, plus options for opacity, startup behavior, hotkeys, window placement, sizing, and always-on-top display.
  - Added menu actions to open, close, configure, and reset the mini-player.

- **Tests**
  - Added coverage for mini-player launch, display, playback state, and control interactions.

### DIFF
--- a/assets/mini-player/index.html
+++ b/assets/mini-player/index.html
@@ -194,12 +194,20 @@
         height: 26px;
       }
 
+      #shuffle,
+      #repeat {
+        width: 24px !important;
+        height: 24px !important;
+      }
+
       #play {
         width: 30px !important;
         height: 30px !important;
       }
 
-      #like.liked {
+      #like.liked,
+      #shuffle.active,
+      #repeat.active {
         color: var(--accent);
         opacity: 1;
       }
@@ -289,6 +297,29 @@
       </div>
 
       <div id="controls">
+        <button id="like" data-action="like" title="Like">
+          <svg width="17" height="17" viewBox="0 0 24 24">
+            <path
+              id="like-path"
+              d="M12 20.4 3.9 12.6a5 5 0 0 1 7.1-7l1 1 1-1a5 5 0 0 1 7.1 7L12 20.4Zm0-2.8 6.7-6.4a3 3 0 1 0-4.3-4.2L12 9.4 9.6 7A3 3 0 1 0 5.3 11.2L12 17.6Z"
+            />
+          </svg>
+        </button>
+        <button id="volume" data-action="mute" title="Mute">
+          <svg width="17" height="17" viewBox="0 0 24 24">
+            <path
+              id="volume-path"
+              d="M4 9h3l4-4v14l-4-4H4V9Zm11.5 3a3.5 3.5 0 0 0-2-3.2v6.4a3.5 3.5 0 0 0 2-3.2Zm0-7.3v2.1a5.5 5.5 0 0 1 0 10.4v2.1a7.5 7.5 0 0 0 0-14.6Z"
+            />
+          </svg>
+        </button>
+        <button id="shuffle" data-action="shuffle" title="Shuffle">
+          <svg width="16" height="16" viewBox="0 0 24 24">
+            <path
+              d="M10.59 9.17 5.41 4 4 5.41l5.17 5.17 1.42-1.41zM14.5 4l2.04 2.04L4 18.59 5.41 20 17.96 7.46 20 9.5V4h-5.5zm.33 9.41-1.41 1.41 3.13 3.13L14.5 20H20v-5.5l-2.04 2.04-3.13-3.13z"
+            />
+          </svg>
+        </button>
         <button data-action="previous" title="Previous">
           <svg width="18" height="18" viewBox="0 0 24 24">
             <path d="M6 6h2v12H6V6Zm3.5 6L18 6v12L9.5 12Z" />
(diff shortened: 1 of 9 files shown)


dcgancan:
Pushed a follow-up: added shuffle and repeat, so the control row is now like / volume / shuffle / previous / play-pause / next / repeat, matching the layout from the screenshot in #214. Repeat cycles none → all → one and shows the repeat-one glyph; both light up when active. Default width went 400 → 480 so the extra buttons do not crowd the title. Screenshots in the description are updated.

dcgancan:
Thanks for the review — all seven are addressed in 2b5276e9, including the three nitpicks.

Two notes where the fix differs slightly from the suggestion:

**Main window restore.** I split it into `unhideMainWindow()` (undoes only a hide this plugin performed, guarded by a `didHideMainWindow` flag) and `showMainWindow()` (the explicit ⤢ / double-click action). The shutdown path is covered by an `isQuitting` flag set from the main window's `close` handler. Worth flagging though: on macOS this is not fully observable, because `app.on('activate')` in `src/index.ts` re-shows the main window whenever the app is activated, and closing the mini player activates the app. I wrote a regression test for it and dropped it again once I traced that behaviour — the plugin no longer touches visibility it does not own, but on macOS the app-level handler still brings the window back.

**Setup messages.** Rather than reordering the `openOnStart` path, the four setup channels plus `peard:get-shuffle` moved into `requestPlayerState()`, called both when the window opens and on `peard:player-api-loaded`. The listeners on the renderer side are `singleton()`-wrapped, so re-sending is a no-op when they are already registered.

The shortcut one was the most valuable catch — `unregisterAll` was silently dropping the local shortcuts that `in-app-menu` and `shortcuts` register on the same window whenever the hotkey changed. It now unregisters only its own accelerator, using the previous value from `onConfigChange`.

`pnpm check`, `pnpm build` and the Playwright test are green.

dcgancan:
Teardown now claims the temp profile before `electron.launch`, so a failed startup cannot leak it, and a failed `app.close` is reported rather than swallowed (fefbf8e7). I stopped short of gating the directory removal on confirmed Electron shutdown — for a throwaway profile under `os.tmpdir()` that seemed like more machinery than the failure mode warrants.

dcgancan:
Added a song menu entry so the mini player is reachable without knowing the hotkey, following the same pattern picture-in-picture uses (`getSongMenu` + a `MutationObserver` on `ytmusic-popup-container`, with an `isPlayerMenu` guard).

This is the only part of the plugin that touches the YouTube Music DOM, and it fails quietly: if those nodes are renamed the entry stops appearing while the hotkey and the plugin menu keep working. The renderer half also cleans up in `stop`, so the plugin still toggles without a restart.

Happy to drop it if you would rather keep the plugin DOM-free, or to move it to the player bar instead — I avoided `.player-minimize-button` deliberately since picture-in-picture already hooks that one.

dcgancan:
Good catch, and it is worse than it looks — I checked what `document.querySelector('#icon')` actually resolves to on the live page: the first match is a `tp-yt-iron-icon` inside a `tp-yt-paper-icon-button` under `ytmusic-app-layout`, nowhere near the song menu. So opening the mini player was also clicking an unrelated button.

Fixed in eb1a7994 by closing the `tp-yt-iron-dropdown` the entry actually sits in, resolved from the click target, with an Escape keydown as the fallback.

Worth flagging: I copied that line from `src/plugins/picture-in-picture/renderer.tsx`, which still does the document-wide `#icon` click. Happy to send a separate PR for that one if you want it fixed too — I left it alone here to keep this diff on topic.

PR #4671, opened 2026-08-31
In app menu performance fix

gothmiku:
## Summary by CodeRabbit

* **Performance Improvements**
  * Improved title bar scrolling for smoother movement and more responsive interactions.
  * Reduced unnecessary visual updates during scrolling.
  * Improved rendering efficiency for browse lists, helping pages load and scroll more smoothly.
  * Reduced background processing for menu tooltips when they are not visible.
  * Improved performance when navigating long lists.


gothmiku:
I thought that writing a custom list would fix the issue of the lag. However, after doing a comparison with another machine I found that it was a plugin that caused the lag. Enabling and disabling some plugins lead me to "In-App menu" plugin causing the lag on large playlists. I let claude do the analysis of the cause and let it fix it. I deleted the recycle-list plugin I was working on since its deprecated.

Issue #4431, opened 2026-04-21
[Bug]: MPRIS reports incorrect volume


qriaa:
### Preflight Checklist

- [x] I use the latest version of Pear Desktop (Application).
- [x] I have searched the [issue tracker](https://github.com/pear-devs/pear-desktop/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **pear-devs/pear-desktop has NO affiliation with Google or YouTube**

### Pear Desktop (Application) Version

3.11.0-1

### Checklists

- [ ] I use the portable version of the Pear Desktop Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Other Linux

### Operating System Version

Arch Linux, 6.19.12-arch1-1

### What CPU architecture are you using?

x64

### Reproduction steps

1. Enable MPRIS integration plugin (Shortcuts)
2. Run `playerctl volume 0.5`
3. Run `playerctl volume`

### Expected Behavior

Output of `playerctl volume` - 0.500000

### Actual Behavior

Output of `playerctl volume` - 0.200000

### Enabled plugins

1. Navigation
2. Shortcuts (MPRIS)

### Additional Information

Before you ask, no, I do not have the Exponential Volume plugin enabled.
Restarting does not affect the issue.
I am looking for this to be fixed in order for me to use volume controls in MPRIS integrations in my desktop environment.

Chaganti-Reddy:
this is the same bug as #4458 - the MPRIS volume-set path goes through the exact same code (setVolume in song-controls.ts -> peard:update-volume -> ytmusic-player-bar.updateVolume), just triggered by playerctl instead of the HTTP API. raised #4574 against #4458, tested it fixes this one too - playerctl volume 0.5 now reads back 0.5 instead of 0.2.

PR #4672, opened 2026-08-31
fix: make volume writes round-trip with the reported volume

### LINKED ISSUE #4458: [Bug]: API server /volume GET and POST use different scales — read-modify-write is broken
### Preflight Checklist
- [x] I use the latest version of Pear Desktop (Application).
- [x] I have searched the issue tracker for a bug report that matches the one I want to file, without success.
- [x] I understand that pear-devs/pear-desktop has NO affiliation with Google or YouTube

### Pear Desktop (Application) Version
3.11.0.0

Windows

### Operating System Version
Windows 11

x64

### Last Known Working Pear Desktop (Application) version
N/A — the relevant code paths haven't changed since v3.11.0.

### Reproduction steps

The API Server's `/api/v1/volume` GET and POST endpoints operate on different scales, so read-modify-write produces unexpected results. The WebSocket `VOLUME_CHANGED` event uses the same scale as GET.

Steps:
1. Enable the API Server plugin (default port 26538), no other audio-related plugins.
2. Authenticate, save the JWT.
3. `POST /api/v1/volume` with `{ "volume": 60 }`.
4. `GET /api/v1/volume` — observe `state` is well below 60.

Sample observed values:

| POST volume | GET state (and WS `VOLUME_CHANGED`) |
|-------------|-------------------------------------|
| 39          | 13                                  |
| 60          | 29                                  |
| 75          | 47                                  |

This was found while building a Stream Deck integration on top of [kimjammer/youtube-music-controller](https://github.com/kimjammer/youtube-music-controller), which does read-modify-write on volume (GET, +/- a step, POST). The pattern is broken for any client doing the same.

### Expected Behavior

`GET /api/v1/volume` followed by `POST /api/v1/volume` with the value just read should be a no-op. External clients should be able to do read-modify-write on volume without applying their own inverse transform.

### Actual Behavior

GET/WS report values on one scale; POST accepts values on another. Round-tripping doesn't preserve the value.

### Source pointers

The two endpoints route into different layers of YouTube Music's UI stack:

- POST handler at `src/plugins/api-server/backend/routes/control.ts:686-690` calls `controller.setVolume(volume)`. That goes through `src/providers/song-controls.ts:78` which IPC-sends `peard:update-volume` to the renderer. The renderer listener at `src/renderer.ts:135` calls `ytmusic-player-bar.updateVolume(volume)` — i.e. it writes through YouTube Music's player-bar custom element, which applies some transform before forwarding to the underlying player. I haven't dug into what the transform is, since which scale the API should expose externally seems like the more important question.

- GET handler at `src/plugins/api-server/backend/routes/control.ts:693-698` reads cached state. The cache is populated from `peard:volume-changed` events sent by `src/providers/song-info-front.ts:118-130`, which on each `volumechange` event reads `api.getVolume()` — the underlying player API, on the post-transform scale.

So GET reads from the player API; POST writes through the player-bar element. The two layers don't agree on the value, and that's where the asymmetry is introduced. Base Pear, no plugin needed to reproduce.

The MPRIS path at `src/plugins/shortcuts/mpris.ts:318-320` already uses `setVolume(newVolume * 100)` on the same scale `getVolume` returns.

### Discussion

The fix is small in code but the choice of contract isn't obvious, so I wanted to ask before opening a PR. Three options:

1. **Make POST use the same layer as GET.** Change the renderer listener to call `api.setVolume(volume)` instead of `playerBar.updateVolume(volume)`. POST and GET round-trip on the player-API scale. Matches MPRIS. Breaks any existing client that was POSTing on the player-bar's scale.

2. **Make GET use the same layer as POST.** Change `song-info-front.ts` to read from the player-bar element. Round-trips on the player-bar's displayed scale. Trickier — `volumechange` on the `<video>` element may not fire on player-bar-side changes that don't move the underlying volume past the transform's resolution.

3. **Convert in the API server route handlers.** Pick a canonical external scale and apply an inverse transform in `control.ts`. Leaves the renderer and song-info-front alone but requires hard-coding the player-bar's transform, which is fragile against YouTube Music updates.

Happy to open a PR once direction is set.

### Enabled plugins

- API Server
- Synced Lyrics
- Album Color Theme

(Repro confirmed with all audio-affecting plugins disabled, including Exponential Volume and Precise Volume.)

### Additional Information

No response

nixoletas:
Fixes #4458.

### Problem

The API server's `/api/v1/volume` GET and POST work on different scales, so read-modify-write on volume is broken for external clients.

The two endpoints write and read at different layers:

- **POST** goes through `controller.setVolume()` → `peard:update-volume` → `ytmusic-player-bar.updateVolume()`, which applies its own curve before forwarding to the player.
- **GET** (and the WebSocket `VOLUME_CHANGED` event) reads the cache filled from `peard:volume-changed`, which reports `MusicPlayer.getVolume()` — the value after that curve.

So posting a volume and reading it back gives a different number.

### Fix

Write through the player API instead, which is what the `precise-volume` plugin already does, and sync the player bar sliders manually — the bar only updates them for changes it drives itself.

### Verification

Launched the app and drove both paths directly, setting the volume and reading `getVolume()` back:

| set | before (via player bar) | after (via player API) |
|----:|------------------------:|-----------------------:|
| 39  | 13                      | 39                     |
| 60  | 29                      | 60                     |
| 75  | 47                      | 75                     |

The "before" column reproduces the table in #4458 exactly. After the change the round-trip is an identity, so clients can do GET → adjust → POST without applying their own inverse transform.

This likely also covers #4431 (MPRIS reporting incorrect volume), since MPRIS writes through the same `peard:update-volume` path — but I have not verified that one.

`pnpm typecheck`, `pnpm lint`, `pnpm format:check` and `pnpm build` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume adjustment accuracy by rounding values and respecting slider step limits.
  * Volume controls now remain synchronized after updates.
  * Volume changes are applied consistently across playback controls and external media controls.
  * Removed an outdated volume update path for more reliable playback behavior.

### DIFF
--- a/src/plugins/shortcuts/mpris.ts
+++ b/src/plugins/shortcuts/mpris.ts
@@ -7,7 +7,6 @@ import MprisPlayer, {
 } from '@jellybrick/mpris-service';
 import { type BrowserWindow, ipcMain } from 'electron';
 
-import * as config from '@/config';
 import { APPLICATION_NAME } from '@/i18n';
 import { getSongControls } from '@/providers/song-controls';
 import {
@@ -308,13 +307,8 @@ export function registerMPRIS(win: BrowserWindow) {
         : Number.parseFloat((newVolumeState.state / 100).toFixed(2));
     });
 
-    player.on('volume', async (newVolume: number) => {
-      if (await config.plugins.isEnabled('precise-volume')) {
-        // With precise volume we can set the volume to the exact value.
-        win.webContents.send('setVolume', ~~(newVolume * 100));
-      } else {
-        setVolume(newVolume * 100);
-      }
+    player.on('volume', (newVolume: number) => {
+      setVolume(newVolume * 100);
     });
 
     registerCallback((songInfo: SongInfo, event) => {
--- a/src/providers/song-controls.ts
+++ b/src/providers/song-controls.ts
@@ -1,6 +1,7 @@
 // This is used for to control the songs
 import { type BrowserWindow, ipcMain } from 'electron';
 
+import * as config from '@/config';
 import { LikeType } from '@/types/datahost-get-state';
 
 // see protocol-handler.ts
@@ -75,10 +76,16 @@ export const getSongControls = (win: BrowserWindow) => {
       }
     },
     // General
-    setVolume: (volume: ArgsType<number>) => {
+    setVolume: async (volume: ArgsType<number>) => {
       const volumeNumber = parseNumberFromArgsType(volume);
-      if (volumeNumber !== null) {
-        win.webContents.send('peard:update-volume', volume);
+      if (volumeNumber === null) return;
+
+      if (await config.plugins.isEnabled('precise-volume')) {
+        // precise-volume owns the volume UI (saved volume, tooltip, slider,
+        // HUD), so let it apply the change instead of writing the player.
+        win.webContents.send('setVolume', volumeNumber);
+      } else {
+        win.webContents.send('peard:update-volume', volumeNumber);
       }
     },
     setFullscreen: (isFullscreen: ArgsType<boolean>) => {
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -131,11 +131,21 @@ async function onApiLoaded() {
     }
   });
   window.ipcRenderer.on('peard:update-volume', (_, volume: number) => {
-    document
-      .querySelector<HTMLElement & { updateVolume: (volume: number) => void }>(
-        'ytmusic-player-bar',
-      )
-      ?.updateVolume(volume);
+    const value = Math.min(100, Math.max(0, Math.round(volume)));
+
+    // Write through the player API instead of `ytmusic-player-bar.updateVolume`:
+    // the player bar applies its own curve, so writing there does not round-trip
+    // with `getVolume()`, which is what `peard:volume-changed` reports.
+    api?.setVolume(value);
+
+    // The player bar only syncs its sliders for changes it drives itself.
+    for (const selector of ['#volume-slider', '#expand-volume-slider']) {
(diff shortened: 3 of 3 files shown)


nixoletas:
Good catch — fixed in 618c7e0, though not quite the way the suggestion described, so here is what I checked.

Reproducing the side effects inside `renderer.ts` would duplicate state that `precise-volume` owns (`options.savedVolume`, the HUD element, the slider timers). Instead I moved the routing decision up into `song-controls.setVolume`: when `precise-volume` is enabled the change is sent on that plugin's own `setVolume` channel, so it runs its shared path (`saveVolume`, `updateVolumeSlider`, `setTooltip`, `showVolumeSlider`, `showVolumeHud`); otherwise it falls through to `peard:update-volume` as before.

This is the same branch `mpris.ts` already had for its own volume events, so that special case moved into `song-controls` and every caller now gets it — the API server included, which is what was missing.

One correction to the finding while verifying it: `setTooltip` and `saveVolume` were partly reachable before, but through `precise-volume`'s slider `MutationObserver`, not through this handler — and only for changes larger than its `> 4` "manually set" threshold. `showVolumeSlider` and `showVolumeHud` were never on this path, before or after my first commit. So the routing change is a fix for something that was already broken rather than for a regression, but the outcome you asked for is the right one.

Verified by running the app with `precise-volume` enabled and driving the channel `song-controls` now sends on:

| requested | `getVolume()` | `savedVolume` persisted |
|----------:|--------------:|------------------------:|
| 63 | 63 | yes |
| 41 | 41 | yes |

`pnpm typecheck`, `pnpm lint`, `pnpm format:check` and `pnpm build` pass.

Issue #4594, opened 2026-07-15
[Bug]: Does not install/Does not open


DrWafflePhD:
### Preflight Checklist

- [x] I use the latest version of Pear Desktop (Application).
- [x] I have searched the [issue tracker](https://github.com/pear-devs/pear-desktop/issues) for a bug report that matches the one I want to file, without success.
- [x] I understand that **pear-devs/pear-desktop has NO affiliation with Google or YouTube**

### Pear Desktop (Application) Version

3.12.0

### Checklists

- [x] I use the portable version of the Pear Desktop Application.
- [ ] I can reproduce this issue in the [official version of (WEB) YTM](https://music.youtube.com).

### What operating system are you using?

Windows

### Operating System Version

Windows 11 25h2

### What CPU architecture are you using?

arm64 (including Apple Silicon)

### Last Known Working Pear Desktop (Application) version

3.11.0

### Reproduction steps

Run the program or Install the web set up of the program

### Expected Behavior

The program opens / installs

### Actual Behavior

The portable program never runs, not even as admin.
The web installer creates an item in my start menu to search that claims it's missing, and it's exe file is no where to be found in the folder it creates.

### Enabled plugins

None

### Additional Information

3.11.0 installs fine, and updating with the web installer from that version still breaks the application.

WackyWRZ:
Can confirm this issue. 3.12.0 has this problem on my Snapdragon Windows laptop running 26H1 too. Portable will not run, web-install installs but there's no EXE. 

Seems to be limited to ARM64 Windows, 3.12.0 is working fine on x86_64. 

3.11.0 installs and works fine on ARM64 Windows

kalyway101:
I can also confirm the same as @WackyWRZ 

3.11.0 works fine, but 3.12.0 has a problem running on my Snapdragon Windows laptop running 25H2. Portal will not run and the web-install installs but does not include the exe to launch.

PR #4676, opened 2026-08-31
feat(win): also build a self-contained NSIS installer

nixoletas:
### Problem

The Windows installer is `nsis-web`, which downloads the application payload from the release while installing. That keeps the download small, but it means the installer does not work without network access to GitHub at install time — offline machines, locked-down corporate networks, or an installer copied onto a USB stick.

A few of the install reports in the tracker read like this (#4594 for instance, though I could not confirm the cause there).

### Change

Adds a plain `nsis` target alongside `nsis-web`, for the same three architectures.

- `nsis-web` stays and still produces `latest.yml`, so `electron-updater` is untouched.
- The artifacts are named differently by electron-builder (`... Web Setup <version>.exe` vs `... Setup <version>.exe`), so nothing collides.
- `oneClick: false` + `allowToChangeInstallationDirectory` matches the behaviour of the web installer flow.

### Cost, which is the real question here

This is not free: it adds three more full installers per Windows release. On my machine an x64 `nsis` build with `compression: maximum` took roughly 3 extra minutes and produced a ~112MB artifact, so expect something like +9 minutes of CI and ~330MB of release assets across the three arches.

If that is too much, an alternative that costs a third as much is building the offline installer for x64 only, which is where practically all of these reports come from. Happy to trim the PR to that — or to close it if the web installer is a deliberate "one installer only" decision.

### What I actually verified

I built the `nsis` target for x64 from this configuration and it produces the installer as expected (`YouTube Music Setup 3.12.0.exe`, ~112MB) with the packaged app running from `win-unpacked`.

To be clear about the limits: I did not run the installer on a disconnected machine, so the offline claim comes from how `nsis` differs from `nsis-web` by construction rather than from a test I ran. I also did not run the full three-arch release build locally.

## Summary by CodeRabbit

* **New Features**
  * Added a Windows installer option for x64, ia32, and arm64 devices.
  * Users can choose the installation directory during setup.
  * The installer no longer launches the application automatically after installation.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -30,6 +30,11 @@ win:
         - x64
         - ia32
         - arm64
+    - target: nsis
+      arch:
+        - x64
+        - ia32
+        - arm64
     - target: portable
       arch:
         - x64
@@ -37,6 +42,10 @@ win:
         - arm64
   compression: maximum
 nsisWeb:
+  runAfterFinish: false
+nsis:
+  oneClick: false
+  allowToChangeInstallationDirectory: true
   runAfterFinish: false
 linux:
   icon: assets/generated/icons/png


michei69:
Isn't this pretty much pointless considering [the readme gives instructions on how to install without an internet connection](https://github.com/pear-devs/pear-desktop#how-to-install-without-a-network-connection-in-windows)?

PR #4682, opened 2026-09-04
feat(plugins): add Slowed Reverb plugin

nathwn12:
Adds a **Slowed Reverb** plugin: slows playback (0.7-1.3x, pitch follows like classic slowed+reverb) and layers a Dattorro plate reverb, ported from a Chrome-extension DSP into the native \peard:audio-can-play\ graph.

**UI:** a small settings cog next to the volume slider opens a panel with Speed / Reverb sliders (hover-wheel supported, like the volume slider), Active toggle and Reset. Panel inherits YTM theme variables (\--yt-spec-menu-background\, \--ytmusic-text-primary\) so it follows transparency/theme plugins.

**Design notes**
- Renderer-only, no backend/preload. Additive graph only (\source -> Dattorro -> wetGain -> destination\); the base \source -> destination\ path is never disconnected, so it composes with equalizer/compressor.
- Disabled by default (\nabled: false\), \
estartNeeded: false\, full cleanup in \stop()\.
- i18n: \n.json\ block added (\plugins.slowed-reverb.*\); other locales fall back to English via Weblate flow.

**Verification**
- \	sc -p tsconfig.json --noEmit\: clean (0 errors repo-wide)
- \oxlint --type-aware src/plugins/slowed-reverb\: 0 warnings, 0 errors
- \oxfmt --check src/plugins/slowed-reverb\: clean
- Dev-tested (\lectron-vite dev\): plugin loads (\slowed-reverb::renderer executed\), cog injects beside \#volume-slider\, slow/reverb apply live and persist; enabling at defaults changes no loudness.
- Windows \
sis-web:x64\ installer builds successfully with the plugin bundled.

Happy to adjust anything to match maintainer conventions.

## Summary by CodeRabbit

* **New Features**
  * Added a Slowed Reverb audio effect for slowing playback while preserving pitch and adding adjustable reverb.
  * Added player-bar access to activation, playback speed, reverb intensity, and reset controls.
  * Added live updates while adjusting playback speed and reverb settings.
  * Added a dedicated settings panel with synchronized controls and YouTube Music theme support.
* **Localization**
  * Added English labels and descriptions for Slowed Reverb and its controls.

### DIFF
--- a/src/i18n/resources/en.json
+++ b/src/i18n/resources/en.json
@@ -831,6 +831,17 @@
       "description": "Automatically skip silences sections in songs",
       "name": "Skip Silences"
     },
+    "slowed-reverb": {
+      "description": "Slow down playback and add reverb (slowed + reverb).",
+      "name": "Slowed Reverb",
+      "panel": {
+        "active": "Active",
+        "reset": "Reset",
+        "reverb": "Reverb",
+        "speed": "Speed",
+        "title": "Slowed + Reverb"
+      }
+    },
     "sponsorblock": {
       "description": "Automatically Skips non-music parts like intro/outro or parts of music videos where the song isn't playing",
       "name": "SponsorBlock"
--- a/src/plugins/slowed-reverb/engine.ts
+++ b/src/plugins/slowed-reverb/engine.ts
@@ -11,7 +11,9 @@ export function clamp(value: number, min: number, max: number): number {
 }
 
 export function clampSlow(value: number): number {
-  return Math.round(clamp(value, SLOW_MIN, SLOW_MAX) * 100) / 100;
+  const numeric = Number(value);
+  if (!Number.isFinite(numeric)) return 1;
+  return Math.round(clamp(numeric, SLOW_MIN, SLOW_MAX) * 100) / 100;
 }
 
 export function clampIntensity(value: number): number {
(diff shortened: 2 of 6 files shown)


nathwn12:
Addressed all four bot findings in 8f47ad97 (verified locally: \	sc\ clean, \oxlint\ 0/0, \oxfmt\ clean):

1. **clampSlow non-finite** — now returns neutral \1\ instead of falling to 0.7.
2. **Unreachable reconnect branch** — verified every path through \	eardownAudio()\ nulls the node first; removed, \nsureWorklet()\ handles creation.
3. **Wheel commit burst** — trailing 200ms debounce on wheel commits only; slider drag/reset still commit immediately.
4. **DSP at wetGain 0** — worklet early-outs with zeroed output; a \
eset\ is posted before params on 0->positive transitions so re-entry is click-free. DSP math for wet>0 untouched.

One known nit (non-blocking): a wheel tick followed within 200ms by a drag/reset can let the stale trailing commit land last — self-corrects on the next gesture, audio path unaffected. Happy to harden with timer-cancel if you'd like.

nathwn12:
Done in 2527fb29 — pending wheel timers are now cancelled at the start of every non-wheel speed/reverb input, change commit, and Reset (plus the pre-existing destroy cancel), so a queued wheel value can no longer overwrite newer input. Wheel gestures still commit once, trailing-edge, with the final value. Verified: \	sc\ clean, \oxlint\ 0/0, \oxfmt\ clean.

Known residual (non-blocking, noting for the record): config writes from outside the panel (e.g. toggling Active, external sync) don't cancel a pending wheel timer — same 200ms window, same self-correcting character. Happy to cover those paths too if you'd like.

